PR #3, opened 2024-05-12
Update MLIR code with dynamic dispatch

mofeing:
This PR updates the code to use the `ScopedValue`-based dynamic dispatch.

This allows to change libMLIR dynamically while using upstream MLIR.jl package, so the replicated MLIR code can be removed.


mofeing:
I formatted the code because there were formatting inconsistencies (e.g. sometimes tab indenting, others space indenting) and it was giving me problems with auto-save formatting and git committing. I guess it's not a problem.

mofeing:
Refactored the PR so it modifies less files. "CI / Documentation" is crashing due to other reasons.

mofeing:
This PR will be reopened when Reactant is more stable and we can replace the local MLIR.jl copy for the upstream dep.

PR #6, opened 2024-05-25
Support `struct` types in `create_result`

mofeing:
This PR adds a case for `create_result` to be able to process custom `struct` types.

It also adds support for `Symbol` and `Array`.


src/Reactant.jl
```diff
@@ -720,10 +720,24 @@ function generate_jlfunc(concrete_result, client, mod, Nargs, linear_args, linea
             end)
             return
         end
-        if T <: Int || T <: AbstractFloat || T <: AbstractString || T <: Nothing
```
wsmoses, line 723: I don't think this works since Float32 <: AbstractFloat but is not in the set

src/Reactant.jl
```diff
@@ -720,10 +720,24 @@ function generate_jlfunc(concrete_result, client, mod, Nargs, linear_args, linea
...
         end
+        if isstructtype(T)
+            elems = Symbol[]
+            nf = fieldcount(T)
+            for i in 1:nf
+                sym = Symbol(resname, :_, i)
+                create_result(getfield(tocopy, i), sym, (path..., i))
+                push!(elems, sym)
+            end
+            push!(concrete_result_maker, quote
```
wsmoses, line 735: I think this depends on whether the structure is mutable (Enzyme.make_zero should have some relevant code to look at).

wsmoses (commented):
Add a test?

src/Reactant.jl
mofeing: ah good catch

src/Reactant.jl
mofeing: Added a test on a `mutable struct` type and seems to work... although it might be too naive?

wsmoses:
So something which definitely fail right now is a recursive struct of
sorts, like a linked list.

test/struct.jl
```diff
@@ -0,0 +1,47 @@
...
+
+@testset "Mutable Struct" begin
+    x = MutableMockTensor(rand(4, 4), [:i, :j])
+    x2 = MutableMockTensor(Reactant.ConcreteRArray(parent(x)), x.inds)
+
+    f = Reactant.compile(cos, (x2,))
+    y = f(x2)
+
+    @test y isa MutableMockTensor{Float64,2,Reactant.ConcreteRArray{Float64,(4, 4),2}}
+    @test isapprox(parent(y), cos.(parent(x)))
```
wsmoses, line 45: Can these test against non array values just in case

wsmoses (approved):

test/struct.jl
mofeing: mmm I don't follow. do you mean to add a test against non-array fields?

src/Reactant.jl
```diff
@@ -342,7 +342,7 @@ using Enzyme
     end
 
     if Val(T) ∈ seen
-        return seen[T]
+        return T
```
mofeing, line 345: This was crashing. I believe this is a bug because `seen` is not a `Dict` here, but a vector of types.

test/struct.jl
mofeing: I've added a test against the `inds` field, which is an `Array` but not a `RArray`. Is this enough?

mofeing:
> So something which definitely fail right now is a recursive struct of sorts, like a linked list.

I've added a test for this kind of structure, although it fails in `create_result` due to a problem with `results_store`.

PR #5, opened 2024-05-14
[WIP] Add a Lux example

avik-pal:
Needs https://github.com/EnzymeAD/Reactant.jl/issues/4

PR #7, opened 2024-05-27
Patches for Lux integration

avik-pal:
I am pulling out the patches from https://github.com/EnzymeAD/Reactant.jl/pull/5, so that we can get them in faster


wsmoses (commented):
Could array interface be done in an extension package if it's not in base

avik-pal:
> Could array interface be done in an extension package if it's not in base

Done

src/Reactant.jl
```diff
@@ -138,6 +138,13 @@ end
         seen[prev] = res
         return res
     end
+
+    if RT <: TracedRArray
+        @assert typeof(prev) <: RT # prev has concrete shape while RT might not have that
+        res = typeof(prev)(prev.paths, prev.mlir_data)
```
wsmoses, line 144: This isn't the right way to construct. You should construct it as a stablehlo.constant op (with zeros). I think we already have a similar overload for base.zero. alternatively we should reasonably be able to dispatch to base.zero here

src/utils.jl
```diff
@@ -7,7 +7,7 @@ function transpose_val(val)
     return MLIR.IR.result(MLIR.Dialects.stablehlo.transpose(val, permutation=attr), 1)
 end
 
-function apply(f, args...; kwargs)
+function apply(f, args...; kwargs...)
```
wsmoses, line 10: Can you add a test for what needed this. I somehow recall the other way being intentional but maybe not

avik-pal: This came up in computing the gradient of the model with Enzyme. But that code is not completely functional yet, see https://github.com/LuxDL/Lux.jl/pull/665#issuecomment-2133946153

wsmoses:
Ah looking at that error I think we should trace the primal and shadow
arrays into tracedrarrays so then the duplicates invariant holds. We'll
need the data as an rarray anyways to run it

wsmoses:
However separate from this fix (which ought be done regardless) it is
surprising that the original model is not an rarray as I would expect. That
should probably be fixed in lux

src/Reactant.jl
avik-pal: Updated to use the code from `Base.fill!`

Project.toml
```diff
@@ -12,16 +12,19 @@ Preferences = "21216c6a-2e73-6563-6e65-726566657250"
...
 ReactantNNlibExt = "NNlib"
 
 [compat]
-Cassette = "0.3"
+ArrayInterface = "7.10"
 CEnum = "0.4, 0.5"
+Cassette = "0.3"
 Enzyme = "0.11, 0.12"
-Reactant_jll = "0.0.6"
+NNlib = "0.9.17"
```
wsmoses, line 27: change this to 0.9, no need for 17 specificity

wsmoses (approved):

Project.toml
avik-pal: Done

src/overloads.jl
```diff
@@ -549,6 +549,12 @@ Cassette.overdub(context::TraceCtx, f::typeof(Base.materialize!), args...) = f(a
 @inline Base.copyto!(dest::TracedRArray, bc::Broadcasted{Nothing}) =
     _copyto!(dest, bc) # Keep it for ArrayConflict
 
+@inline function Base.copyto!(dest::TracedRArray{ElType, Shape, N},
```
avik-pal, line 552: @wsmoses can you confirm that this is the right way to implement copyto!? I am getting https://github.com/LuxDL/Lux.jl/pull/665#issuecomment-2134147847 so it is possible I got something wrong

wsmoses: Yup that is perfect!

wsmoses (approved):

.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
wsmoses, line 1: remove this and add to .gitignore?

Issue #8, opened 2024-05-27
Define `create_result` for `NamedTuple`


avik-pal:
Currently it errors if the result has a NamedTuple. All Lux models return a state (NamedTuple), so it would be nice to add a case for that.

avik-pal:
Maybe https://github.com/EnzymeAD/Reactant.jl/pull/6/files already does this?

PR #11, opened 2024-05-28
Refactor `create_result` for `Tuple` types to include `NamedTuple`s

mofeing:
@avik-pal regarding #8, does this PR fixes your problem?


src/Reactant.jl
```diff
@@ -708,11 +708,11 @@ function generate_jlfunc(concrete_result, client, mod, Nargs, linear_args, linea
             push!(concrete_result_maker, :($resname = $T($(result_stores[path]))))
             return
         end
-        if T <: Tuple
+        if T <: Tuple || T <: NamedTuple
```
wsmoses, line 711: I don't think this resolves.

Specifically the resname = ... won't create a named tuple

wsmoses:
add a test?

src/Reactant.jl
mofeing: right, fixed.

mofeing:
> add a test?

Added!

Why is the `CI / Julia 1.9 - macOS-latest - x64 - packaged libReactant - assertions=false - pull_request (pull_request)` workflow failing?

wsmoses (approved):

wsmoses:
Seemed like random non-determinstic failure

PR #34, opened 2024-06-28
Missing `Meta.quot`

avik-pal:
code gets generated as `((Base.getfield(Base.getfield(args[3], layer_1), bias)).data = concrete_res_1)` instead of `((Base.getfield(Base.getfield(args[3], :layer_1), :bias)).data = concrete_res_1)`

### DIFF
--- a/src/Reactant.jl
+++ b/src/Reactant.jl
@@ -798,7 +798,7 @@ end
                 path = path[3:end]
             end
             for p in path
-                res = :(Base.getfield($res, $p))
+                res = :(Base.getfield($res, $(Meta.quot(p))))
             end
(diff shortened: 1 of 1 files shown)


src/Reactant.jl
mofeing, line 1: You also need to add it in lines 827 and 832

.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
mofeing, line 1: I think you committed this accidentally? Can you delete it please?

I'm pushing some update gitignore rules so this doesn't happen again.

avik-pal: oops yeah my bad. new system don't have my global git config correctly setup yet.

PR #35, opened 2024-06-29
Generalize broadcast

wsmoses:


wsmoses:
Companion PR to https://github.com/EnzymeAD/Enzyme/pull/1952

Untested, but should enable arbitrary broadcast support (if all scalar functions are overloaded)

src/Reactant.jl
```diff
@@ -429,7 +429,7 @@ function append_path(path, i)
     return (path..., i)
 end
 
-@inline function make_tracer(seen::IdDict, prev::RT, path, mode, data) where {RT}
+@inline function make_tracer(seen::IdDict, prev::RT, path, mode, data; toscalar=false, tobatch=nothing) where {RT}
```
mofeing, line 432: What do these kwargs do?

wsmoses: whether the tracer we create should automatically change all arrays to scalars, or scalars to arrays

src/overloads.jl
```diff
@@ -457,6 +457,95 @@ for (jlop, hloop) in (
     end
 end
 
+
+function elem_apply(f, args::VarArgs{Nargs})
```
mofeing, line 461: Need `where` here:
```suggestion
function elem_apply(f, args::VarArgs{Nargs}) where {Nargs}
```

src/utils.jl
```diff
@@ -13,7 +13,7 @@ function apply(f, args...; kwargs...)
     return f(args...; kwargs...)
 end
 
-function make_mlir_fn(mod, f, args, kwargs, name="main", concretein=true)
+function make_mlir_fn(mod, f, args, kwargs, name="main", concretein=true; toscalar=false)
```
mofeing, line 16: Maybe we could turn `name` and `concretein` into kwargs?
```suggestion
function make_mlir_fn(mod, f, args, kwargs; name="main", concretein=true, toscalar=false)
```

src/overloads.jl
```diff
@@ -457,6 +457,95 @@ for (jlop, hloop) in (
...
+    primf = f.val
+
+    mod = MLIR.IR.module()
```
mofeing, line 464:
```suggestion
    mod = MLIR.IR.mmodule()
```

src/overloads.jl
```diff
@@ -457,6 +457,95 @@ for (jlop, hloop) in (
...
+        end
+    end
+
+
+    function act_attr(val)
+        val = @ccall MLIR.API.mlir_c.enzymeActivityAttrGet(
+            MLIR.IR.context()::MLIR.API.MlirContext, val::Int32
+        )::MLIR.API.MlirAttribute
+        return MLIR.IR.Attribute(val)
+    end
```
mofeing, line 511: This should be declared out of `elem_apply`'s definition if it's not capturing any variable.

mofeing (changes requested):
Just some fix requests and a couple of aesthetic suggestions.

Also, we should add a test for this.

avik-pal:
https://github.com/EnzymeAD/Reactant.jl/blob/5134fe0d2bb6f6d7a4f72c1f0e2d996fe744d0f9/deps/clang#L2 need to be generalized?

src/overloads.jl
```diff
@@ -457,6 +457,95 @@ for (jlop, hloop) in (
...
+
+    fnwrap, func2, traced_result, result, seen_args, ret, linear_args, in_tys, linear_results = make_mlir_fn(
+        mod, f, args, (), string(f) * "_broadcast_scalar", false
+    )
+
+    invmap = IdDict()
+    OutShape = nothing
+    for (k, v) in seen_args
+        invmap[v] = k
+        OutShape size(k)
```
avik-pal, line 474:
```suggestion
        OutShape = size(k)
```

PR #36, opened 2024-07-01
Support closures

mofeing:
This PR adds support for closures. The main changes are to `traced_type`, so that tracing correctly generates the new type of the traced closure, and to the generated `Thunk` function so that it correctly prepends the closure as an argument.

It also provides a test.


mofeing:
mmm none of the check run the tests?

wsmoses:
oof we should fix that

src/Reactant.jl
```diff
@@ -865,6 +892,13 @@ end
     resexpr = create_result(concrete_result_ty, (), result_stores)
     expr = quote
         Base.@_inline_meta
+        $(
+            # if `f` is a closure, then prepend the closure into `args`
+            # the closure fields will be correctly extracted from it as the tracer has already passed through it
+            if !(closure_ty <: Nothing)
+                :(args = [thunk.fnwrap, args...])
```
wsmoses, line 899: Can we do this with a tuple rather than list for performance

mofeing: Sure!

wsmoses:
Alternatively would it make sense to modify the expressions that use args
to do an offset.

That’s what’s done for example in the AD rewrite and the batch PR

mofeing:
That's what I started to do, but it soon started to get quirky. The reason is that since we trace `Reactant.apply(f, args...)` instead of `f(args...)`, then `linear_arg_paths` and `preserved_arg_paths` consider that `f` will be the first argument (e.g. the enclosed `ConcreteRArray`s in `f` will be read from `args[1] = f`).

It can be done, but involves a lot more refactoring and probably changing the way we trace the closures. I did it this way because (1) it has no performance penalty and (2) modifies the least amount of code.

wsmoses:
Yeah that sounds reasonable to me!

mofeing:
I'm merging this since tests for this feature pass. The failing tests are unrelated to this PR and are the CI's configuration fault.

PR #37, opened 2024-07-07
Remove Cassette

mofeing:
As far as I seen in the code, Cassette's contextual dispatch is only needed for overriding `Enzyme.autodiff`. But Cassette has a very high overhead that kills performance. I have started to see very high compile times in some local examples.

This PR tries to remove Cassette from the equation by defining a new `autodiff` function (`Reactant.autodiff`). This might not be the best solution, but it's an idea I want to experiment with.


wsmoses:
Er, I would like to keep the behavior of being able to override the enzyme
autodiff in these cases.

I think the correct answer here is perhaps to leverage the abstract
interpreter stuff but I’m more quite sure how to use it here

mofeing:
@jumerckx is working in something similar for upstream MLIR.jl https://github.com/JuliaLabs/MLIR.jl/pull/77

Maybe we can implement some simplified version of this.

wsmoses:
@mofeing I think emulating this is the right solution https://github.com/JuliaGPU/GPUCompiler.jl/pull/582/files

cc @vchuravy

PR #38, opened 2024-07-13
fix elem_apply usage in mapreduce

Pangoraw:


wsmoses (approved):

mofeing:
The tests pass. It's Windows and macOS which only fail.

PR #47, opened 2024-07-15
set global module in compile_to_module

Pangoraw:
Otherwise, `@code_hlo` was actually failing. I added a simple test for `@code_hlo`.


wsmoses (approved):
Oh whoops that’s my bad from the batch integration. Good call on the test

PR #49, opened 2024-07-17
Run pass pipeline from Julia

Pangoraw:
This prevents exiting when pass fails, and throws a julia exception instead.


mofeing (approved):
awesome!

PR #52, opened 2024-07-19
Refactor tracing system to get rid of huge compilation times

mofeing:
I'm getting a huge compilation times (>99%) in my tests and this is due to the fact that Reactant is riddled with type unstable code, overspecialization and maybe too much metaprogramming. Meanwhile, Jax gets this done instantaneously.

This is my try to get rid of all those problems by refactoring how we trace the shapes of the arrays and cleaning some parts of the code so we don't depend so much on Julia's runtime compilation.

:warning: This is a WIP, not ready to merge yet. Since, this PR changes the public API, it might be a good idea to do a minor release or wait to v0.2.

## Main changes

- `RArray` (and thus, `ConcreteRArray` and `TracedRArray`) no longer track their shape through a type parameter but through a `shape` field.
- Moved all code related to the tracing system to a new `Tracing.jl` file.
- Remove `@inline` from almost all methods.

## To do
- [ ] Refactor `Thunk` to avoid the `@generated` functor.
  - Might be a good idea to move the codegen to the `generate_jlfunc` method.
- [ ] Refactor `traced_type` to rely on multiple dispatch (i.e. like `make_tracer`).


ext/ReactantNNlibExt.jl
```diff
@@ -7,9 +7,9 @@ for (jlop, hloop) in ((:(NNlib.tanh), :tanh), (:(NNlib.tanh_fast), :tanh))
     @eval begin
         if $jlop != Base.tanh && $jlop != Base.FastMath.tanh_fast
             function Reactant.elem_apply(
-                ::typeof($jlop), lhs::Reactant.TracedRArray{ElType,Shape,N}
```
wsmoses, line 10: So some of these changes [like here don't actually require moving the shape type param]. Instead we can do
`lhs::Reactant.TracedRArray{ElType,Shape,N} where Shape` and presumably get the same speedups?

I'm curious what the breakdown of compile time is presently

mofeing: > So some of these changes [like here don't actually require moving the shape type param]. Instead we can do `lhs::Reactant.TracedRArray{ElType,Shape,N} where Shape` and presumably get the same speedups?

I guess this would still be type-unstable.

> I'm curious what the breakdown of compile time is presently

I don't have the plots with me right now, but I tried using `Profiler` with no results (obviously, because `Profiler` only measures run-time but there is a 99% compile time) and `SnoopCompile`.

With `SnoopCompile`, approximately a 20-30% was spent in type-inference. The 70-80% rest is probably spent in codegen.

mofeing:
# Update

The code has been refactored to not depend on `@generated` functions and try to be more type-stable. Also, now performance analysis should be easier and should show more info.

wsmoses:
Saying this from my phone without access to the code but we definitely want
to keep the generated function for the call.

Specifically we want that call to be equivalent to just doing the Julia
level load stores and call to xla since it is certainly going to be the hot
loop.

We definitely want to also speed up compilation but it can’t come at the
expense of the hot loop

src/Compiler.jl
```diff
@@ -0,0 +1,500 @@
...
+                        Val($(length(linear_results))),
+                    )
+                end
+            end
+        end
+
+        resexpr = create_result(concrete_result, (), result_stores)
+
+        fname = gensym(Symbol(Symbol(f), :_reactant))
+
```
wsmoses, line 466: this isn't a valid key -- itn's not unique [since we will conditionally branch on the runtime values].

src/Compiler.jl
```diff
@@ -0,0 +1,500 @@
...
+            )
+            $exec_call
+            $(concretize...)
+            # Needs to store into result
+            result = $resexpr
+            $(delinearized_results...)
+            return result
+        end)
+
+        body = expr.args[2]
```
wsmoses, line 483: can we just return a Thunk{body} or something?

mofeing: I believe not because `Expr` is not a `isbitstype`. But this technique works, as they use it in RuntimeGeneratedFunctions.jl, and haven't seen any performance regression.

wsmoses: No, my concern here is a correctness issue per above. We don't have a valid unique key we can use for the map.

wsmoses:
Can you separate out the different features [e.g. shape from type, reorganization, thunk, etc] into separate PR's?

src/Compiler.jl
mofeing: mmm why not? `gensym` generates a unique identifier. And though I don't know which branches are you reference about, any branch inside the function body (the `Expr` we generate) should be preserved.

wsmoses: oh I had just read quickly and assumed it was append.

My worry is a conflict in some case like this. The 1 and 2 are embedded into the expression itself so a cache that uniques only on fn names, will accidentally have these two collide [and produce wrong answers].

```
f(x, y) = x .+ y

x = Rarray([4.1])
compile(f, x, 1)
compile(f, x, 2)
```

If gensym is unique each time, I don't understand why this improvess compile perf from the current thunk?

In any case, if you can separate out the features and we can do a proper ablation tests

mofeing:
Some of them are interlinked, so I cannot separate all of them. But I think I can do different PRs for the following features:
1. `RArray` (and thus, `ConcreteRArray` and `TracedRArray`) no longer track their shape through a type parameter but through a shape field.
2. Move all tracing-related code to `Tracing.jl`, refactor `traced_type`.
3. Move all compiler-related code to `Compiler.jl` and refactor `create_result (also introduce `traced_getfield`).
4. Replace `Thunk` serialization through type parameters for a `Expr` caching system inspired by [RuntimeGeneratedFunctions.jl](https://github.com/SciML/RuntimeGeneratedFunctions.jl).

wsmoses:
also to quantify the gaps maybe we can add something like: https://github.com/EnzymeAD/Enzyme.jl/blob/main/benchmark/benchmarks.jl here

src/Compiler.jl
mofeing: Ah, this doesn't improve the compiler performance _per se_: all the other changes do. What it does is simplify the `Thunk` code in <10 lines with the same performance.

Also, `gensym` generates a unique identifier on each call (the argument is just to prepend sth, but a counter is added in the end) so that example would work without problem.

```julia
# inspired by RuntimeGeneratedFunction.jl
 const __thunk_body_cache = Dict{Symbol,Expr}()

 struct Thunk{tag} end

 @generated function (thunk::Thunk{tag})(args...) where {tag}
     return __thunk_body_cache[tag]
 end

 function register_thunk(tag, body)
     __thunk_body_cache[tag] = body
     return Thunk{tag}()
 end
```

IMO serializing a `Expr` to type params can be problematic in the future.

mofeing:
Closing in favor of several other minor PRs:
1. #58
2. #59 
3. #60 
4. #61 

They need to be merged in that order

PR #53, opened 2024-07-20
Fix multisymbol issues

wsmoses:
Discovered while looking at macos arm.

@mofeing do you know how to ensure the default is on the release not debug mode [it looks like I default hit debug, which was good for now, but also a problem in the future for users].

Incidentally otherwise the debug one works...

### DIFF
--- a/src/mlir/IR/SymbolTable.jl
+++ b/src/mlir/IR/SymbolTable.jl
@@ -1,4 +1,4 @@
-struct SymbolTable
+mutable struct SymbolTable
     st::API.MlirSymbolTable
 
     function SymbolTable(st)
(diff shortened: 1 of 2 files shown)


mofeing:
what's the issue with `SymbolTable`? I mean it to fix it in upstream MLIR.jl

about the default artifact, let me try some stuff

wsmoses:
The finalizer is otherwise broken

mofeing:
Can you tell what does this code returns to you?
```julia
using Base.BinaryPlatforms
using Reactant_jll

Reactant_jll.augment_platform!(HostPlatform())
```

In my case, it correctly sets to `opt` mode:
```julia
julia> Reactant_jll.augment_platform!(HostPlatform())
macOS aarch64 {cuda=none, cxxstring_abi=cxx11, julia_version=1.10.4, libgfortran_version=5.0.0, mode=opt}
```

PR #55, opened 2024-07-26
feat: more coverage for common NN operations

avik-pal:
## Overview

- [x] Activations
  - [x] `sigmoid`
  - [x] `sigmoid_fast`
  - [x] `relu`
  - [x] `gelu`
- [x] `abs2`
- [x] normalization
  - [x] `mean`
  - [x] `var`
- [x] Add tests
  - [x] Activations
  - [x] abs2
  - [x] normalization


avik-pal:
Need to fix the compats. Lux failure will be gone with https://github.com/LuxDL/LuxLib.jl/pull/105.

ext/ReactantNNlibExt.jl
```diff
@@ -3,23 +3,38 @@ module ReactantNNlibExt
...
+                ),
+            )
         end
     end
 end
 
+function Reactant.elem_apply(
+    ::typeof(NNlib.relu), lhs::Reactant.TracedRArray{ElType,Shape,N}
+) where {ElType,Shape,N}
+    return (lhs .> zero(ElType)) .* lhs  # base case uses ifelse, so we compile the product
```
wsmoses, line 28: We should just add an ifelse overload to use select

avik-pal: Ah still figuring out what the operations are called in stablehlo specs. Will update this

avik-pal: @wsmoses seems like `select` derivative isn't implemented?

```julia
error: Unimplemented derivative for argument 0 in reverse mode for op %4 = "stablehlo.select"(%3, %2, %1) : (tensor<10x10xi1>, tensor<10x10xf32>, tensor<10x10xf32>) -> tensor<10x10xf32>

error: could not compute the adjoint for this operation %3 = "stablehlo.compare"(%2, %1) <{comparison_direction = #stablehlo<comparison_direction GT>}> : (tensor<10x10xf32>, tensor<10x10xf32>) -> tensor<10x10xi1>
```

Should I use the multiply version for now and we can replace it once the adjoint is implemented?

avik-pal:
1.11 failures are due to Enzyme not working on 1.11

ext/ReactantNNlibExt.jl
avik-pal: my bad using the multiply version wont help as well, the adjoint for `compare` is not defined.

Pangoraw: I opened https://github.com/EnzymeAD/Enzyme-JAX/pull/106 for the derivative of compare. 
in the meantime, running tests with
```
function Reactant.elem_apply(
    ::typeof(NNlib.relu), lhs::Reactant.TracedRArray{ElType,Shape,N}
) where {ElType,Shape,N}
    return max.(lhs, zero(ElType))
```
make relu test pass

ext/ReactantNNlibExt.jl
```diff
@@ -3,23 +3,38 @@ module ReactantNNlibExt
...
+                (),
+                Reactant.MLIR.IR.result(
+                    Reactant.MLIR.Dialects.stablehlo.$hloop(lhs.mlir_data), 1
+                ),
+            )
         end
     end
 end
 
+function Reactant.elem_apply(
```
wsmoses, line 25: We generally shouldn’t need to add more elem alllirs anymore now that we have batching properly. I think just defining this for an rarray of size zero should suffice

src/overloads.jl
```diff
@@ -59,28 +59,49 @@ for (jlop, hloop, RT) in (
...
+                    MLIR.IR.result(
+                        MLIR.Dialects.stablehlo.$hloop(lhs.mlir_data, rhs.mlir_data), 1
+                    ),
+                )
+            end
         end
     end
 end
 
+Base.abs2(x::Reactant.TracedRArray{T,(),0}) where {T} = x * x
```
wsmoses, line 103: Technically should we do this as conjugate of x

src/overloads.jl
```diff
@@ -140,6 +172,20 @@ for (jlop, hloop, RT) in (
     end
 end
 
+function elem_apply(
```
wsmoses, line 175: Same comment on just defining an overload for ifelse directly rather than elemapply

ext/ReactantNNlibExt.jl
```diff
@@ -3,23 +3,38 @@ module ReactantNNlibExt
...
+    ::typeof(NNlib.relu), lhs::Reactant.TracedRArray{ElType,Shape,N}
+) where {ElType,Shape,N}
+    return ifelse.((lhs .> zero(ElType)), lhs, zero(ElType))
+end
+
+function Reactant.elem_apply(
+    ::typeof(NNlib.gelu), lhs::Reactant.TracedRArray{ElType,Shape,N}
+) where {ElType,Shape,N}
+    # See https://arxiv.org/pdf/1606.08415v5 Section 2
+    return lhs .* sigmoid.(ElType(1.702) .* lhs)
```
avik-pal, line 35: Is there an `erf` op in HLO?

wsmoses: One of the most positive things of setting up docs for the hlo stuff: https://enzymead.github.io/Reactant.jl/dev/api/#Reactant.MLIR.Dialects.chlo.erf-Tuple{Reactant.MLIR.IR.Value}

ext/ReactantNNlibExt.jl
avik-pal: I tried this

```julia
function NNlib.relu(x::Reactant.TracedRArray{T,(),0}) where {T}
    return max(x, zero(T))
end
```

but I am getting

```julia
error: 'stablehlo.constant' op inferred type(s) 'tensor<f32>' are incompatible with return type(s) of operation 'tensor<2x3xf32>'
error: 'stablehlo.constant' op failed to infer returned types
ERROR: "failed to run pass manager on module"
Stacktrace:
 [1] run!
   @ /mnt/research/lux/XLA/Reactant.jl/src/mlir/IR/Pass.jl:70 [inlined]
 [2] run_pass_pipeline!(mod::Reactant.MLIR.IR.Module, pass_pipeline::String)
   @ Reactant /mnt/research/lux/XLA/Reactant.jl/src/Reactant.jl:1178
 [3] compile_to_module(mod::Reactant.MLIR.IR.Module, f::Function, args::Vector{Any}; optimize::Bool)
   @ Reactant /mnt/research/lux/XLA/Reactant.jl/src/Reactant.jl:1198
 [4] (::var"#51#52")()
   @ Main /mnt/research/lux/XLA/Reactant.jl/src/Reactant.jl:1315
 [5] context!(f::var"#51#52", ctx::Reactant.MLIR.IR.Context)
   @ Reactant.MLIR.IR /mnt/research/lux/XLA/Reactant.jl/src/mlir/IR/Context.jl:71
 [6] top-level scope
   @ /mnt/research/lux/XLA/Reactant.jl/src/Reactant.jl:1313
```

Pre-optimize the code is

```stablehlo
Module:
module {
  func.func private @relu_broadcast_scalar(%arg0: tensor<f32>) -> (tensor<f32>, tensor<f32>) {
    %0 = stablehlo.transpose %arg0, dims = [] : (tensor<f32>) -> tensor<f32>
    %1 = stablehlo.maximum %0, %cst : tensor<f32>
    %2 = stablehlo.transpose %0, dims = [] : (tensor<f32>) -> tensor<f32>
    %3 = stablehlo.transpose %1, dims = [] : (tensor<f32>) -> tensor<f32>
    return %2, %3 : tensor<f32>, tensor<f32>
  }
  func.func @main(%arg0: tensor<3x2xf32>) -> (tensor<3x2xf32>, tensor<3x2xf32>) {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<3x2xf32>) -> tensor<2x3xf32>
    %1:2 = enzyme.batch @relu_broadcast_scalar(%0) {batch_shape = array<i64: 2, 3>} : (tensor<2x3xf32>) -> (tensor<2x3xf32>, tensor<2x3xf32>)
    %2 = stablehlo.transpose %1#0, dims = [1, 0] : (tensor<2x3xf32>) -> tensor<3x2xf32>
    %3 = stablehlo.transpose %1#1, dims = [1, 0] : (tensor<2x3xf32>) -> tensor<3x2xf32>
    return %2, %3 : tensor<3x2xf32>, tensor<3x2xf32>
  }
}
```

ext/ReactantNNlibExt.jl
avik-pal: Done. Thanks!

ext/ReactantNNlibExt.jl
avik-pal: So we can use any of the dialects? I was using https://openxla.org/s/results?q=erf#gsc.tab=0&gsc.q=erf&gsc.sort= as a reference

avik-pal: What do we feel about adding a direct dep on `SpecialFunctions` for `erf`/`erfinv`...? Else we will have to create a 2nd NNlibSpecialFunctionsExt to define the exact gelu impl

wsmoses: presently yes (and we need to make sure we have corresponding lowering from one to the others, and potentially derivatives)

ext/ReactantNNlibExt.jl
wsmoses: This is the "we dont support constants in batching yet" which I'm presently working on. I'll try to get this squared today/tomorrow.

avik-pal: Should I just leave a "TODO" in comment for relu or wait for the support?

ext/ReactantNNlibExt.jl
avik-pal: `erf` doesn't seem to have the derivatives implemented https://github.com/EnzymeAD/Enzyme-JAX/issues/88, so I am more inclined to keep the current implementation, and switch it later once derivatives are implemented.

wsmoses:
It shouldn’t be hard to add (like two lines) if you want to go for it. I
also have an open PR fixing batched constants so we can get the new jll of
all the stuff at once

ext/ReactantNNlibExt.jl
wsmoses: do you want to take a stab at it? It would go right here https://github.com/EnzymeAD/Enzyme-JAX/blob/4eeaef06e0da144bebd08ec739cf01911dcddb47/src/enzyme_ad/jax/Implementations/CHLODerivatives.td#L142 and shouldn't be bad cc @mofeing and or @Pangoraw who may be able to help with

avik-pal: I will give it a shot

wsmoses:
@avik-pal the jll bump with constant propagation of broadcast is merged. rebase this?

avik-pal:
Still seems to error https://github.com/EnzymeAD/Reactant.jl/actions/runs/10207710411/job/28242989697?pr=55#step:9:753

wsmoses:
Darn, can you extract the failing input hlo?

avik-pal:
```mlir
Module:
module {
  func.func private @relu_broadcast_scalar(%arg0: tensor<f64>) -> (tensor<f64>, tensor<f64>) {
    %1 = stablehlo.maximum %0, %cst : tensor<f64>
    %2 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
    %3 = stablehlo.transpose %1, dims = [] : (tensor<f64>) -> tensor<f64>
    return %2, %3 : tensor<f64>, tensor<f64>
  }
  func.func @main(%arg0: tensor<2x2xf64>) -> (tensor<2x2xf64>, tensor<2x2xf64>) {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<2x2xf64>) -> tensor<2x2xf64>
    %1:2 = enzyme.batch @relu_broadcast_scalar(%0) {batch_shape = array<i64: 2, 2>} : (tensor<2x2xf64>) -> (tensor<2x2xf64>, tensor<2x2xf64>)
    %2 = stablehlo.transpose %1#0, dims = [1, 0] : (tensor<2x2xf64>) -> tensor<2x2xf64>
    %3 = stablehlo.transpose %1#1, dims = [1, 0] : (tensor<2x2xf64>) -> tensor<2x2xf64>
    return %2, %3 : tensor<2x2xf64>, tensor<2x2xf64>
  }
}
```

wsmoses:
@avik-pal the answer was silly that we just didn't bump the repo dependency commit: https://github.com/EnzymeAD/Reactant.jl/commit/8f25b97054405ad827eddf62f571b72d397d3ae7

wsmoses:
Now awaiting jll: https://github.com/JuliaPackaging/Yggdrasil/pull/9196 [tho also @avik-pal this won't include the erf derivative yet -- unless you have a PR we can quickly merge and then get into the new jll]

avik-pal:
let's move forward without the erf derivative for now. I am trying to help one of our GSoCs with some benchmarking so that needs to be finished first :sweat:

wsmoses:
@avik-pal the jll merged, can you rebase here?

Project.toml
```diff
@@ -15,6 +10,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 PackageExtensionCompat = "65ce6f38-6b18-4e1d-a461-8949797d7930"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reactant_jll = "0192cb87-2b54-54ad-80e0-3be72ad8a3c0"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
```
wsmoses, line 13: can we remove this as a full dep and leave as weak?

ext/ReactantNNlibExt.jl
```diff
@@ -3,23 +3,29 @@ module ReactantNNlibExt
 using NNlib
 using Reactant
 
-for (jlop, hloop) in ((:(NNlib.tanh), :tanh), (:(NNlib.tanh_fast), :tanh))
+for (jlop, hloop) in (
+    (:(NNlib.tanh_fast), :tanh),
```
wsmoses, line 7: We shouldn't need these as elem apply's anymore, and just have them apply to the 0-dim case

src/Reactant.jl
```diff
@@ -1,6 +1,7 @@
 module Reactant
 
 using PackageExtensionCompat
+using Statistics: Statistics
```
wsmoses, line 4: can we remove this?

src/overloads.jl
```diff
@@ -140,6 +172,17 @@ for (jlop, hloop, RT) in (
     end
 end
 
+function Base.ifelse(
+    pred::TracedRArray{Bool,(),0}, x::TracedRArray{T1,(),0}, y::TracedRArray{T2,(),0}
```
wsmoses, line 176: can we make this generalize to any shape/size, not just 0?

src/overloads.jl
```diff
@@ -632,3 +675,17 @@ function Base.mapreducedim!(f, op, R::TracedRArray, A::Base.AbstractArrayOrBroad
     R.mlir_data = elem_apply(op, R, tmp).mlir_data
     return R
 end
+
+# Stdlib overloads
+## Statistics
+function Statistics.mean(A::TracedRArray{T,Shape,N}; dims=:) where {T,Shape,N}
```
wsmoses, line 681: ah I see. It would be great if this could be an extension, if not that's fine.

src/overloads.jl
avik-pal: Won't the broadcasting handle the shape automatically?

ext/ReactantNNlibExt.jl
avik-pal: done

src/overloads.jl
wsmoses: Yes, but someone could also call ifelse(true, ones(4,4), zeros(4,4)) or ifelse(trues(4,4), ones(4,4), zeros(4,4)), etc, outside a broadcast [tho yes the 0 dim one will generalize to anything in a broadcast]

wsmoses: Though I don't think the latter case is legal in julia atm, so just generalizing to `ifelse(true, ones(4,4), zeros(4,4))` probably makes sense

avik-pal: Something is wrong with this version I defined:

```julia
julia> f(x) = ifelse.(true, x, x)

julia> Reactant.@code_hlo optimize=false f(x)
Module:
module {
  func.func private @ifelse_broadcast_scalar(%arg0: tensor<i1>, %arg1: tensor<f64>) -> (tensor<i1>, tensor<f64>, tensor<f64>) {
    %0 = stablehlo.transpose %arg0, dims = [] : (tensor<i1>) -> tensor<i1>
    %1 = stablehlo.transpose %arg1, dims = [] : (tensor<f64>) -> tensor<f64>
    %2 = stablehlo.select %0, %1, %1 : tensor<i1>, tensor<f64>
    %3 = stablehlo.transpose %0, dims = [] : (tensor<i1>) -> tensor<i1>
    %4 = stablehlo.transpose %1, dims = [] : (tensor<f64>) -> tensor<f64>
    %5 = stablehlo.transpose %2, dims = [] : (tensor<f64>) -> tensor<f64>
    return %3, %4, %5 : tensor<i1>, tensor<f64>, tensor<f64>
  }
  func.func @main(%arg0: tensor<3x2xf64>) -> (tensor<3x2xf64>, tensor<3x2xf64>) {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<3x2xf64>) -> tensor<2x3xf64>
    %c = stablehlo.constant dense<true> : tensor<2x3xi1>
    %1:3 = enzyme.batch @ifelse_broadcast_scalar(%c, %0) {batch_shape = array<i64: 2, 3>} : (tensor<2x3xi1>, tensor<2x3xf64>) -> (tensor<2x3xi1>, tensor<2x3xf64>, tensor<2x3xf64>)
    %2 = stablehlo.transpose %1#2, dims = [1, 0] : (tensor<2x3xf64>) -> tensor<3x2xf64>
    %3 = stablehlo.transpose %1#1, dims = [1, 0] : (tensor<2x3xf64>) -> tensor<3x2xf64>
    return %2, %3 : tensor<3x2xf64>, tensor<3x2xf64>
  }
}

julia> Reactant.@code_hlo f(x)
Module:
  func.func @main(%arg0: tensor<3x2xf64>) {
    return
  }
}

```

PR #56, opened 2024-07-26
test: testing using ReTestItems

avik-pal:
See https://github.com/JuliaTesting/ReTestItems.jl

The main benefits are

- parallel testing (not super relevant for GH Actions since they are 2 core machines but helps a lot on the buildkite ones)
- prevents stuff from spilling over to other files
- loads Reactant and Test in every @testitem block automatically
- lets you run specific tests locally by specifying `name = ...`


avik-pal:
@wsmoses can you make the buildkite pipeline public?

Project.toml
```diff
@@ -32,12 +26,23 @@ ArrayInterface = "7.10"
...
+MLUtils = "f1d291b0-491e-4a28-83b9-f70985020b54"
+OneHotArrays = "0b1bfda6-eb8a-41d2-88d8-f5af5cad476f"
+Optimisers = "3bd65402-5787-11e9-1adc-39752487f4e2"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+ReTestItems = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["BenchmarkTools", "Flux", "Hwloc", "InteractiveUtils", "Lux", "MLUtils", "OneHotArrays", "Optimisers", "Random", "ReTestItems", "Statistics", "Test"]
```
mofeing, line 48: Why delete the `test/Project.toml` file and extend the `extras` and `targets` section?

Nowadays, it's better practice to use `test/Project.toml`.

deps/ReactantExtra/make.jl
mofeing, line 1: I'm okay with formatting this files, but they're unrelated to this PR and probably should be managed by the formatter workflow?

avik-pal: reverted. did not check the formatter only runs on src/test/ext

mofeing: Yeah, it has happened to me too. I think I need to update the formatter workflow to avoid these problems

Project.toml
avik-pal: That is one thing that `ReTestItems.jl` doesn't let you do https://github.com/JuliaTesting/ReTestItems.jl/issues/165.

Doesn't having separate Project file cause issues with merging projects when trying to install specific deps while testing? Since we might want to add CUDA/AMDGPU tests later on, that could be a problem, right?

wsmoses:
@avik-pal pipeline should be public now

Project.toml
mofeing: > That is one thing that `ReTestItems.jl` doesn't let you do [JuliaTesting/ReTestItems.jl#165](https://github.com/JuliaTesting/ReTestItems.jl/issues/165).

This seems like a bug they should fix.

> Doesn't having separate Project file cause issues with merging projects when trying to install specific deps while testing? Since we might want to add CUDA/AMDGPU tests later on, that could be a problem, right?

mmm I haven't heard anything around that. Do you have a link to an issue where they talk about it? What I usually do in these cases is never do a conditional installation, but include everything in the `Project.toml` and conditionally load the packages.

mofeing:
@avik-pal I have a couple of questions:
- does `TestItems` allow for some form of hierarchical organization like `@testset`?
- the parallelization is thread- or process-based? if it's thread-based, @wsmoses do we know if XLA compilation is thread-safe?

wsmoses:
Compilation ought be, but cpu is multi core so we shouldn’t do multiple
concurrent tests

avik-pal:
> does TestItems allow for some form of hierarchical organization like @testset?

ReTestItems' `@testitem` is a superset of the `@testitem` from `TestItems.jl`. Inside a `@testitem` you can have hierarchical `@testset`s. You can think of a `@testitem` as a module isolating the environment of the test.

> the parallelization is thread- or process-based? 

It is process-based.

Project.toml
avik-pal: > Do you have a link to an issue where they talk about it?

On checking it looks like the ones in Pkg.jl on that were closed (last one around Sept '23), so it has been fixed upstream.

avik-pal: This was my mistake in usage. Should be fixed now.

PR #58, opened 2024-08-04
Move shape information of `RArray` to field

mofeing:
Discussed in #52

@wsmoses Unfortunately, these commits are tangled with the move to the `Tracing.jl` file.

# Changes

- Shape information of an `RArray` will no longer be contained in a type parameter, but in a field.
- Move tracing-related code to `Tracing.jl` file.


wsmoses (approved):
This generally seems reasonable to me -- but would you be able to add the ci for benchmarking so we can test the compile time impact here?

mofeing:
This might be hard because what needs to be measured is only the compilation time of the first execution.

Also, how should the info be displayed? Just print the timings?

wsmoses:
@mofeing so it should be fairly easy to do this: https://github.com/EnzymeAD/Enzyme.jl/blob/main/benchmark/benchmarks.jl via https://github.com/EnzymeAD/Enzyme.jl/blob/main/.github/workflows/benchmark_pr.yml

which puts automatically a comment in the PR with the overhead change

wsmoses:
@mofeing can you rebase this and see if benchmarking runs?

mofeing:
I'm on it!

mofeing:
The workflow seems to work but it's BenchmarkTools which is giving me a small headache

mofeing:
Finally got it working.

@wsmoses I just added some basic benchmarks. Don't know if you want some more?

PR #59, opened 2024-08-05
Refactor `traced_type` to allow customization for external types

mofeing:
This PR depends on #58

~~DO NOT MERGE IT BEFORE IT~~


src/Tracing.jl
```diff
@@ -647,3 +647,430 @@ function _copyto!(dest::TracedRArray, bc::Broadcasted)
     dest.mlir_data = res.mlir_data
     return dest
 end
+
+@enum TraceMode begin
```
wsmoses, line 651: maybe even move this in a new file for TypeTracing.jl ?

mofeing: I guess you want to split `TracedRArray` tracing (what was in `overloads.jl`) and `traced_type` right?

wsmoses: oh I was just thinking about putting the function tracing in a diff file from the type tracing.

but that sounds wise/good too!

mofeing:
This PR is ready. The main changes introduced are:
- Move `TracedRArray`-related code to "TracedRArray.jl"
- Move `ConcreteRArray`-related code to "ConcreteRArray.jl"
- Move `make_tracer` and `traced_type` to "Tracing.jl"
- Split `traced_type` into multiple methods and rely on multiple dispatch

This last point can make `traced_type` a lil bit slower, due to dynamic dispatch, but it's not really a problem because the bottleneck doesn't seem to be there. The benchmarks results seem to show only noisy variations.

The refactor to rely on multiple dispatch is beneficial because it adds clarity to the code and is easier to customize this way.

In case we detect that dynamic dispatch is a performance issue, we could speed it up with something like ValSplit.jl

wsmoses (approved):

PR #60, opened 2024-08-05
Refactor compilation

mofeing:
Depends on #59

DON'T MERGE BEFORE IT

@wsmoses sorry i couldn't split the move to `Compiler.jl` and the proposal for `Thunk`

# Changes

- Move `compile`-related code to a new file `Compiler.jl`
- Introduce `traced_getfield` to allow customization of "trace-path deserialization"
- Test world-age issue
- Replace `Expr` serialization to types-parameters in `Thunk` for `RuntimeGeneratedFunctions`-esque `Thunk`s


src/utils.jl
```diff
@@ -112,12 +112,12 @@ function make_mlir_fn(f, args, kwargs, name="main", concretein=true; toscalar=fa
         seen_results, result, (:result,), concretein ? TracedTrack : TracedSetPath
     )
 
-    retraced_args = ntuple(Val(N)) do i
-        Base.@_inline_meta
-        return make_tracer(
-            seen_results, traced_args[i], concretein ? (:resargs, i) : (), TracedTrack
-        )
-    end
+    # retraced_args = ntuple(Val(N)) do i
```
wsmoses, line 115: this is still needed, no?

Otherwise we won't be able to tell that we can avoid computing an arg since it's the same in place?

test/basic.jl
```diff
@@ -9,9 +9,11 @@ fastmax(x::AbstractArray{T}) where {T} = reduce(max, x; dims=1, init=float(T)(-I
 using InteractiveUtils
 
 @testset "2D sum" begin
-    r_res = sum(ones(2, 10))
+    x = rand(2, 10)
```
wsmoses, line 12: We should keep non-random numbers here to keep things easier to debug. Happy if they're not all ones though

src/utils.jl
mofeing: It's unused. That's why I commented it.

test/basic.jl
mofeing: Maybe I can set the seed for `rand`?

wsmoses: I'm okay with that, but still would slightly prefer just a constant here (just as you move things around and add tests, it remains consistent and easy to pick out and debug a MWE from the test suite with exactly the same #'s).

src/utils.jl
wsmoses: but doesn't it modify the args in place [though sure the results may be unused]?

or am I misremembering

PR #61, opened 2024-08-05
Despecialize some methods to lower excessive codegen

mofeing:
Depends on #68 

~~DON'T MERGE BEFORE IT~~

It basically ads some `@nospecialize` annotations to some methods to lower the amount of codegen.

Also, it fixes a small performance bug (I wasn't reusing the `ReactantInterpreter`, so we weren't reusing the cache).

### DIFF
--- a/src/Tracing.jl
+++ b/src/Tracing.jl
@@ -217,7 +217,14 @@ end
 
 append_path(path, i) = (path..., i)
 
-function make_tracer(seen, prev::RT, path, mode; toscalar=false, tobatch=nothing) where {RT}
+function make_tracer(
+    seen,
+    @nospecialize(prev::RT),
+    @nospecialize(path),
+    mode;
+    toscalar=false,
+    tobatch=nothing,
+) where {RT}
     if haskey(seen, prev)
         return seen[prev]
     end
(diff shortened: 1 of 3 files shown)


src/Tracing.jl
```diff
@@ -122,7 +122,9 @@ for (jlop, hloop) in (
     (:(Base.:-), :subtract),
 )
     @eval begin
-        function $jlop(lhs::TracedRArray{T,N}, rhs::TracedRArray{T2,N}) where {T,T2,N}
+        function $jlop(
+            @nospecialize(lhs::TracedRArray{T,N}), @nospecialize(rhs::TracedRArray{T2,N})
```
wsmoses, line 126: if this is already a tracedrarray, is there a reason we need to nospecialize?

Since isn't this already as much info about the type as possible [and no more specialization is possible]

mofeing: Without `@nospecialize`, it would call codegen for each different combination of `T`, `T2` and `N` when we are just really just using them as runtime arguments to MLIR. We don't again anything from Julia's specializing compilation, so we avoid the cost of codegen for each different combination.

mofeing:
@wsmoses now this is ready too. Note that the big performance improvement comes from reusing the interpreter and not the de specialization (although it may help)

mofeing:
@wsmoses this PR is now ready. Note that the big compile time performance gain is due to reusing the `ReactantInterpreter`.

wsmoses (approved):
nice!

wsmoses:
CI fails tho so that should be fixed

mofeing:
Apparently, it fails on reusing the interpreter... gonna remove it from this PR and try fix it in another PR.

PR #65, opened 2024-08-13
Add benchmarking infrastructure

mofeing:


wsmoses (approved):

mofeing:
@wsmoses is that error on the benchmark workflow normal?

wsmoses:
@mofeing I think that may be expected to start with (since there's no benchmarks to check on the current repo).

PR #73, opened 2024-08-22
Fix broadcasting on `ConcreteRArray`

mofeing:


src/ConcreteRArray.jl
```diff
@@ -116,3 +116,36 @@ function Base.getindex(a::ConcreteRArray{T}, args::Vararg{Int,N}) where {T,N}
...
+    ElType = Base.Broadcast.combine_eltypes(bc.f, bc.args)
+    if !Base.isconcretetype(ElType)
+        throw(
+            ErrorException(
+                "`copy` on `ConcreteRArray` for non-concrete eltype is not implemented"
+            ),
+        )
+    end
+
+    aux = copyto!(similar(Array{ElType}, axes(bc)), bc)
```
wsmoses, line 149: eventually we may want this to compile the broadcast ops themselves, but we can do that later

wsmoses (approved):
LGTM with same add a test comment

PR #74, opened 2024-08-22
Small fixes on abstract interpreter

mofeing:


wsmoses (approved):
Add a test to prevent future regression?

mofeing:
> Add a test to prevent future regression?

It's a lil bit hard. What we need is a whole testset for `traced_getfield` and another for Enzyme (with some tests for custom types).

I would prefer to do it in another PR with the whole set of tests.

PR #77, opened 2024-08-22
Implement `only` for `TracedRArray`

mofeing:
@wsmoses In Julia, a 0-dim array is not (programmatically) equal to a scalar, but we're treating as such in Reactant. We should be more careful about this, but meanwhile, I'm using this method implementation to keep compatibility between Reactant and non-Reactant code.

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -29,7 +29,7 @@ function Base.show(io::IOty, X::TracedRArray{T,N}) where {T,N,IOty<:Union{IO,IOC
     return print(io, X.mlir_data, ")")
 end
 
-Base.only(A::TracedRArray) = A
+Base.only(A::TracedRArray{T,0}) where {T} = A
 
 function Base.reshape(A::TracedRArray{T,N}, dims::NTuple{NT,Int}) where {T,N,NT}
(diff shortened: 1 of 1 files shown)


wsmoses:
Yeah the issue is overloading or an acutual AbstractFloat requires a new type in addition to the RArray types, so we've been using the 0-dim for this for ease. We definitely should figure out a good general way to handle

src/TracedRArray.jl
```diff
@@ -29,6 +29,8 @@ function Base.show(io::IOty, X::TracedRArray{T,N}) where {T,N,IOty<:Union{IO,IOC
     return print(io, X.mlir_data, ")")
 end
 
+Base.only(A::TracedRArray) = A
```
wsmoses, line 32: Can we have this only apply for 0-dim arrays?

mofeing:
An alternative could be to implement something like `TrackedReal` in ReverseDiff.jl, but in the MLIR-side is still a 0-dim tensor.

```julia
struct TracedNumber{T} <: Number
    mlir_data
end
```

src/TracedRArray.jl
mofeing: done!

wsmoses (approved):

PR #71, opened 2024-08-22
ci(buildkite): fix scripts

avik-pal:

PR #85, opened 2024-08-28
perf: setup more detailed benchmarking

avik-pal:
- [ ] Add a llama model similar to Enzyme-JAX.
- [ ] Add `BUILDKITE_TOKEN` (`@wsmoses` needs to do this).
- [x] `ASV` specific tests for load times and overheads moved to a separate file.
- [ ] Bring over the benchmarking script from https://github.com/LuxDL/Lux.jl/blob/main/.buildkite/benchmarks.yml
  - Threading needs to be replaced with a comparison on different pipelines
- [ ] needs #71 for buildkite builds to work


avik-pal:
oops bad rebase

PR #86, opened 2024-08-28
perf: more detailed benchmarking

avik-pal:
- [ ] Add a llama model similar to Enzyme-JAX.
- [ ] Add `BUILDKITE_TOKEN` (`@wsmoses` needs to do this).
- [x] `ASV` specific tests for load times and overheads moved to a separate file.
- [ ] Bring over the benchmarking script from https://github.com/LuxDL/Lux.jl/blob/main/.buildkite/benchmarks.yml
  - Threading needs to be replaced with a comparison on different pipelines
- [ ] needs #71 for buildkite builds to work


avik-pal:
Some of the more complicated models are blocked by:
1. https://github.com/EnzymeAD/Reactant.jl/issues/107 -- prevents models carrying over state information (any normalization layers) from being compiled
2. Mixer models are blocked by https://github.com/EnzymeAD/Reactant.jl/issues/106

avik-pal:
Too much conflict. I will open a new PR based off of #130

PR #96, opened 2024-09-03
Add back support for buffer donation

mofeing:


test/buffer_donation.jl
```diff
@@ -0,0 +1,27 @@
...
+function donate_inplace_mul(x, y)
+    x .*= y
+    return nothing
+end
+
+@testset "buffer_donation" begin
+    a = Reactant.ConcreteRArray(ones(2, 2))
+    b = Reactant.ConcreteRArray(3 * ones(2, 2))
+    f = Reactant.compile(donate_fill_x_with_2, (a, b))
+    f(a, b)
```
wsmoses, line 19: Can we add a test to confirm the buffer semantics?

wsmoses: Specifically we'll need some utility like code_mhlo that lets us see what things are marked donated

wsmoses:
So we should explicitly check the optimized hlo that it doesn't have unnecessary inputs/outputs

We also need to find a way to check the donated info, which is here: https://github.com/EnzymeAD/Reactant.jl/blob/86707705adaa4aa3a743ebadf54db1878b7ca7c2/src/Compiler.jl#L453

test/buffer_donation.jl
mofeing: yeah, current impl of `code_hlo` doesn't show any attribute like that:
```julia
julia> Reactant.@code_hlo donate_fill_x_with_2(a,b)
Module:
  func.func @main(%arg0: tensor<2x2xf64>, %arg1: tensor<2x2xf64>) -> tensor<2x2xf64> {
    %cst = stablehlo.constant dense<2.000000e+00> : tensor<2x2xf64>
    return %cst : tensor<2x2xf64>
  }
}
```

PR #99, opened 2024-09-04
Automatically fetch and register PjRT Metal plugin

mofeing:
- [ ] Move artifact to a new `PjRT_Metal_jll` package
- [ ] Try local build

@wsmoses is there a Metal infra available for testing?

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -183,6 +183,32 @@ extern "C" PjRtClient* MakeTPUClient(const char* tpu_path , const char** error) 
     return GetCApiClient("TPU");
 }
 
+const char* const kEnvMetalLibraryPath = "METAL_LIBRARY_PATH";
+
+extern "C" PjRtClient* MakeMetalClient(const char* libpath, const char** error) {
+    // Prefer $METAL_LIBRARY_PATH if set
+    std::string metal_library_path;
+    if (auto path = llvm::sys::Process::GetEnv(kEnvMetalLibraryPath)) {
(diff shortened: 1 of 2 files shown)


wsmoses:
I think just macos CI should do it, no?

deps/ReactantExtra/API.cpp
```diff
@@ -183,6 +183,32 @@ extern "C" PjRtClient* MakeTPUClient(const char* tpu_path , const char** error)
     return GetCApiClient("TPU");
 }
 
+const char* const kEnvMetalLibraryPath = "METAL_LIBRARY_PATH";
+
+extern "C" PjRtClient* MakeMetalClient(const char* libpath, const char** error) {
```
wsmoses, line 188: Can we avoid making a separate api change atm since the oher loadpjrt plugins have an api presently?

Just for the sake of setting it up without a jll bump

avik-pal:
> I think just macos CI should do it, no?

We might want to use the buildkite ones here. The macos-large GH runners have GPUs but those are paid. The free ones don't have GPU.

Buildkite script for that https://github.com/LuxDL/MLDataDevices.jl/blob/b7e05516c0c0ec554259d90b5d620b6d0bf12154/.buildkite/testing.yml#L110-L134.

wsmoses:
Fair nuff/sgtm

wsmoses:
It should work fine on your Mac without this Pr

avik-pal:
We should also close this. the plugin registration code is already in reactant though we are waiting for fixes to the upstream plugin from apple

PR #102, opened 2024-09-05
Compile NNlib.maxpool and NNlib.meanpool

Pangoraw:


ext/ReactantNNlibExt.jl
```diff
@@ -109,4 +109,84 @@ function NNlib.conv(
...
+            push!(body, block)
+
+            body
+        end
+
+    attr = fill(Reactant.MLIR.IR.Attribute(init), unranked)
+    init_value = Reactant.MLIR.IR.result(
+        Reactant.MLIR.Dialects.stablehlo.constant(; value=attr)
+    )
+    reduction = Reactant.MLIR.Dialects.stablehlo.reduce_window(
```
wsmoses, line 164: Hm would a slice do here as well?

wsmoses (approved):

PR #111, opened 2024-09-19
ci(github-actions): update the formatter action

avik-pal:
the formatter hasn't been running for sometime

### DIFF
--- a/.github/workflows/format-pr.yml
+++ b/.github/workflows/format-pr.yml
@@ -26,7 +26,7 @@ jobs:
         shell: julia --color=yes {0}
         run: |
           using JuliaFormatter
-          format(["src", "ext", "test"])
+          format(".")
       - name: Create Pull Request
         id: cpr
         uses: peter-evans/create-pull-request@v6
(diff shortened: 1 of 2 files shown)


avik-pal:
something went wrong. reviewdog is supposed to comment here :cry:

avik-pal:
oh nvm it is supposed to fail https://github.com/julia-actions/julia-format/tree/v3/?tab=readme-ov-file#setup-workflow-with-julia-format-action

.github/workflows/format-pr.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         shell: julia --color=yes {0}
         run: |
           using JuliaFormatter
-          format(["src", "ext", "test"])
+          format(".")
```
mofeing, line 29: The reason why I explicitly add the folders is not to trigger this workflow when there are changes in non-code file (e.g. on `.github` or on README.md). But it's ok 👍

mofeing (approved):

PR #119, opened 2024-09-25
chore: testing Enzyme 0.13

avik-pal:
dropping 1.10- since Enzyme bumps it minimum version to 1.10


.github/workflows/CI.yml
```diff
@@ -21,9 +21,9 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.9'
-          - '1.10'
-          - ~1.11.0-0
```
mofeing, line 26: I prefer to have one per version, so I would manually put v1.10.

I'm ok with dropping v1.9.

wsmoses (approved):
looks good, but maybe we keep the names 1.10, 1.11, etc in CI for clarity?

.github/workflows/CI.yml
avik-pal: done

PR #126, opened 2024-09-26
Refactor `compile` and test buffer donation

mofeing:
This PR refactors `compile` in a series of smaller functions that makes it easier to test different parts of it. In particular, we use it to test buffer donation but could be used to test other parts as well.

The new functions are:
- `compile_mlir` This is the old `compile_to_module!` but added a new method such that it creates and activates the MLIR context by itself and returns the MLIR module.
  - This patter was used in several ocasions (i.e. `code_hlo` and `compile`) with slight variations.
  - We don't really need to store the MLIR context (so no need for `CompiledModule`).
  - Also the MLIR context is only needed in the beginning but it's activated through all `compile`... so this just activates the context where is needed.
- `compile_xla` Calls to `compile_mlir` and passes the result to XLA for compilation.
- `codegen_flatten` Generates the Julia code that unwraps the input arguments into the XLA buffers (i.e. the _linearization_).
  - The `_flatten` suffix is a reference to the `flatten` function passed to `register_pytree_node` in JAX, which has a similar functionality.
- `codegen_unflatten` Generates the Julia code that rewraps the processed XLA buffers into the result datatypes (i.e. the _delinearization_).
  - The `_unflatten` suffix is a reference to the `unflatten` function passed to `register_pytree_node` in JAX, which has a similar functionality.
- `codegen_xla_call` Generates the Julia code that calls the XLA-generated executable.

This way, buffer donation can just be using `compile_mlir` and then checking manually againts the returned `linear_args` and `preserved_args`.

The refactor of `compile` is due to the fact that I had to debug it many times, and it's hard to follow without more structure.

I'm trying to rebase #123 since it changed some stuff on `compile`.


src/Compiler.jl
```diff
@@ -1,3 +1,62 @@
+traced_getfield(obj, field) = Base.getfield(obj, field)
```
wsmoses, line 1: can we add an inline and nospecialize here?

src/Compiler.jl
```diff
@@ -1,3 +1,62 @@
...
+function create_result(tocopy::D, path, result_stores) where {K,V,D<:AbstractDict{K,V}}
+    elems = Expr[]
+    for (i, p) in enumerate(pairs(tocopy))
+        push!(elems, create_result(p, append_path(path, i), result_stores))
+    end
+    return :($D([$(elems...)]))
+end
+
+for T in [Int, AbstractFloat, AbstractString, Nothing, Type, Symbol]
+    @eval create_result(tocopy::$T, path, result_stores) = Meta.quot(tocopy)
```
wsmoses, line 57: why not just do 

create_result(tocopy::Union{Int, AbstractFloat, Nothing, Type, Symbol}, path, result_stores) = tocopy

src/Compiler.jl
wsmoses, line 1: Can we put this entire thing in a module Compile or something so then we don't accidentally export another compile command?

wsmoses (approved):
LGTM just ome minor nits

src/Compiler.jl
mofeing: sure

src/Compiler.jl
mofeing: Because it must return a `Expr` and `tocopy` is not.

mofeing: Ahh about the `Union`... I can do it that way.

PR #137, opened 2024-10-01
Add fallback gpu setindex

wsmoses:


mofeing:
mmm i see why we need `mysetindex!` here but you will need to override it's behavior on the interpreter or there would be an infinite recursion right?

wsmoses:
No because it should only ever be called with tracedrarray's

wsmoses:
@mofeing @avik-pal so this "works" up to the point that clearly compile doesn't handle varargs. Do you mind getting this to the endpoint?

I'm sadly stuckl staring at bazel build systems =/

wsmoses:
going to merge for now but we should fix this

PR #140, opened 2024-10-01
Optimize `to_rarray` to avoid recursion for simple inputs

### LINKED ISSUE #138: Shortcuts for `to_rarray`
In DifferentiationInterface, the most frequent inputs for `Reactant.to_rarray` will be `Number`s and `AbstractArray`s. For these types, we can short-circuit the recursive tracing and gain a lot of performance:

```julia
julia> using Reactant: ConcreteRArray, to_rarray

julia> using BenchmarkTools

julia> to_rarray_fast(x::Number) = x
to_rarray_fast (generic function with 2 methods)

julia> to_rarray_fast(x::AbstractArray) = ConcreteRArray(x)
to_rarray_fast (generic function with 2 methods)

julia> x0, x1 = 1.0, [1.0]
(1.0, [1.0])

julia> @btime to_rarray($x0);
  228.749 ns (9 allocations: 672 bytes)

julia> @btime to_rarray_fast($x0);
  2.271 ns (0 allocations: 0 bytes)

julia> @btime to_rarray($x1);
  1.058 μs (19 allocations: 1.36 KiB)

julia> @btime to_rarray_fast($x1);
  563.778 ns (4 allocations: 144 bytes)
```

Should I open a PR?

gdalle:
Fixes #138 by adding methods for `to_rarray` on `::Number`, `::AbstractArray{<:Number}` and `::ConcreteRArray`. Adds tests with JET to ensure type stability (which did not hold before).

Question: what should happen to a number going through `to_rarray`? If it keeps the same type, how can it be traced through the function?


src/Tracing.jl
```diff
@@ -445,3 +445,7 @@ end
 @inline function to_rarray(@nospecialize(x))
     return make_tracer(OrderedIdDict(), x, (), Reactant.ArrayToConcrete)
 end
+
+to_rarray(x::Number) = x  # TODO: should this be a `ConcreteRArray{_,0}`?
```
wsmoses, line 449: this depends on context. Do you want to trace the value, or not. Presently we assume no under the default semantics so this should be excluded

wsmoses: or rather, this is correct under current semantics**

src/Tracing.jl
```diff
@@ -445,3 +445,7 @@ end
...
+to_rarray(x::ConcreteRArray) = x
+to_rarray(x::AbstractArray{<:Number}) = ConcreteRArray(x)
```
wsmoses, line 451: cc @avik-pal can we make a union type of all the primitive types that Reactant supports conversion into (e.g. int, float, complex float, etc). We should then use this instead of Number here

src/Tracing.jl
mofeing: @avik-pal and I are discussing the semantics of 0-dim arrays on #141

mofeing:
Now that Adapt is a strong dependency, shouldn't we just overload `adapt`?

src/Tracing.jl
gdalle: In DifferentiationInterface there are `derivative` operators for which which we may want to trace a function with scalar input

wsmoses:
> Now that Adapt is a strong dependency, shouldn't we just overload `adapt`? I don't get the different between `to_rarray` and `adapt`.

Well to_rarray predates adapt as a strong dep [which we might remove if we can work around later], but also are we sure that adapt correctly deals with recursive data structures?

src/Tracing.jl
avik-pal: union over https://mlir.llvm.org/docs/Dialects/Builtin/#types? (atleast the ones that are in base)

avik-pal:
> Well to_rarray predates adapt as a strong dep [which we might remove if we can work around later], but also are we sure that adapt correctly deals with recursive data structures?

Not really. You need to opt-in for every custom structure

mofeing:
> > Well to_rarray predates adapt as a strong dep [which we might remove if we can work around later], but also are we sure that adapt correctly deals with recursive data structures?
> 
> Not really. You need to opt-in for every custom structure

But then, we could maybe overload `Adapt.adapt_structure(::Type{ConcreteRArray}, x)` to do what `to_rarray` is doing right now.

mofeing:
> > Now that Adapt is a strong dependency, shouldn't we just overload `adapt`? I don't get the different between `to_rarray` and `adapt`.
> 
> Well to_rarray predates adapt as a strong dep [which we might remove if we can work around later], ...

Yeah, but we introduced `to_rarray` just because we didn't want Adapt as a strong dep xD

src/Tracing.jl
wsmoses: For now the default tracer does not trace scalars, only arrays so as is this should be left alone [and can be amended down the line].

src/Tracing.jl
wsmoses: These ones specificaily: https://github.com/EnzymeAD/Reactant.jl/blob/626cb725c00cc96acedc13daa2ae2543f813bebc/src/XLA.jl#L213

We can then also amend https://github.com/EnzymeAD/Reactant.jl/blob/626cb725c00cc96acedc13daa2ae2543f813bebc/src/Tracing.jl#L383 to be is a subtype of Reactant.PrimitiveTypes

avik-pal: Added in #161

gdalle:
What should we do about this PR?

src/Tracing.jl
wsmoses: @gdalle I think we're just waiting for this to get addressed, then lgtm

avik-pal: It's merged but we still shouldn't do `to_rarray(x::Number) = x` because the semantics won't work out

mofeing:
> What should we do about this PR?

I guess we can talk about this on Wednesday. Would you mind putting it in the agenda?

PR #146, opened 2024-10-02
feat: generalize indexing to all wrappers

avik-pal:


wsmoses:
lgtm but maybe add a test?

PR #156, opened 2024-10-04
fix: broadcasted type casting

avik-pal:
## TODOs

- [x] Fix the main optimisers issue
  - [x] `Optimisers.update!`
  - [x] `Optimisers.update`
- [x] Add tests

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -29,9 +29,13 @@ function create_result(tocopy::T, path, result_stores) where {T}
 end
 
 function create_result(tocopy::ConcreteRArray{T,N}, path, result_stores) where {T,N}
-    restore = result_stores[path]
-    delete!(result_stores, path)
-    return :(ConcreteRArray{$T,$N}($restore, $(tocopy.shape)))
+    if haskey(result_stores, path)
+        restore = result_stores[path]
+        delete!(result_stores, path)
+        return :(ConcreteRArray{$T,$N}($restore, $(tocopy.shape)))
+    end
+    # We will set the data for this later
+    return :(ConcreteRArray{$T,$N}($(tocopy.data), $(tocopy.shape)))
 end
 
 function create_result(tocopy::Array{T,N}, path, result_stores) where {T,N}
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -29,6 +29,7 @@ function create_result(tocopy::T, path, result_stores) where {T}
 end
 
 function create_result(tocopy::ConcreteRArray{T,N}, path, result_stores) where {T,N}
+
     restore = result_stores[path]
     delete!(result_stores, path)
     return :(ConcreteRArray{$T,$N}($restore, $(tocopy.shape)))
@@ -67,7 +68,7 @@ function create_result(tocopy::D, path, result_stores) where {K,V,D<:AbstractDic
 end
 
 function create_result(
-    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol}, path, result_stores
+    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol,Bool}, path, result_stores
 )
     return Meta.quot(tocopy)
 end
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -71,7 +71,7 @@ function create_result(tocopy::D, path, result_stores) where {K,V,D<:AbstractDic
 end
 
 function create_result(
-    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol,Bool},
+    tocopy::Union{Integer,AbstractFloat,AbstractString,Nothing,Type,Symbol},
     path,
     result_stores,
 )
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -29,7 +29,6 @@ function create_result(tocopy::T, path, result_stores) where {T}
 end
 
 function create_result(tocopy::ConcreteRArray{T,N}, path, result_stores) where {T,N}
-
     restore = result_stores[path]
     delete!(result_stores, path)
     return :(ConcreteRArray{$T,$N}($restore, $(tocopy.shape)))
@@ -68,7 +67,9 @@ function create_result(tocopy::D, path, result_stores) where {K,V,D<:AbstractDic
 end
 
 function create_result(
-    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol,Bool}, path, result_stores
+    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol,Bool},
+    path,
+    result_stores,
 )
     return Meta.quot(tocopy)
 end
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -344,8 +344,8 @@ function (::TypeCast{T})(x::TracedRArray{T2,0}) where {T,T2}
     return promote_to(TracedRArray{T,0}, x)
 end
 
-elem_apply(::Type{T}, x::TracedRArray{T}) where {T} = x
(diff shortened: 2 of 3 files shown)


avik-pal:
```julia
julia> @code_hlo optimize=false bcast_type(ps.a)
Module:
module {
  func.func private @"Reactant.TypeCast{Float64}()_broadcast_scalar"(%arg0: tensor<f64>) -> tensor<f64> {
    %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
    return %1 : tensor<f64>
  }
  func.func @main(%arg0: tensor<10xf64>) -> (tensor<10xf64>, tensor<10xf64>) {
    %0 = stablehlo.transpose %arg0, dims = [0] : (tensor<10xf64>) -> tensor<10xf64>
    %1 = stablehlo.broadcast_in_dim %0, dims = [0] : (tensor<10xf64>) -> tensor<10xf64>
    %2 = enzyme.batch @"Reactant.TypeCast{Float64}()_broadcast_scalar"(%1) {batch_shape = array<i64: 10>} : (tensor<10xf64>) -> tensor<10xf64>
    %3 = stablehlo.transpose %2, dims = [0] : (tensor<10xf64>) -> tensor<10xf64>
    %4 = stablehlo.transpose %0, dims = [0] : (tensor<10xf64>) -> tensor<10xf64>
    return %3, %4 : tensor<10xf64>, tensor<10xf64>
  }
}

julia> @code_hlo bcast_type(ps.a)
Module:
  func.func @main(%arg0: tensor<10xf64>) {
    return
  }
}
```

This shouldn't be happening right? The output got optimized out

wsmoses:
in this case actually yes it should be happening, if we can prove that a returned value is just an input value we optimize it out and don't bother computing it (since it's already available as an input)

avik-pal:
shouldn't it be `return %arg0`?

wsmoses:
yeah but that got optimized out [specifically after optimizing it to become return arg0, we look for any returned values which are arguments and avoid shuffling them to xla entirely]

avik-pal:
This patch is sufficient for `Optimisers.update!` to work. But `Optimisers.update` causes:

```julia
ERROR: cannot copy false of type Bool
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] create_result(tocopy::Bool, path::Tuple{Int64, Int64, Int64}, result_stores::Dict{Tuple, Symbol})
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:18
  [3] create_result(tocopy::Optimisers.Leaf{…}, path::Tuple{…}, result_stores::Dict{…})
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:24
  [4] create_result(tocopy::@NamedTuple{…}, path::Tuple{…}, result_stores::Dict{…})
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:56
  [5] create_result(tocopy::Tuple{@NamedTuple{…}, @NamedTuple{…}}, path::Tuple{}, result_stores::Dict{Tuple, Symbol})
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:48
  [6] codegen_unflatten!(linear_args::Vector{…}, preserved_args::Vector{…}, concretized_res_names::Vector{…}, linear_results::Vector{…}, concrete_result::Tuple{…}, result_stores::Dict{…})
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:482
  [7] compile(f::Function, args::Tuple{@NamedTuple{…}, @NamedTuple{…}, @NamedTuple{…}}; client::Nothing)
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:621
  [8] compile(f::Function, args::Tuple{@NamedTuple{…}, @NamedTuple{…}, @NamedTuple{…}})
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:599
  [9] macro expansion
    @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:360 [inlined]
 [10] top-level scope
    @ /mnt/software/lux/Reactant.jl/envs/optimisers/simple.jl:11
```

avik-pal:
> yeah but that got optimized out [specifically after optimizing it to become return arg0, we look for any returned values which are arguments and avoid shuffling them to xla entirely]

I see what you mean. But this leads to:

```julia
julia> @compile bcast_type(ps.a)
ERROR: KeyError: key () not found
Stacktrace:
 [1] getindex
   @ ./dict.jl:498 [inlined]
 [2] create_result(tocopy::ConcreteRArray{Float64, 1}, path::Tuple{}, result_stores::Dict{Tuple, Symbol})
   @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:32
 [3] codegen_unflatten!(linear_args::Vector{…}, preserved_args::Vector{…}, concretized_res_names::Vector{…}, linear_results::Vector{…}, concrete_result::ConcreteRArray{…}, result_stores::Dict{…})
   @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:482
 [4] compile(f::Function, args::Tuple{ConcreteRArray{Float64, 1}}; client::Nothing)
   @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:621
 [5] compile(f::Function, args::Tuple{ConcreteRArray{Float64, 1}})
   @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:599
 [6] top-level scope
   @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:360
```

wsmoses:
here is the relevant code: https://github.com/EnzymeAD/Reactant.jl/blob/626cb725c00cc96acedc13daa2ae2543f813bebc/src/Compiler.jl#L288

clearly we don't sufficiently test for the special cases where this optimizations happen

src/Compiler.jl
```diff
@@ -67,7 +68,7 @@ function create_result(tocopy::D, path, result_stores) where {K,V,D<:AbstractDic
 end
 
 function create_result(
-    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol}, path, result_stores
+    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol,Bool}, path, result_stores
```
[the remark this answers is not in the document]
wsmoses: @avik-pal do we want to just have this be Integer instead of Int and Bool

avik-pal:
fixed it. will add some tests in the morning.

src/Compiler.jl
```diff
@@ -67,7 +71,9 @@ function create_result(tocopy::D, path, result_stores) where {K,V,D<:AbstractDic
 end
 
 function create_result(
-    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol}, path, result_stores
+    tocopy::Union{Int,AbstractFloat,AbstractString,Nothing,Type,Symbol,Bool},
```
wsmoses, line 74: can we just use integer here instead of Int, Bool

src/TracedRArray.jl
```diff
@@ -338,6 +338,18 @@ for (jlop, hloop) in (
...
 end
 
+struct TypeCast{T<:Number} <: Function end
+
+function (::TypeCast{T})(x::TracedRArray{T2,0}) where {T,T2}
+    return promote_to(TracedRArray{T,0}, x)
+end
+
+elem_apply(::Type{T}, x::TracedRArray{T}) where {T} = x
+function elem_apply(::Type{T}, x::TracedRArray{T2}) where {T,T2}
```
wsmoses, line 348: this doesn't have the number restriction that Typecast does

wsmoses (approved):
lgtm with a minor nit

PR #161, opened 2024-10-04
feat: implement a separate `TracedRNumber`

avik-pal:
~currently very WIP~


src/Reactant.jl
```diff
@@ -8,6 +8,7 @@ include("OrderedIdDict.jl")
 using Enzyme
 
 abstract type RArray{T,N} <: AbstractArray{T,N} end
+abstract type RScalar{T} <: Number end
```
mofeing, line 11: maybe let's call it `RNumber` to be as similar to Julia as posible?
```suggestion
abstract type RNumber{T} <: Number end
```

src/TracedRArray.jl
```diff
@@ -17,9 +17,38 @@ mutable struct TracedRArray{T,N} <: RArray{T,N}
...
 end
 
+function Base.setproperty!(x::TracedRArray, f::Symbol, v)
+    if f === :mlir_data && !isnothing(v)
+        @assert size(MLIR.IR.type(v)) == size(x)
+    end
+    return setfield!(x, f, v)
+end
+
+mutable struct TracedRScalar{T} <: RScalar{T}
```
mofeing, line 27:
```suggestion
mutable struct TracedRNumber{T} <: RNumber{T}
```

src/TracedRArray.jl
```diff
@@ -760,33 +655,3 @@ function _copyto!(dest::TracedRArray, bc::Broadcasted)
...
-        ),
-        shape,
-    )
-    return Res
-end
-
-function maybe_expand_dims(x::AbstractArray{T,N}, ::Val{D}) where {T,N,D}
-    D ≤ N && return x
-    return reshape(x, ntuple(i -> i ≤ N ? size(x, i) : 1, Val(D)))
-end
```
mofeing, line 792: mmm don't we still need this for `cat` methods on `TracedRArrays`?

avik-pal: just moved up, outside the broadcasting part

mofeing:
Let me just merge #163 after CI checks pass and rebase this PR on top of it because it adds some tests that should be passed by `TracedRNumbers` too.

src/Reactant.jl
```diff
@@ -7,7 +7,26 @@ include("OrderedIdDict.jl")
...
+    UInt8,
+    Int16,
+    UInt16,
+    Int32,
+    UInt32,
+    Int64,
+    UInt64,
+    Float16,
+    Float32,
+    # BFloat16,
```
wsmoses, line 22: Can we have a static check if defined and use or not? Eg https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/typetree.jl#L82

wsmoses (approved):
Lgtm only minor comment: would defining traced value = union{tracedrarrsy, travednumber} simplify some things

avik-pal:
Do we also need a ConcreteRScalar? If we want to extend `to_rarray` to all of ReactantPrimitives

wsmoses:
For now we can leave concrete as is

mofeing:
I don't think we need a concrete scalar type, because indexing on a concrete type is just a normal number.

avik-pal:
> I don't think we need a concrete scalar type, because indexing on a concrete array is just a normal number.

I meant more from a tracing perspective. What should `to_rarray(::ReactantPrimitives)` return? If it is `Number` then we can't convert it to a `TracedRNumber`. If we want to convert it then we need something like a ConcreteRNumber so that users are aware that it is going to be traced.

wsmoses:
So I think this is two different questions.

1) Should we upgrade scalars into traced values (like we do for arrays) or compile with respect to them (like we do constants).

For now I'm fine leaving them as is which is compile as constants

2) Can we represent a ConcreteScalar, if desired. I think this will be necessary in the future.

For example we may want to have a function with a user defined index offset. 

e.g.
``` 
A = ConcreteRarray
v = ConcreteScalar
f = @compile getindex(A, v)
```

without a concrete scalar we need to recompile for each index. I do think we need such a scalar, even if the default behavior for conversion is only converting arrays

wsmoses:
Even more simply though, we may need as the return type of a function with array inputs and a scalar output.

i.e.

```
x = ConcreteArray(ones(4,4))
f = @compile sum(x)
f(x) # this should return a scalar now, but we cannot actually represent this as a regularfloat64, since it will be the result of a future/etc

**Comment by avik-pal** [comment]
Looking at the above example I also realized the mapreduce semantics is incorrect at-present. For example `sum(x)` should return `TracedRNumber` and not `TracedRArray{T,0}`

> Even more simply though, we may need as the return type of a function with array inputs and a scalar output.

We could store it like an array in the ConcreteRArray but expose it to the enduser as a scalar. https://github.com/JuliaGPU/GPUArrays.jl/pull/550 is something similar in the GPUArrays world. But let's keep this PR simple and avoid this for now

File: src/utils.jl [hunk 5]
```diff
@@ -121,10 +130,10 @@ function make_mlir_fn(f, args, kwargs, name="main", concretein=true; toscalar=fa
         )
     end
 
-    linear_results = TracedRArray[]
+    linear_results = TracedTypes[]
 
     for (k, v) in seen_results
-        if !(v isa TracedRArray)
+        if !(v isa TracedRArray) && !(v isa TracedRNumber)
```
wsmoses, line 136: this should also just be a if not a tracedType

wsmoses (approved):
a small comment but otherwise lgtm

mofeing:
now that #163 has been merged, we should test concatenation on numbers.

@avik-pal should I do it or do you prefer to do it yourself?

mofeing:
There's a bug on concatenation of `TracedRNumber`. In particular, if we run this kernel:

```julia
using Reactant

x = fill(true)
x_concrete = Reactant.to_rarray(x)

function traced_vcat(x)
	a = x[]
	[a; a; a]
end

f = @compile traced_vcat(x_concrete)

f(x_concrete)
```

It fails with the following error:

```julia
ERROR: BoundsError: attempt to access 3-element Vector{ConcreteRArray{Bool}} at index [1]
Stacktrace:
 [1] traced_getfield
   @ ~/Developer/Reactant.jl/src/Compiler.jl:18 [inlined]
 [2] macro expansion
   @ ~/Developer/Reactant.jl/src/Compiler.jl:649 [inlined]
 [3] (::Reactant.Compiler.Thunk{Symbol("##test_vcat_reactant#229")})(args::ConcreteRArray{Bool, 0})
   @ Reactant.Compiler ~/Developer/Reactant.jl/src/Compiler.jl:665
 [4] top-level scope
   @ REPL[10]:1
```

The error seems to be that it's not doing any XLA call and it's setting `result` returning to 3 empty buffers. Check out the generated Julia code of `f`:

```julia
quote
    #= /Users/mofeing/Developer/Reactant.jl/src/Compiler.jl:639 =#
    #= /Users/mofeing/Developer/Reactant.jl/src/Compiler.jl:640 =#
    nothing
    #= /Users/mofeing/Developer/Reactant.jl/src/Compiler.jl:647 =#
    usbuf_1 = (getindex(args, 1)).data
    sbuf_1 = XLA.synced_buffer(usbuf_1)
    #= /Users/mofeing/Developer/Reactant.jl/src/Compiler.jl:648 =#
    ()
    #= /Users/mofeing/Developer/Reactant.jl/src/Compiler.jl:649 =#
    result = (ConcreteRArray{Bool})[ConcreteRArray{Bool, 0}(Reactant.XLA.AsyncBuffer(Reactant.XLA.Buffer(Ptr{Nothing} @0x0000000000000000), nothing), ()), ConcreteRArray{Bool, 0}(Reactant.XLA.AsyncBuffer(Reactant.XLA.Buffer(Ptr{Nothing} @0x0000000000000000), nothing), ()), ConcreteRArray{Bool, 0}(Reactant.XLA.AsyncBuffer(Reactant.XLA.Buffer(Ptr{Nothing} @0x0000000000000000), nothing), ())]
    (traced_getfield(result, $(Expr(:quote, 1)))).data = (args[1]).data
    (traced_getfield(result, $(Expr(:quote, 2)))).data = (args[1]).data
    (traced_getfield(result, $(Expr(:quote, 3)))).data = (args[1]).data
    #= /Users/mofeing/Developer/Reactant.jl/src/Compiler.jl:650 =#
    return result
end
```

wsmoses:
That actually makes sense here since it’s making an array of values all
available as args.

What’s the issue?

mofeing:
> That actually makes sense here since it’s making an array of values all available as args. What’s the issue?

I don't get it. It's behavior should like the one on 0-dim array test since we are passing a 0-dim array and should return a vector.

From the MLIR point of view, it should be calling `stablehlo.concatenate` on a 0-dim tensor.

wsmoses:
ah I see, I think this is a bug on the new number and cat perhaps?

avik-pal:
I think we won't do any XLA call here, see the julia fallbacks for number:

```julia
vcat(X::Number...) = hvcat_fill!(Vector{promote_typeof(X...)}(undef, length(X)), X)
hcat(X::Number...) = hvcat_fill!(Matrix{promote_typeof(X...)}(undef, 1,length(X)), X)
```

It is going to fill into a regular array

avik-pal:
Should work now

```julia
julia> function traced_vcat(x)
               a = x[]; 
               Float64[a; a; a]
       end
traced_vcat (generic function with 1 method)

julia> @code_hlo optimize=false traced_vcat(x_concrete)
Module:
module {
  func.func @main(%arg0: tensor<i1>) -> (tensor<3xf64>, tensor<i1>) {
    %0 = stablehlo.transpose %arg0, dims = [] : (tensor<i1>) -> tensor<i1>
    %1 = stablehlo.broadcast_in_dim %0, dims = [] : (tensor<i1>) -> tensor<1xi1>
    %2 = stablehlo.broadcast_in_dim %0, dims = [] : (tensor<i1>) -> tensor<1xi1>
    %3 = stablehlo.broadcast_in_dim %0, dims = [] : (tensor<i1>) -> tensor<1xi1>
    %4 = stablehlo.convert %1 : (tensor<1xi1>) -> tensor<1xf64>
    %5 = stablehlo.convert %2 : (tensor<1xi1>) -> tensor<1xf64>
    %6 = stablehlo.convert %3 : (tensor<1xi1>) -> tensor<1xf64>
    %7 = stablehlo.concatenate %4, %5, %6, dim = 0 : (tensor<1xf64>, tensor<1xf64>, tensor<1xf64>) -> tensor<3xf64>
    %8 = stablehlo.transpose %7, dims = [0] : (tensor<3xf64>) -> tensor<3xf64>
    %9 = stablehlo.transpose %0, dims = [] : (tensor<i1>) -> tensor<i1>
    return %8, %9 : tensor<3xf64>, tensor<i1>
  }
}

julia> @code_hlo traced_vcat(x_concrete)
Module:
  func.func @main(%arg0: tensor<i1>) -> tensor<3xf64> {
    %0 = stablehlo.convert %arg0 : (tensor<i1>) -> tensor<f64>
    %1 = stablehlo.broadcast_in_dim %0, dims = [] : (tensor<f64>) -> tensor<3xf64>
    return %1 : tensor<3xf64>
  }
}
```

mofeing (approved):

PR #164, opened 2024-10-04
Write Bazel rule for IFRT dialect

mofeing:
Currently it gives this error:

```
external/xla/xla/python/ifrt/ir/ifrt_ops.td:24:9: error: Could not find include file 'xla/python/ifrt/ir/ifrt_dialect.td'
include "xla/python/ifrt/ir/ifrt_dialect.td"
```

I believe that it's missing a include path flag.

### DIFF
--- a/deps/ReactantExtra/BUILD
+++ b/deps/ReactantExtra/BUILD
@@ -514,6 +514,20 @@ gentbl_cc_library(
     tblgen = "//:mlir-jl-tblgen",
 )
 
+gentbl_cc_library(
+    name = "IFRTJLIncGen",
+    tbl_outs = [(
+            ["--generator=jl-op-defs", "--disable-module-wrap=0"],
+            "IFRT.inc.jl"
+        )
+    ],
+    td_file = "@xla//xla/python/ifrt/ir:ifrt_ops.td",
+    deps = [
+        "@xla//xla/python/ifrt/ir:ifrt_td",
+    ],
+    tblgen = "//:mlir-jl-tblgen",
+)
+
 genrule(
     name = "libMLIR_h.jl",
     tags = [


mofeing:
i added some utilities for Julia-like conversion in C++ and better communication between Julia and C.

- `Type<T>` is just a placeholder for representing a target type
- `convert` converts the passing object to the object we can pass through FFI (it helps a lot deduplicating the code and avoiding bugs)
- `span<T>` is a small `std::span`-like object but for C++ <20 and ABI-stable so we can view it from Julia as a `Tuple{Csize, Ptr{Cvoid}}`

for example, imagine a function that returns a `absl::Span<tsl::RCReference<Device>>`. then our C binding would be like:
```c++
extern "C" span<Device*> the_function(...) {
    return convert(Type<span<Device*>>(), some_func_that_returns_a_absl_Span(...))
}
```

conceptually works, in practice a need to add some particular implementations so it works (i'm sure it can be done more generically but i don't have the time nor the energy to learn the terribly complicated C++)

mofeing:
@ftynse given this is a priority for the distributed working group, do you mind adding this to https://github.com/PRONTOLab/GB-25/issues/1?

ftynse:
Sure, we are also still figuring out the project management stack

mofeing (commented):
build fails, so it's not yet ready to merge but it can be reviewed while we fix it

deps/ReactantExtra/src/error_handling.hpp
```diff
@@ -0,0 +1,20 @@
...
+    if (ReactantThrowError) {
+        if (!v.ok()) {
+            ReactantThrowError(v.status().ToString().c_str());
+            throw xla::XlaRuntimeError(v.status().ToString().c_str());
+        }
+        return std::move(v).value();
+    } else {
+        return xla::ValueOrThrow(std::move(v));
+    }
+}
```
giordano, line 20: Missing newline
```suggestion
}

```

deps/ReactantExtra/src/ifrt/Value.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+extern "C" bool ifrt_value_is_deleted(Value* value)
+{
+    return value->IsDeleted();
+}
+
+extern "C" const char* ifrt_value_debug_string(Value* value)
+{
+    return convert(Type<const char*>(), value->DebugString());
+}
```
giordano, line 30: Missing newline
```suggestion
}

```

src/IFRT/IFRT.jl
```diff
@@ -0,0 +1,801 @@
...
+
+# Backend trait
+abstract type Backend end
+
+"""
+    backend(x)
+
+Returns the `Backend` trait of the given IFRT object.
+"""
+function backend end
```
giordano, line 22:
```
│ 
│     Reactant.IFRT.backend
│ 
```

giordano (commented):
As mentioned already on Slack, the changes to ReactantExtra should be done in a separate PR, to be merged before this done, to allow a build on Yggdrasil

deps/ReactantExtra/src/error_handling.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "absl/status/statusor.h"
+#include "xla/pjrt/exceptions.h"
+#include "xla/pjrt/status_casters.h"
+
+extern "C" extern void (*ReactantThrowError)(const char *);
```
giordano, line 7: Locally I get
````
[05:42:43] ERROR: /workspace/srcdir/Reactant.jl/deps/ReactantExtra/BUILD:355:11: Compiling src/ifrt/DynamicShape.cpp failed: (Exit 1): cc failed: error executing command (from target //:ReactantExtraLib)
[...]
[05:42:43] In file included from src/ifrt/DynamicShape.cpp:2:
[05:42:43] ./src/error_handling.hpp:7:12: error: invalid use of ‘extern’ in linkage specification
[05:42:43]     7 | extern "C" extern void (*ReactantThrowError)(const char *);
[05:42:43]       |            ^~~~~~
[05:42:44] [8,094 / 8,136] Compiling src/enzyme_ad/jax/TransformOps/TransformOps.cpp; 20s processwrapper-sandbox ... (41 actions running)
[05:42:45] [8,104 / 8,136] Compiling src/ifrt/Client.cpp; 12s processwrapper-sandbox ... (31 actions running)
[05:42:46] Target //:libReactantExtra.so failed to build
```
I presume this should be
```suggestion
extern "C" void (*ReactantThrowError)(const char *);
```
?

mofeing: i remember i wrote like that before but failed (on aarch64-macos). let me try again first locally.

src/IFRT/IFRT.jl
mofeing: mmm weird? there's nothing fancy around this. do you have an idea of what can be happening?

giordano: You simply need to add this docstring to the documentation

deps/ReactantExtra/src/error_handling.hpp
giordano: This is inconsistent with https://github.com/EnzymeAD/Reactant.jl/blob/e1e62cf654ec6a07666451d2819fd8de684c2d6c/deps/ReactantExtra/src/error_handling.cpp#L1

> also, the build succeeds on aarch64-macos

Not very useful to say "another compiler is happy with broken stuff", is it? :slightly_smiling_face:

src/IFRT/IFRT.jl
mofeing: ah, misunderstood the problem

giordano:
More errors in BB:
```
[07:55:57] ERROR: /workspace/srcdir/Reactant.jl/deps/ReactantExtra/BUILD:355:11: Compiling src/ifrt/Program.cpp failed: (Exit 1): aarch64-apple-darwin20-cc failed: error executing command (from target //:ReactantExtraLib)
[...]
[07:55:57] # Configuration: 733f0a17cb611a25065571f976b63834782b2e1741e54e6f530060798c5acd3e
[07:55:57] # Execution platform: @local_execution_config_platform//:platform
[07:55:57]
[07:55:57] Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
[07:55:57] In file included from src/ifrt/Program.cpp:1:
[07:55:57] In file included from ./src/type_conversion.hpp:8:
[07:55:57] ./src/memory_management.hpp:144:29: error: no member named 'reinterpret_pointer_cast' in namespace 'std'
[07:55:57]   144 |         capture_shared(std::reinterpret_pointer_cast<void>(std::const_pointer_cast<G>(ptr)))
[07:55:57]       |                        ~~~~~^
[07:55:57] ./src/memory_management.hpp:144:58: error: expected '(' for function-style cast or type construction
[07:55:57]   144 |         capture_shared(std::reinterpret_pointer_cast<void>(std::const_pointer_cast<G>(ptr)))
[07:55:57]       |                                                      ~~~~^
[07:55:57] ./src/memory_management.hpp:175:17: error: no member named 'reinterpret_pointer_cast' in namespace 'std'
[07:55:57]   175 |     return std::reinterpret_pointer_cast<T>(get_shared(ptr));
[07:55:57]       |            ~~~~~^
[07:55:57] ./src/memory_management.hpp:175:42: error: 'T' does not refer to a value
[07:55:57]   175 |     return std::reinterpret_pointer_cast<T>(get_shared(ptr));
[07:55:57]       |                                          ^
[07:55:57] ./src/memory_management.hpp:171:19: note: declared here
[07:55:57]   171 | template<typename T>
[07:55:57]       |                   ^
[07:55:57] 4 errors generated.
```

deps/ReactantExtra/src/error_handling.hpp
giordano: > and `extern` is used to instruct that the symbol is declared in another object file, so don't declare it.

But this is a header file and the implementation is in the `.cpp` file.  Why do you need to mark the symbol as `extern` in a header?

mofeing:
mmm `std::reinterpret_pointer_cast` was introduced in C++17, which I think is the version we are using in ReactantExtra. do the logs show the used compiler flags?

deps/ReactantExtra/src/error_handling.hpp
mofeing: look that `MyValueOrThrow` is a template function, so it must go on the header and it calls `ReactantThrowError`. all files that use `MyValueOrThrow` must know about it

giordano: This isn't about `MyValueOrThrow` but `ReactantThrowError`.

mofeing:
> Sounds like `std::reinterpret_pointer_cast` was introduced in SDK 11.3, based on a [research](https://github.com/search?q=repo%3Aphracker%2FMacOSX-SDKs+reinterpret_pointer_cast+path%3A%2F%5E.*%5C%2Fusr%5C%2Finclude%5C%2Fc%5C%2B%5C%2B%2F&type=code) in https://github.com/phracker/MacOSX-SDKs and https://github.com/roblabla/MacOSX-SDKs.

> [#164 (comment)](https://github.com/EnzymeAD/Reactant.jl/pull/164#discussion_r1946405813) still needs to be resolved though.

sure. If it's not giving linking problems, I can set it as just `extern "C"`.

giordano:
> i'm still having this issue though when building it with clang
> 
> ```
> dlopen(/Users/mofeing/Developer/Reactant.jl/deps/ReactantExtra/bazel-bin/libReactantExtra.so, 0x0001): symbol not found in flat namespace '__ZNSt13exception_ptr31__from_native_exception_pointerEPv'
> ```
> 

I don't see that symbol in the BinaryBuilder build:

```console
sandbox:${WORKSPACE} # c++filt -_ __ZNSt13exception_ptr31__from_native_exception_pointerEPv
std::exception_ptr::__from_native_exception_pointer(void*)
sandbox:${WORKSPACE} # nm libReactantExtra.dylib | grep __ZNSt13exception_ptr31__from_native_exception_pointerEPv
sandbox:${WORKSPACE} # nm libReactantExtra.dylib | grep exception | /opt/${MACHTYPE}/bin/${MACHTYPE}-c++filt -_
000000000687eec8 t llvm::ARMAttributeParser::ABI_FP_exceptions(llvm::ARMBuildAttrs::AttrType)
000000000687eee4 t llvm::ARMAttributeParser::ABI_FP_user_exceptions(llvm::ARMBuildAttrs::AttrType)
                 U std::exception::what() const
                 U std::exception_ptr::exception_ptr(std::exception_ptr const&)
                 U std::exception_ptr::~exception_ptr()
                 U std::exception_ptr::operator=(std::exception_ptr const&)
                 U std::__1::__assoc_sub_state::set_exception(std::exception_ptr)
                 U std::exception::~exception()
                 U std::current_exception()
                 U std::rethrow_exception(std::exception_ptr)
0000000001d6cf58 t std::exception_ptr std::make_exception_ptr<std::__1::future_error>(std::__1::future_error)
0000000001d073cc t std::exception_ptr std::make_exception_ptr<std::__1::future_error>(std::__1::future_error)
                 U typeinfo for std::exception
000000000845cee0 s llvm::ARMAttributeParser::ABI_FP_exceptions(llvm::ARMBuildAttrs::AttrType)::strings
000000000845cef0 s llvm::ARMAttributeParser::ABI_FP_user_exceptions(llvm::ARMBuildAttrs::AttrType)::strings
                 U ___cxa_allocate_exception
                 U ___cxa_free_exception
                 U _task_get_exception_ports
                 U _task_set_exception_ports
```

Maybe it's a problem with your toolchain?

src/IFRT/IFRT.jl
giordano: Bump

mofeing:
> Maybe it's a problem with your toolchain?

probably, just found out that it links to `/usr/lib/libc++.1.dylib` but... there is no such file 🫠

this would explain the "mixing between libstdc++ and libc++": there's no such mix, it's just that in my local build, it expects libc++ to provide that symbol and can't link it because it doesn't find it.

i was building with Homebrew's LLVM 19, but trying again with LLVM 16.

giordano:
> probably, just found out that it links to `/usr/lib/libc++.1.dylib` but... there is no such file 🫠

That's normal, macOS hides system libraries and header files from the filesystem: https://en.wikipedia.org/wiki/System_Integrity_Protection

giordano:
Also: https://github.com/EnzymeAD/Reactant.jl/pull/164#pullrequestreview-2601516718

> As mentioned already on Slack, the changes to ReactantExtra should be done in a separate PR, to be merged before this done, to allow a build on Yggdrasil

:point_up:

mofeing:
> As mentioned already on Slack, the changes to ReactantExtra should be done in a separate PR, to be merged before this one, to allow a build on Yggdrasil

i'm leaving this PR just for the changes in ReactantExtra and gonna move the Julia part of the bindings to a new PR

> Maybe it's a problem with your toolchain?

it seems like so. using `--cc gcc --gcc_host_compiler_path gcc` makes the error disappear. still trying to understand what happened these last months because it worked before, but it could be on XLA's side and not on ours.

deps/ReactantExtra/src/memory_management.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }, storage);
+}
+
+void RCRef::destroy() noexcept {
+    storage = std::monostate{};
+}
+
+template<>
+void* capture_shared(std::shared_ptr<void> ptr) {
+    captured_shared_ptr[ptr.get()] = ptr;
```
wsmoses, line 56: we should use the new wrapper type as we talked about earlier (the global can work, but will cause concurrency issues that won't happen with the wrapper)

deps/ReactantExtra/src/memory_management.hpp
```diff
@@ -0,0 +1,186 @@
...
+    void destroy() noexcept;
+
+    template<typename T> T* get() const noexcept;
+    template<typename T> tsl::RCReference<T> get_rcref() const noexcept;
+    template<typename T> bool is() const noexcept; 
+
+private:
+    // TODO can we avoid using `Array` here and cast from/to `Value`?
+    using storage_t = std::variant<
+        std::monostate,
```
wsmoses, line 48: don't make this a std::variant. Make 

```
template<typename T>
class RCRef {
....
   tcl::RCReference<T> storage
}
```

wsmoses (commented):
This is a lot to review. Can you separate this into two PR's

The first, a non-functinoal change which moves around the code into separate files, and the second which adds the new functions?

deps/ReactantExtra/src/memory_management.hpp
mofeing: the reason for this is to try unify all captured `tsl::RCReference` into one map. `std::shared_ptr` allows this by casting from/to `std::shared_ptr<void>`, but for `tsl::RCReference` I'm not sure about the supported casting. it allows to cast from derived to base classes (e.g. `Tuple`, `Array` to `Value`) but not sure about the other way around. also, very few classes inherit from `ReferenceCounted<T>` which is the condition to be managed by `tsl::RCReference`.

wsmoses: you shouldn't need a map though if you use the wrapper struct?

Variant is also going to be a lot slower here than separate types (in essence its akin to type instability in julia)

deps/ReactantExtra/src/memory_management.cpp
mofeing: sure, using the new wrapper type solves concurrency issues but AFAIK the logic in Reactant.jl works on single thread as of today right?

since this seems to work as of now, i wanted to get this merged first and then switch to the new holder types on the next PR.

wsmoses: no, Reactant.jl logic isn't limited to a single thread (and user code can allocate in multiple threads)

mofeing: okay, then let me prototype sth

PR #167, opened 2024-10-05
test: separate out test groups

avik-pal:


avik-pal:
much more manageable to look at

wsmoses:
indeed! separately if you can open a mwe for that failure on enzyme proper we can also get that fixed

PR #178, opened 2024-10-15
Add IFRT C-API bindings

mofeing:


deps/ReactantExtra/API.cpp
```diff
@@ -471,3 +524,863 @@ extern "C" void InitializeRegistryAndPasses(MlirDialectRegistry creg) {
...
+
+extern "C" bool ifrt_memorykind_ne(ifrt::MemoryKind* mk1, ifrt::MemoryKind* mk2) {
+    return *mk1 != *mk2;
+}
+
+extern "C" const char* ifrt_memorykind_string(ifrt::MemoryKind* memory_kind) {
+    if (memory_kind->memory_kind().has_value())
+        return cstr_from_string(memory_kind->memory_kind().value());
+    else
+        return "";
```
wsmoses, line 821: returning null here probably is useful to distinguish from a set memory kind to empty string?

mofeing: I think is unprovable to find a memory kind set to the empty string but yeah, it's better to return a nullptr since the C++ version returns a `optional`.

mofeing:
```suggestion
        return nullptr;
```

PR #185, opened 2024-10-19
feat: support conditionals

avik-pal:
```julia
using Reactant

function conditional_fn(x)
    v = sum(x)
    ff = -1.0
    @trace if v > 0
        z = 1.0
        v2 = sum(x)
        x = x .+ ff
    # elseif v > -2.0
    #     z = 2.0
    # elseif v > -4.0
        # z = 3.0
    else
        z = 0.0
        v2 = -1.0
    end
    return z
end

conditional_fn(rand(2, 3))
```

## TODOs

- [x] general variable tracking that can be used for loops later on
- [x] rewrite the julia expression into something that can be traced
- [x] tracing overload for `ControlFlow.traced_if`
- [x] can we rewrite it to an expression without perf penalty ?? Currently leads to Boxing
- [x] support elseif blocks
- [x] different variables defined in each block
  - [x] nice to have but additionally helps support blocks without an else
- [x] tests
- [x] move into `ReactantCore`?
- [x] CI scripts to install ReactantCore


avik-pal:
```julia
julia> @macroexpand     @trace if v > 0
               z = 1.0
               v2 = sum(x)
               x = x .+ ff
           # elseif v > -2.0
           #     z = 2.0
           # elseif v > -4.0
               # z = 3.0
           else
               z = 0.0
               v2 = -1.0
           end
quote
    #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:55 =#
    begin
        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:37 =#
        var"##true_branch#292" = let x = x, ff = ff
                #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:38 =#
                (x, ff)->begin
                        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:38 =#
                        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:39 =#
                        begin
                            #= REPL[40]:2 =#
                            z = 1.0
                            #= REPL[40]:3 =#
                            v2 = sum(x)
                            #= REPL[40]:4 =#
                            x = x .+ ff
                        end
                        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:40 =#
                        return (z, v2, x)
                    end
            end
    end
    #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:56 =#
    begin
        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:46 =#
        var"##false_branch#293" = let x = x, ff = ff
                #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:47 =#
                (x, ff)->begin
                        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:47 =#
                        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:48 =#
                        begin
                            #= REPL[40]:10 =#
                            z = 0.0
                            #= REPL[40]:11 =#
                            v2 = -1.0
                        end
                        #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:49 =#
                        return (z, v2, x)
                    end
            end
    end
    #= /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:57 =#
    (z, v2, x) = (Reactant.ControlFlow.custom_if_condition)(v > 0, var"##true_branch#292", var"##false_branch#293", (x, ff))
end
```

avik-pal:
```julia
using Reactant

function conditional_fn(x)
    v = sum(x)
    ff = -1.0
    @trace if v > 0
        z = 1.0
        v2 = sum(x)
        # x = x .+ ff
    # elseif v > -2.0
    #     z = 2.0
    # elseif v > -4.0
        # z = 3.0
    else
        z = 0.0
        v2 = -sum(x)
    end
    return z, v2
end

conditional_fn(rand(2, 3))

x = rand(2, 3)

x_ra = Reactant.to_rarray(x)

conditional_fn_compiled = @compile conditional_fn(x_ra)
```

<details>
<summary>Error</summary>

```
free(): invalid pointer

[970911] signal 6 (-6): Aborted
in expression starting at REPL[8]:1
unknown function (ip: 0x7b2c897013f4)
gsignal at /usr/lib/libc.so.6 (unknown line)
abort at /usr/lib/libc.so.6 (unknown line)
unknown function (ip: 0x7b2c89690353)
unknown function (ip: 0x7b2c8970b764)
unknown function (ip: 0x7b2c8970dc4b)
__libc_free at /usr/lib/libc.so.6 (unknown line)
_ZN4mlir14OperationStateD1Ev at /home/avikpal/.julia/artifacts/683e7a1671039feec0b9faea91812e8b4e08ded5/lib/libReactantExtra.so (unknown line)
mlirOperationCreate at /home/avikpal/.julia/artifacts/683e7a1671039feec0b9faea91812e8b4e08ded5/lib/libReactantExtra.so (unknown line)
mlirOperationCreate at /mnt/software/lux/Reactant.jl/src/mlir/libMLIR_h.jl:985 [inlined]
#create_operation#56 at /mnt/software/lux/Reactant.jl/src/mlir/IR/Operation.jl:315
create_operation at /mnt/software/lux/Reactant.jl/src/mlir/IR/Operation.jl:273 [inlined]
#if_#53 at /mnt/software/lux/Reactant.jl/src/mlir/Dialects/StableHLO.jl:2287
if_ at /mnt/software/lux/Reactant.jl/src/mlir/Dialects/StableHLO.jl:2274 [inlined]
traced_if at /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:127
unknown function (ip: 0x7b2c57f34986)
macro expansion at /mnt/software/lux/Reactant.jl/src/ControlFlow.jl:70 [inlined]
conditional_fn at ./REPL[2]:4 [inlined]
opaque closure at ./<missing>:0
unknown function (ip: 0x7b2c57f33832)
#26 at /mnt/software/lux/Reactant.jl/src/utils.jl:113
block! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Block.jl:201
unknown function (ip: 0x7b2c6dbd1d96)
#make_mlir_fn#20 at /mnt/software/lux/Reactant.jl/src/utils.jl:81
make_mlir_fn at /mnt/software/lux/Reactant.jl/src/utils.jl:30 [inlined]
#6 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:260 [inlined]
block! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Block.jl:201
#5 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:259 [inlined]
mmodule! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Module.jl:93
unknown function (ip: 0x7b2c6db5dcf6)
#compile_mlir!#4 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:256
compile_mlir! at /mnt/software/lux/Reactant.jl/src/Compiler.jl:255 [inlined]
#30 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:636
context! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Context.jl:71
unknown function (ip: 0x7b2c6db56756)
#compile_xla#29 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:633
compile_xla at /mnt/software/lux/Reactant.jl/src/Compiler.jl:627 [inlined]
#compile#34 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:660
compile at /mnt/software/lux/Reactant.jl/src/Compiler.jl:659
unknown function (ip: 0x7b2c6db4f5ed)
jl_apply at /cache/build/builder-demeter6-6/julialang/julia-master/src/julia.h:2157 [inlined]
do_call at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:126
eval_value at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:223
eval_stmt_value at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:174 [inlined]
eval_body at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:663
jl_interpret_toplevel_thunk at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:821
jl_toplevel_eval_flex at /cache/build/builder-demeter6-6/julialang/julia-master/src/toplevel.c:943
jl_toplevel_eval_flex at /cache/build/builder-demeter6-6/julialang/julia-master/src/toplevel.c:886
eval_body at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:625
eval_body at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:539
jl_interpret_toplevel_thunk at /cache/build/builder-demeter6-6/julialang/julia-master/src/interpreter.c:821
jl_toplevel_eval_flex at /cache/build/builder-demeter6-6/julialang/julia-master/src/toplevel.c:943
jl_toplevel_eval_flex at /cache/build/builder-demeter6-6/julialang/julia-master/src/toplevel.c:886
jl_toplevel_eval_flex at /cache/build/builder-demeter6-6/julialang/julia-master/src/toplevel.c:886
ijl_toplevel_eval_in at /cache/build/builder-demeter6-6/julialang/julia-master/src/toplevel.c:994
eval at ./boot.jl:430 [inlined]
eval_user_input at /cache/build/builder-demeter6-6/julialang/julia-master/usr/share/julia/stdlib/v1.11/REPL/src/REPL.jl:245
repl_backend_loop at /cache/build/builder-demeter6-6/julialang/julia-master/usr/share/julia/stdlib/v1.11/REPL/src/REPL.jl:342
#start_repl_backend#59 at /cache/build/builder-demeter6-6/julialang/julia-master/usr/share/julia/stdlib/v1.11/REPL/src/REPL.jl:327
start_repl_backend at /cache/build/builder-demeter6-6/julialang/julia-master/usr/share/julia/stdlib/v1.11/REPL/src/REPL.jl:324
#run_repl#72 at /cache/build/builder-demeter6-6/julialang/julia-master/usr/share/julia/stdlib/v1.11/REPL/src/REPL.jl:483
run_repl at /cache/build/builder-demeter6-6/julialang/julia-master/usr/share/julia/stdlib/v1.11/REPL/src/REPL.jl:469
jfptr_run_repl_10088 at /home/avikpal/.julia/juliaup/julia-1.11.1+0.x64.linux.gnu/share/julia/compiled/v1.11/REPL/u0gqU_GYsA8.so (unknown line)
#1139 at ./client.jl:446
jfptr_YY.1139_14649 at /home/avikpal/.julia/juliaup/julia-1.11.1+0.x64.linux.gnu/share/julia/compiled/v1.11/REPL/u0gqU_GYsA8.so (unknown line)
jl_apply at /cache/build/builder-demeter6-6/julialang/julia-master/src/julia.h:2157 [inlined]
jl_f__call_latest at /cache/build/builder-demeter6-6/julialang/julia-master/src/builtins.c:875
#invokelatest#2 at ./essentials.jl:1055 [inlined]
invokelatest at ./essentials.jl:1052 [inlined]
run_main_repl at ./client.jl:430
repl_main at ./client.jl:567 [inlined]
_start at ./client.jl:541
jfptr__start_72144.1 at /home/avikpal/.julia/juliaup/julia-1.11.1+0.x64.linux.gnu/lib/julia/sys.so (unknown line)
jl_apply at /cache/build/builder-demeter6-6/julialang/julia-master/src/julia.h:2157 [inlined]
true_main at /cache/build/builder-demeter6-6/julialang/julia-master/src/jlapi.c:900
jl_repl_entrypoint at /cache/build/builder-demeter6-6/julialang/julia-master/src/jlapi.c:1059
main at /cache/build/builder-demeter6-6/julialang/julia-master/cli/loader_exe.c:58
unknown function (ip: 0x7b2c89690e07)
__libc_start_main at /usr/lib/libc.so.6 (unknown line)
unknown function (ip: 0x4010b8)
Allocations: 71349677 (Pool: 71347786; Big: 1891); GC: 47
[1]    970911 IOT instruction (core dumped)  julia --project=. --threads=12
```

</details>

Can I get some help in the MLIR operation part here? I am sure I am doing something illegal here

src/ControlFlow.jl
```diff
@@ -0,0 +1,214 @@
...
+    false_branch_region = MLIR.IR.Region(
+        MLIR.API.mlirOperationGetRegion(false_branch_compiled, 0), true
+    )
+
+    if_compiled = MLIR.Dialects.stablehlo.if_(
+        cond.mlir_data;
+        true_branch=true_branch_region,
+        false_branch=false_branch_region,
+        result_0=results,
+    )
```
avik-pal, line 130: `if_` needs `owned` to be true for the region

src/ControlFlow.jl
```diff
@@ -0,0 +1,214 @@
+module ControlFlow
+
+using ..Reactant: Reactant, TracedRNumber, TracedRArray
+using ..MLIR: MLIR
+
+using MacroTools: MacroTools
+
+macro trace(expr)
+    expr.head == :if && return esc(trace_if(__module__, expr))
+    return error("Only `if-elseif-else` blocks are currently supported by `@trace`")
```
wsmoses, line 10: we should make sure to test [bare if  end / if else end /  if elseif end / if elseif else end] x [returning different values from the if /  overwriting a local value (e.g. x = x + 2) / in place updating values (e.g. x[2] = 4) ]

src/ControlFlow.jl
```diff
@@ -0,0 +1,214 @@
...
+end
+
+function trace_if(mod, expr)
+    @assert expr.head == :if
+    @assert length(expr.args) == 3 "`@trace` expects an `else` block for `if` blocks."
+    # XXX: support `elseif` blocks
+    @assert expr.args[3].head == :block "`elseif` blocks are not supported yet."
```
wsmoses, line 17: well above tests for what's supported at least xD

src/ControlFlow.jl
```diff
@@ -0,0 +1,214 @@
...
+        return false_fn(args...)
+    end
+end
+
+function traced_if(
+    cond::TracedRNumber{Bool}, true_fn::TFn, false_fn::FFn, args
+) where {TFn,FFn}
+    _, true_branch_compiled, true_branch_results, _, _, _, _, _, true_linear_results = Reactant.make_mlir_fn(
+        true_fn, args, (), string(gensym("true_branch")), false
+    )
```
wsmoses, line 102: how would this work if the args / sides were distinct.

For example

```julia

function f(x, y)
  @traced if x < y
     x = x + 1
   else
     y = y + 1
    end
end
```

wsmoses: [here this could be extra obonxious since if you don't union the returned value set of both the number and types of args are the same]

src/ControlFlow.jl
wsmoses: what do you mean by this?

avik-pal: If I do `MLIR.IR.region(false_branch_compiled, 1)` and pass it to `false_branch` I get an error in `create_operation` stating that `owned` is false.

src/ControlFlow.jl
avik-pal: The above example will be fine, since the returns are union-ed. But the following errors:

```julia
function f(x, y)
  @traced if x < y
     x = x + 1
	 p = 2
   else
     y = y + 1
    end
end
```

wsmoses:
Probably the solution is to not pass existing regions, but then to move the
child blocks from the old regions into the if

Traced:
您好，我看到以后就回复您

src/ControlFlow.jl
wsmoses: hm, why?

src/ControlFlow.jl
Pangoraw: You might want to do something like this:
```julia
false_branch_region = let reg = Region()
  MLIR.API.mlirRegionTakeBody(reg, MLIR.API.mlirOperationGetRegion(false_branch_compiled, 0))
  reg
end
```
Setting owned to true when it's not the case will cause a double-free error.

src/ControlFlow.jl
```diff
@@ -0,0 +1,214 @@
...
+        true_branch=true_branch_region,
+        false_branch=false_branch_region,
+        result_0=results,
+    )
+
+    return error("WIP")
+end
+
+# NOTE: Adapted from https://github.com/c42f/FastClosures.jl/blob/master/src/FastClosures.jl
+function find_var_uses!(varlist, bound_vars, ex)
```
Pangoraw, line 136: Just pointing out that [ExpressionExplorer.jl](https://github.com/JuliaPluto/ExpressionExplorer.jl) should be more up to date for things like this.

src/ControlFlow.jl
avik-pal: Thanks, this does it

src/ControlFlow.jl
avik-pal: Thanks for pointing it out. Using it gives much more accurate results

src/ControlFlow.jl
wsmoses: honestly here I'm fine if we define p as a new Constant Reactant.UndefinedValue or something

avik-pal:
```julia
using Reactant

function conditional_fn(x)
    v = sum(x)
    @trace if v > 0
        v2 = v
    else
        v2 = -v
    end
    return v2, x
end

conditional_fn(rand(2, 3))

x = rand(2, 3)

x_ra = Reactant.to_rarray(x)

conditional_fn_compiled = @compile conditional_fn(x_ra)
```

```mlir
julia> @code_hlo optimize=false conditional_fn(x_ra)
Module:
"builtin.module"() ({
  "func.func"() <{function_type = (tensor<f64>) -> tensor<f64>, sym_name = "identity_broadcast_scalar", sym_visibility = "private"}> ({
  ^bb0(%arg5: tensor<f64>):
    %17 = "stablehlo.transpose"(%arg5) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
    %18 = "stablehlo.transpose"(%17) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
    "func.return"(%18) : (tensor<f64>) -> ()
  }) : () -> ()
  "func.func"() <{function_type = (tensor<f64>) -> tensor<f64>, sym_name = "##true_branch#320", sym_visibility = "private"}> ({
  }) : () -> ()
  "func.func"() <{function_type = (tensor<f64>) -> (tensor<f64>, tensor<f64>), sym_name = "##false_branch#321", sym_visibility = "private"}> ({
  }) : () -> ()
  "func.func"() <{function_type = (tensor<3x2xf64>) -> (tensor<f64>, tensor<3x2xf64>), sym_name = "main"}> ({
  ^bb0(%arg0: tensor<3x2xf64>):
    %0 = "stablehlo.transpose"(%arg0) <{permutation = array<i64: 1, 0>}> : (tensor<3x2xf64>) -> tensor<2x3xf64>
    %1 = "stablehlo.constant"() <{value = dense<0.000000e+00> : tensor<f64>}> : () -> tensor<f64>
    %2 = "stablehlo.broadcast_in_dim"(%0) <{broadcast_dimensions = array<i64: 0, 1>}> : (tensor<2x3xf64>) -> tensor<2x3xf64>
    %3 = "enzyme.batch"(%2) <{batch_shape = array<i64: 2, 3>, fn = @identity_broadcast_scalar}> : (tensor<2x3xf64>) -> tensor<2x3xf64>
    %4 = "stablehlo.reduce"(%3, %1) <{dimensions = array<i64: 0, 1>}> ({
    ^bb0(%arg3: tensor<f64>, %arg4: tensor<f64>):
      %16 = "stablehlo.add"(%arg3, %arg4) : (tensor<f64>, tensor<f64>) -> tensor<f64>
      "stablehlo.return"(%16) : (tensor<f64>) -> ()
    }) : (tensor<2x3xf64>, tensor<f64>) -> tensor<f64>
    %5 = "stablehlo.constant"() <{value = dense<0.000000e+00> : tensor<f64>}> : () -> tensor<f64>
    %6 = "stablehlo.compare"(%4, %5) <{comparison_direction = #stablehlo<comparison_direction GT>}> : (tensor<f64>, tensor<f64>) -> tensor<i1>
    %7 = "stablehlo.if"(%6) ({
    ^bb0(%arg2: tensor<f64>):
      %14 = "stablehlo.transpose"(%arg2) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      %15 = "stablehlo.transpose"(%14) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      "stablehlo.return"(%15) : (tensor<f64>) -> ()
    }, {
    ^bb0(%arg1: tensor<f64>):
      %10 = "stablehlo.transpose"(%arg1) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      %11 = "stablehlo.negate"(%10) : (tensor<f64>) -> tensor<f64>
      %12 = "stablehlo.transpose"(%11) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      %13 = "stablehlo.transpose"(%10) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      "stablehlo.return"(%12, %13) : (tensor<f64>, tensor<f64>) -> ()
    }) : (tensor<i1>) -> tensor<f64>
    %8 = "stablehlo.transpose"(%7) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
    %9 = "stablehlo.transpose"(%0) <{permutation = array<i64: 1, 0>}> : (tensor<2x3xf64>) -> tensor<3x2xf64>
    "func.return"(%8, %9) : (tensor<f64>, tensor<3x2xf64>) -> ()
  }) : () -> ()
}) : () -> ()
```

Doesn't yet work completely. The branches in `stablehlo.if` needs to take in 0 args so the generated code is currently incorrect

wsmoses:
so this shouldn't be bad to fix (assuming that the true/false regions just accept the args from the parent -- which is notably _not_ how while works (which takes them as blockarguments like you have setup here). Essentially we need to do a replace all users of the blocks with the actual values in the parent (there should be a replace api call in mlir.jl cc @mofeing) and then delete the now unused blockargs

avik-pal:
```mlir
Module:
"builtin.module"() ({
  "func.func"() <{function_type = (tensor<f64>) -> tensor<f64>, sym_name = "identity_broadcast_scalar", sym_visibility = "private"}> ({
  ^bb0(%arg5: tensor<f64>):
    %17 = "stablehlo.transpose"(%arg5) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
    %18 = "stablehlo.transpose"(%17) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
    "func.return"(%18) : (tensor<f64>) -> ()
  }) : () -> ()
  "func.func"() <{function_type = (tensor<f64>) -> tensor<f64>, sym_name = "##true_branch#282", sym_visibility = "private"}> ({
  }) : () -> ()
  "func.func"() <{function_type = (tensor<f64>) -> (tensor<f64>, tensor<f64>), sym_name = "##false_branch#283", sym_visibility = "private"}> ({
  }) : () -> ()
  "func.func"() <{function_type = (tensor<3x2xf64>) -> (tensor<f64>, tensor<3x2xf64>), sym_name = "main"}> ({
  ^bb0(%arg0: tensor<3x2xf64>):
    %0 = "stablehlo.transpose"(%arg0) <{permutation = array<i64: 1, 0>}> : (tensor<3x2xf64>) -> tensor<2x3xf64>
    %1 = "stablehlo.constant"() <{value = dense<0.000000e+00> : tensor<f64>}> : () -> tensor<f64>
    %2 = "stablehlo.broadcast_in_dim"(%0) <{broadcast_dimensions = array<i64: 0, 1>}> : (tensor<2x3xf64>) -> tensor<2x3xf64>
    %3 = "enzyme.batch"(%2) <{batch_shape = array<i64: 2, 3>, fn = @identity_broadcast_scalar}> : (tensor<2x3xf64>) -> tensor<2x3xf64>
    %4 = "stablehlo.reduce"(%3, %1) <{dimensions = array<i64: 0, 1>}> ({
    ^bb0(%arg3: tensor<f64>, %arg4: tensor<f64>):
      %16 = "stablehlo.add"(%arg3, %arg4) : (tensor<f64>, tensor<f64>) -> tensor<f64>
      "stablehlo.return"(%16) : (tensor<f64>) -> ()
    }) : (tensor<2x3xf64>, tensor<f64>) -> tensor<f64>
    %5 = "stablehlo.constant"() <{value = dense<0.000000e+00> : tensor<f64>}> : () -> tensor<f64>
    %6 = "stablehlo.compare"(%4, %5) <{comparison_direction = #stablehlo<comparison_direction GT>}> : (tensor<f64>, tensor<f64>) -> tensor<i1>
    %7 = "stablehlo.if"(%6) ({
    ^bb0(%arg2: tensor<f64>):
      %14 = "stablehlo.transpose"(%arg2) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      %15 = "stablehlo.transpose"(%4) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      "stablehlo.return"(%15) : (tensor<f64>) -> ()
    }, {
    ^bb0(%arg1: tensor<f64>):
      %10 = "stablehlo.transpose"(%arg1) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      %11 = "stablehlo.negate"(%4) : (tensor<f64>) -> tensor<f64>
      %12 = "stablehlo.transpose"(%11) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      %13 = "stablehlo.transpose"(%4) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
      "stablehlo.return"(%12, %13) : (tensor<f64>, tensor<f64>) -> ()
    }) : (tensor<i1>) -> tensor<f64>
    %8 = "stablehlo.transpose"(%7) <{permutation = array<i64>}> : (tensor<f64>) -> tensor<f64>
    %9 = "stablehlo.transpose"(%0) <{permutation = array<i64: 1, 0>}> : (tensor<2x3xf64>) -> tensor<3x2xf64>
    "func.return"(%8, %9) : (tensor<f64>, tensor<3x2xf64>) -> ()
  }) : () -> ()
}) : () -> ()
```

How do I delete the arguments? I don't see an equivalent to https://mlir.llvm.org/doxygen/classmlir_1_1Block.html#ab497ed8c6f52f6faa5e9a8ac0c6b1014 in reactant

Pangoraw:
> How do I delete the arguments?

yeah it seems that `mlirBlockEraseArgument` is only available in LLVM >= v19. We can try to clone operations in a new block without argument ?

mofeing:
> > How do I delete the arguments?
> 
> yeah it seems that `mlirBlockEraseArgument` is only available in LLVM >= v19. We can try to clone operations in a new block without argument ?

Remember that Reactant uses LLVM 20 currently, because it has its own LLVM build with XLA.

mofeing:
> > > How do I delete the arguments?
> > 
> > 
> > yeah it seems that `mlirBlockEraseArgument` is only available in LLVM >= v19. We can try to clone operations in a new block without argument ?
> 
> Remember that Reactant uses LLVM 20-DEV currently, because it has its own LLVM build with XLA.

Pangoraw:
oh nice, then we need to regenerate the LibMLIR file to get `mlirBlockEraseArgument`

src/ControlFlow.jl
avik-pal:
```julia
julia> using Reactant

julia> function f(x, y)
           @trace if sum(x) < sum(y)
               x = x .+ 1
               p = 2
           else
               y = y .+ 1
           end
           return x, y, p
       end

julia> x = rand(2, 3)
2×3 Matrix{Float64}:
 0.382612  0.859697  0.575209
 0.38412   0.992373  0.881686

julia> y = rand(2, 3)
2×3 Matrix{Float64}:
 0.564671  0.244165  0.990224
 0.483621  0.811683  0.63983

julia> x_ra = Reactant.to_rarray(x)
2×3 ConcreteRArray{Float64, 2}:
 0.382612  0.859697  0.575209
 0.38412   0.992373  0.881686

julia> y_ra = Reactant.to_rarray(y)
2×3 ConcreteRArray{Float64, 2}:
 0.564671  0.244165  0.990224
 0.483621  0.811683  0.63983

julia> @code_hlo optimize = true f(x_ra, y_ra)
Module:
  func.func @main(%arg0: tensor<3x2xf64>, %arg1: tensor<3x2xf64>) -> (tensor<3x2xf64>, tensor<3x2xf64>, tensor<i64>) {
    %cst = stablehlo.constant dense<1.000000e+00> : tensor<2x3xf64>
    %c = stablehlo.constant dense<0> : tensor<i64>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<2x3xf64>
    %c_1 = stablehlo.constant dense<2> : tensor<i64>
    %cst_2 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<3x2xf64>) -> tensor<2x3xf64>
    %1 = stablehlo.transpose %arg1, dims = [1, 0] : (tensor<3x2xf64>) -> tensor<2x3xf64>
    %2 = stablehlo.reduce(%0 init: %cst_2) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x3xf64>, tensor<f64>) -> tensor<f64>
    %3 = stablehlo.reduce(%1 init: %cst_2) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x3xf64>, tensor<f64>) -> tensor<f64>
    %4 = stablehlo.compare  LT, %2, %3 : (tensor<f64>, tensor<f64>) -> tensor<i1>
    %5:3 = "stablehlo.if"(%4) ({
      %8 = stablehlo.add %0, %cst : tensor<2x3xf64>
      stablehlo.return %c_1, %8, %cst_0 : tensor<i64>, tensor<2x3xf64>, tensor<2x3xf64>
    }, {
      %8 = stablehlo.add %1, %cst : tensor<2x3xf64>
      stablehlo.return %c, %cst_0, %8 : tensor<i64>, tensor<2x3xf64>, tensor<2x3xf64>
    }) : (tensor<i1>) -> (tensor<i64>, tensor<2x3xf64>, tensor<2x3xf64>)
    %6 = stablehlo.transpose %5#1, dims = [1, 0] : (tensor<2x3xf64>) -> tensor<3x2xf64>
    %7 = stablehlo.transpose %5#2, dims = [1, 0] : (tensor<2x3xf64>) -> tensor<3x2xf64>
    return %6, %7, %5#0 : tensor<3x2xf64>, tensor<3x2xf64>, tensor<i64>
  }
}
```

Now we can support this as well.

avik-pal:
Can I get a preliminary review? I still need to fix certain things but the conditionals should work now.

wsmoses (commented):
Generally looks reasonable to me, but we should probably test the various funky ways (e.g. mutation in place, returned values, missing values, etc) it could be used

src/ControlFlow.jl
```diff
@@ -0,0 +1,235 @@
...
+    all_vars = all_input_vars ∪ all_output_vars
+
+    non_existant_true_branch_vars = setdiff(all_output_vars, all_true_branch_vars)
+    true_branch_extras = Expr(
+        :block,
+        [:($(var) = $(MissingTracedValue())) for var in non_existant_true_branch_vars]...,
+    )
+
+    true_branch_fn = quote
+        $(true_branch_fn_name) =
```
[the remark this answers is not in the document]
avik-pal: I couldn't figure out a way to push the function to the top-level. It would also let me remove the MacroTools hack to prevent boxing when Reactan't isn't used

src/ControlFlow.jl
```diff
@@ -0,0 +1,235 @@
...
+            ($(all_input_vars...),) -> begin
```
wsmoses, line 69: Similarly it's worth checking how the macro behaves when not inside of reactant tracing

avik-pal: completely bypasses if none of the variables are traced

src/ControlFlow.jl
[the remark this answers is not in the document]
wsmoses: I guess if we could find a way to encode the expressions in types it would work.

e.g. 

```julia
struct TrueBranch{NVars, Exprs} end

@genereated (::TrueBranch{NVars, Exprs})(args::Varargs{NVars, Any}) where {NVars, Any}
   ....
   return quote
       ...
   end
end
```

wsmoses:
> One shortcoming of having to define an `else` branch forcefully is that the following will give incorrect results:
> 
> ```julia
> function relu(x::T) where {T}
>     @trace if x < 0
>         x = 0.0f0
>     end
>     return x
> end
> ```
> 
> Maybe we should warn if user doesn't define an else?

So this should be fixable, We'd just define the else region as the identity where the inputs are the variables in scope of the if

avik-pal:
Failure case for broadcasting:

```julia
using Reactant

function relu(x::T) where {T <: Number}
    @trace if x < 0
        x = 0.0f0
    end
    return x
end

@jit relu(Reactant.to_rarray(3.0f0; track_numbers=(Number,))) # Works

relu_act(x::AbstractArray) = relu.(x)

x = randn(Float32, 1024);
x_ra = Reactant.to_rarray(x);

@jit relu_act(x_ra)
# error: operand should be rank 0 tensor but got rank 1
# error: 'stablehlo.if' op failed to infer returned types
# ERROR: "failed to run pass manager on module"
# Stacktrace:
#   [1] run!
#     @ /mnt/software/lux/Reactant.jl/src/mlir/IR/Pass.jl:70 [inlined]
#   [2] run_pass_pipeline!(mod::Reactant.MLIR.IR.Module, pass_pipeline::String)
#     @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:251
#   [3] compile_mlir!(mod::Reactant.MLIR.IR.Module, f::Function, args::Tuple{ConcreteRArray{Float32, 1}}; optimize::Bool)
#     @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:284
#   [4] compile_mlir!
#     @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:266 [inlined]
#   [5] (::Reactant.Compiler.var"#30#32"{Bool, typeof(relu_act), Tuple{ConcreteRArray{Float32, 1}}})()
#     @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:671
#   [6] context!(f::Reactant.Compiler.var"#30#32"{Bool, typeof(relu_act), Tuple{ConcreteRArray{Float32, 1}}}, ctx::Reactant.MLIR.IR.Context)
#     @ Reactant.MLIR.IR /mnt/software/lux/Reactant.jl/src/mlir/IR/Context.jl:71
#   [7] compile_xla(f::Function, args::Tuple{ConcreteRArray{Float32, 1}}; client::Nothing, optimize::Bool)
#     @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:668
#   [8] compile_xla
#     @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:662 [inlined]
#   [9] compile(f::Function, args::Tuple{ConcreteRArray{Float32, 1}}; client::Nothing, optimize::Bool)
#     @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:695
#  [10] top-level scope
#     @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:454
#  [11] top-level scope
#     @ none:1

@code_hlo optimize = false relu_act(x_ra)
```

```mlir
Module:
module {
  func.func private @relu_broadcast_scalar(%arg0: tensor<f32>) -> (tensor<f32>, tensor<f32>) {
    %0 = stablehlo.transpose %arg0, dims = [] : (tensor<f32>) -> tensor<f32>
    %1 = stablehlo.compare  LT, %0, %cst : (tensor<f32>, tensor<f32>) -> tensor<i1>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f32>
    %2 = "stablehlo.if"(%1) ({
      stablehlo.return %cst_0 : tensor<f32>
    }, {
      stablehlo.return %0 : tensor<f32>
    }) : (tensor<i1>) -> tensor<f32>
    %3 = stablehlo.transpose %2, dims = [] : (tensor<f32>) -> tensor<f32>
    %4 = stablehlo.transpose %0, dims = [] : (tensor<f32>) -> tensor<f32>
    return %3, %4 : tensor<f32>, tensor<f32>
  }
  func.func @main(%arg0: tensor<1024xf32>) -> (tensor<1024xf32>, tensor<1024xf32>) {
    %0 = stablehlo.transpose %arg0, dims = [0] : (tensor<1024xf32>) -> tensor<1024xf32>
    %1 = stablehlo.broadcast_in_dim %0, dims = [0] : (tensor<1024xf32>) -> tensor<1024xf32>
    %2:2 = enzyme.batch @relu_broadcast_scalar(%1) {batch_shape = array<i64: 1024>} : (tensor<1024xf32>) -> (tensor<1024xf32>, tensor<1024xf32>)
    %3 = stablehlo.transpose %2#0, dims = [0] : (tensor<1024xf32>) -> tensor<1024xf32>
    %4 = stablehlo.transpose %0, dims = [0] : (tensor<1024xf32>) -> tensor<1024xf32>
    return %3, %4 : tensor<1024xf32>, tensor<1024xf32>
  }
}
```

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -0,0 +1,147 @@
...
+        ($(all_output_vars...),) = $(traced_if)(
+            $(expr.args[1]),
+            $(true_branch_fn_name),
+            $(false_branch_fn_name),
+            ($(all_input_vars...),),
+        )
+    end
+
+    expr.head != :if &&
+        return reactant_code_block, (true_branch_fn_name, false_branch_fn_name)
```
Pangoraw, line 112: I don't really understand what the intent is here, shouldn't this throw?

lib/ReactantCore/Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "ReactantCore"
+uuid = "a3311ec8-5e00-46d5-b541-4f83e724a433"
+authors = ["William Moses <wmoses@mit.edu>", "Valentin Churavy <vchuravy@mit.edu>", "Sergio Sánchez Ramírez <sergio.sanchez.ramirez@bsc.es>", "Paul Berg <paul@plutojl.org>"]
```
mofeing, line 3: Wanna add yourself @avik-pal? 😆

lib/ReactantCore/src/ReactantCore.jl
avik-pal: The only codepath that can hit this is if it was an elseif. I should probably change it to check for elseif and not a negation on if

lib/ReactantCore/Project.toml
avik-pal: yeah I will :sweat_smile:

avik-pal:
Needs fixing:

```julia

function condition2_nested_if(x, y)
    x_sum = sum(x)
    @trace if x_sum > 0
        y_sum = sum(y)
        @trace if y_sum > 0
            z = x_sum + y_sum
        else
            z = x_sum - y_sum
        end
    else
        y_sum = sum(y)
        z = x_sum - y_sum
    end
    return z
end

x = rand(2, 10)
y = rand(2, 10)
x_ra = Reactant.to_rarray(x)
y_ra = Reactant.to_rarray(y)

@code_hlo optimize=false condition2_nested_if(x_ra, y_ra)
```

I am definitely doing some incorrect IR manipulation. `%15` shouldn't be repeated

```mlir
Module:
module {
  func.func private @identity_broadcast_scalar(%arg0: tensor<f64>) -> tensor<f64> {
    %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
    return %1 : tensor<f64>
  }
  func.func private @identity_broadcast_scalar1(%arg0: tensor<f64>) -> tensor<f64> {
    %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
    return %1 : tensor<f64>
  }
  func.func private @identity_broadcast_scalar2(%arg0: tensor<f64>) -> tensor<f64> {
    %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
    return %1 : tensor<f64>
  }
  func.func @main(%arg0: tensor<10x2xf64>, %arg1: tensor<10x2xf64>) -> (tensor<f64>, tensor<10x2xf64>, tensor<10x2xf64>) {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
    %1 = stablehlo.transpose %arg1, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
    %2 = stablehlo.broadcast_in_dim %0, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
    %3 = enzyme.batch @identity_broadcast_scalar(%2) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
    %4 = stablehlo.reduce(%3 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
    %5 = stablehlo.compare  GT, %4, %cst_0 : (tensor<f64>, tensor<f64>) -> tensor<i1>
    %6:2 = "stablehlo.if"(%5) ({
      %cst_1 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
      %cst_2 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
      %10 = stablehlo.broadcast_in_dim %1, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
      %11 = enzyme.batch @identity_broadcast_scalar1(%10) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
      %12 = stablehlo.reduce(%11 init: %cst_2) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
      %cst_3 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
      %13 = stablehlo.compare  GT, %12, %cst_3 : (tensor<f64>, tensor<f64>) -> tensor<i1>
      %14 = "stablehlo.if"(%13) ({
        %15 = stablehlo.add %4, %12 : tensor<f64>
        stablehlo.return %15 : tensor<f64>
      }, {
        %15 = stablehlo.subtract %4, %12 : tensor<f64>
        stablehlo.return %15 : tensor<f64>
      }) : (tensor<i1>) -> tensor<f64>
      %c = stablehlo.constant dense<false> : tensor<i1>
      stablehlo.return %12, %cst_1 : tensor<f64>, tensor<f64>
    }, {
      %cst_1 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
      %10 = stablehlo.broadcast_in_dim %1, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
      %11 = enzyme.batch @identity_broadcast_scalar2(%10) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
      %12 = stablehlo.reduce(%11 init: %cst_1) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
      %13 = stablehlo.subtract %4, %12 : tensor<f64>
      stablehlo.return %12, %13 : tensor<f64>, tensor<f64>
    }) : (tensor<i1>) -> (tensor<f64>, tensor<f64>)
    %7 = stablehlo.transpose %6#1, dims = [] : (tensor<f64>) -> tensor<f64>
    %8 = stablehlo.transpose %0, dims = [1, 0] : (tensor<2x10xf64>) -> tensor<10x2xf64>
    %9 = stablehlo.transpose %1, dims = [1, 0] : (tensor<2x10xf64>) -> tensor<10x2xf64>
    return %7, %8, %9 : tensor<f64>, tensor<10x2xf64>, tensor<10x2xf64>
  }
}
```

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -0,0 +1,201 @@
...
+end
+
+MissingTracedValue() = MissingTracedValue(())
+
+# Code generation
+"""
+    @trace <expr>
+
+Converts certain expressions like control flow into a Reactant friendly form. Importantly,
+if no traced value is found inside the expression, then there is no overhead.
```
wsmoses, line 23: Can you add a sentence about that this assumes the results/updates are of the same julia types (in contrast to not using @trace which will permits this behavior)

test/control_flow.jl
```diff
@@ -0,0 +1,267 @@
...
+    res_ra = @jit(condition7_bare_elseif(x_ra))
+    res = condition7_bare_elseif(x)
+    @test res_ra ≈ res
+
+    x = 0.0
+    x_ra = Reactant.to_rarray(x; track_numbers=(Number,))
+
+    res_ra = @jit(condition7_bare_elseif(x_ra))
+    res = condition7_bare_elseif(x)
+    @test res_ra ≈ res
```
wsmoses, line 266: just for some sanity checks, can you add cases where the if statement returns values (e.g.

```julia
a, b = @traced if c
  ....
else
  ...
end
```

as well as cases where it returns nothing
```julia
if a 
  x = ...
  nothing
else
  ...
   nothing
end
```

and an in place mutation (e.g. not just a variable scope question).

```julia
if c
   mem[3] = ...
   nothing
end   
```

avik-pal:
Mutation gets completely optimized out

```julia
julia> function condition10_condition_with_setindex(x)
           @trace if sum(x) > 0
               x[1, 1] = -1.0
           else
               x[1, 1] = 1.0
           end
           return x
       end
condition10_condition_with_setindex (generic function with 1 method)

julia> @code_hlo condition10_condition_with_setindex(x_ra)
Module:
  func.func @main(%arg0: tensor<10x2xf64>) {
    return
  }
}

julia> @code_hlo optimize=false condition10_condition_with_setindex(x_ra)
Module:
module {
  func.func private @identity_broadcast_scalar(%arg0: tensor<f64>) -> tensor<f64> {
    %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
    return %1 : tensor<f64>
  }
  func.func @main(%arg0: tensor<10x2xf64>) -> tensor<10x2xf64> {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
    %1 = stablehlo.broadcast_in_dim %0, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
    %2 = enzyme.batch @identity_broadcast_scalar(%1) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
    %3 = stablehlo.reduce(%2 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
    %4 = stablehlo.compare  GT, %3, %cst_0 : (tensor<f64>, tensor<f64>) -> tensor<i1>
    "stablehlo.if"(%4) ({
      %cst_1 = stablehlo.constant dense<-1.000000e+00> : tensor<1x1xf64>
      %c_2 = stablehlo.constant dense<1> : tensor<i64>
      %6 = stablehlo.subtract %c, %c_2 : tensor<i64>
      %c_3 = stablehlo.constant dense<1> : tensor<i64>
      %c_4 = stablehlo.constant dense<1> : tensor<i64>
      %7 = stablehlo.subtract %c_3, %c_4 : tensor<i64>
      %8 = stablehlo.dynamic_update_slice %0, %cst_1, %6, %7 : (tensor<2x10xf64>, tensor<1x1xf64>, tensor<i64>, tensor<i64>) -> tensor<2x10xf64>
      stablehlo.return 
    }, {
      %cst_1 = stablehlo.constant dense<1.000000e+00> : tensor<1x1xf64>
      %c_2 = stablehlo.constant dense<1> : tensor<i64>
      %6 = stablehlo.subtract %c, %c_2 : tensor<i64>
      %c_3 = stablehlo.constant dense<1> : tensor<i64>
      %c_4 = stablehlo.constant dense<1> : tensor<i64>
      %7 = stablehlo.subtract %c_3, %c_4 : tensor<i64>
      %8 = stablehlo.dynamic_update_slice %0, %cst_1, %6, %7 : (tensor<2x10xf64>, tensor<1x1xf64>, tensor<i64>, tensor<i64>) -> tensor<2x10xf64>
      stablehlo.return 
    }) : (tensor<i1>) -> ()
    %5 = stablehlo.transpose %0, dims = [1, 0] : (tensor<2x10xf64>) -> tensor<10x2xf64>
    return %5 : tensor<10x2xf64>
  }
}
```

avik-pal:
https://github.com/EnzymeAD/Reactant.jl/pull/185#issuecomment-2450745712 and the one above seem to be the weird cases left

wsmoses:
> Needs fixing:
> 
> ```julia
> function condition2_nested_if(x, y)
>     x_sum = sum(x)
>     @trace if x_sum > 0
>         y_sum = sum(y)
>         @trace if y_sum > 0
>             z = x_sum + y_sum
>         else
>             z = x_sum - y_sum
>         end
>     else
>         y_sum = sum(y)
>         z = x_sum - y_sum
>     end
>     return z
> end
> 
> x = rand(2, 10)
> y = rand(2, 10)
> x_ra = Reactant.to_rarray(x)
> y_ra = Reactant.to_rarray(y)
> 
> @code_hlo optimize=false condition2_nested_if(x_ra, y_ra)
> ```
> 
> Pre-optimization
> 
> ```mlir
> Module:
> module {
>   func.func private @identity_broadcast_scalar(%arg0: tensor<f64>) -> tensor<f64> {
>     %0 = stablehlo.transpose %arg0, dims = [] : (tensor<f64>) -> tensor<f64>
>     %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
>     return %1 : tensor<f64>
>   }
>   func.func private @identity_broadcast_scalar1(%arg0: tensor<f64>) -> tensor<f64> {
>     %0 = stablehlo.transpose %arg0, dims = [] : (tensor<f64>) -> tensor<f64>
>     %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
>     return %1 : tensor<f64>
>   }
>   func.func private @identity_broadcast_scalar2(%arg0: tensor<f64>) -> tensor<f64> {
>     %0 = stablehlo.transpose %arg0, dims = [] : (tensor<f64>) -> tensor<f64>
>     %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
>     return %1 : tensor<f64>
>   }
>   func.func @main(%arg0: tensor<10x2xf64>, %arg1: tensor<10x2xf64>) -> (tensor<f64>, tensor<10x2xf64>, tensor<10x2xf64>) {
>     %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
>     %1 = stablehlo.transpose %arg1, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
>     %cst = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>     %2 = stablehlo.broadcast_in_dim %0, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
>     %3 = enzyme.batch @identity_broadcast_scalar(%2) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
>     %4 = stablehlo.reduce(%3 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
>     %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>     %5 = stablehlo.compare  GT, %4, %cst_0 : (tensor<f64>, tensor<f64>) -> tensor<i1>
>     %6:2 = "stablehlo.if"(%5) ({
>       %cst_1 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>       %cst_2 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>       %10 = stablehlo.broadcast_in_dim %1, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
>       %11 = enzyme.batch @identity_broadcast_scalar1(%10) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
>       %12 = stablehlo.reduce(%11 init: %cst_2) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
>       %cst_3 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>       %13 = stablehlo.compare  GT, %12, %cst_3 : (tensor<f64>, tensor<f64>) -> tensor<i1>
>       %14 = "stablehlo.if"(%13) ({
>         %15 = stablehlo.add %4, %12 : tensor<f64>
>         stablehlo.return %15 : tensor<f64>
>       }, {
>         %15 = stablehlo.subtract %4, %12 : tensor<f64>
>         stablehlo.return %15 : tensor<f64>
>       }) : (tensor<i1>) -> tensor<f64>
>       %c = stablehlo.constant dense<false> : tensor<i1>
>       stablehlo.return %12, %cst_1 : tensor<f64>, tensor<f64>
>     }, {
>       %cst_1 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>       %10 = stablehlo.broadcast_in_dim %1, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
>       %11 = enzyme.batch @identity_broadcast_scalar2(%10) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
>       %12 = stablehlo.reduce(%11 init: %cst_1) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
>       %13 = stablehlo.subtract %4, %12 : tensor<f64>
>       stablehlo.return %12, %13 : tensor<f64>, tensor<f64>
>     }) : (tensor<i1>) -> (tensor<f64>, tensor<f64>)
>     %7 = stablehlo.transpose %6#1, dims = [] : (tensor<f64>) -> tensor<f64>
>     %8 = stablehlo.transpose %0, dims = [1, 0] : (tensor<2x10xf64>) -> tensor<10x2xf64>
>     %9 = stablehlo.transpose %1, dims = [1, 0] : (tensor<2x10xf64>) -> tensor<10x2xf64>
>     return %7, %8, %9 : tensor<f64>, tensor<10x2xf64>, tensor<10x2xf64>
>   }
> }
> ```
> 
> gets optimized to
> 
> ```mlir
> Module:
> module attributes {transform.with_named_sequence} {
>   func.func @main(%arg0: tensor<10x2xf64>, %arg1: tensor<10x2xf64>) -> tensor<f64> {
>     %cst = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>     %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
>     %1 = stablehlo.transpose %arg1, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
>     %2 = stablehlo.reduce(%0 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
>     %3 = stablehlo.compare  GT, %2, %cst : (tensor<f64>, tensor<f64>) -> tensor<i1>
>     %4:2 = "stablehlo.if"(%3) ({
>       %5 = stablehlo.reduce(%1 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
>       stablehlo.return %5, %cst : tensor<f64>, tensor<f64>
>     }, {
>       %5 = stablehlo.reduce(%1 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
>       %6 = stablehlo.subtract %2, %5 : tensor<f64>
>       stablehlo.return %5, %6 : tensor<f64>, tensor<f64>
>     }) : (tensor<i1>) -> (tensor<f64>, tensor<f64>)
>     return %4#1 : tensor<f64>
>   }
> }
> ```
> 
> EDIT: ~Now if I run it in a new session it works so...~ NVM still doesn't work, if I introduce an extra `z = x_sum + y_sum` before the 2nd if then the optimized IR is correct

is this an optimization problem or an emission problem? [cc @Pangoraw ]

wsmoses:
> Mutation gets completely optimized out
> 
> ```julia
> julia> function condition10_condition_with_setindex(x)
>            @trace if sum(x) > 0
>                x[1, 1] = -1.0
>            else
>                x[1, 1] = 1.0
>            end
>            return x
>        end
> condition10_condition_with_setindex (generic function with 1 method)
> 
> julia> @code_hlo condition10_condition_with_setindex(x_ra)
> Module:
> module attributes {transform.with_named_sequence} {
>   func.func @main(%arg0: tensor<10x2xf64>) {
>     return
>   }
> }
> 
> julia> @code_hlo optimize=false condition10_condition_with_setindex(x_ra)
> Module:
> module {
>   func.func private @identity_broadcast_scalar(%arg0: tensor<f64>) -> tensor<f64> {
>     %0 = stablehlo.transpose %arg0, dims = [] : (tensor<f64>) -> tensor<f64>
>     %1 = stablehlo.transpose %0, dims = [] : (tensor<f64>) -> tensor<f64>
>     return %1 : tensor<f64>
>   }
>   func.func @main(%arg0: tensor<10x2xf64>) -> tensor<10x2xf64> {
>     %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<10x2xf64>) -> tensor<2x10xf64>
>     %cst = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>     %1 = stablehlo.broadcast_in_dim %0, dims = [0, 1] : (tensor<2x10xf64>) -> tensor<2x10xf64>
>     %2 = enzyme.batch @identity_broadcast_scalar(%1) {batch_shape = array<i64: 2, 10>} : (tensor<2x10xf64>) -> tensor<2x10xf64>
>     %3 = stablehlo.reduce(%2 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x10xf64>, tensor<f64>) -> tensor<f64>
>     %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
>     %4 = stablehlo.compare  GT, %3, %cst_0 : (tensor<f64>, tensor<f64>) -> tensor<i1>
>     "stablehlo.if"(%4) ({
>       %cst_1 = stablehlo.constant dense<-1.000000e+00> : tensor<1x1xf64>
>       %c = stablehlo.constant dense<1> : tensor<i64>
>       %c_2 = stablehlo.constant dense<1> : tensor<i64>
>       %6 = stablehlo.subtract %c, %c_2 : tensor<i64>
>       %c_3 = stablehlo.constant dense<1> : tensor<i64>
>       %c_4 = stablehlo.constant dense<1> : tensor<i64>
>       %7 = stablehlo.subtract %c_3, %c_4 : tensor<i64>
>       %8 = stablehlo.dynamic_update_slice %0, %cst_1, %6, %7 : (tensor<2x10xf64>, tensor<1x1xf64>, tensor<i64>, tensor<i64>) -> tensor<2x10xf64>
>       stablehlo.return 
>     }, {
>       %cst_1 = stablehlo.constant dense<1.000000e+00> : tensor<1x1xf64>
>       %c = stablehlo.constant dense<1> : tensor<i64>
>       %c_2 = stablehlo.constant dense<1> : tensor<i64>
>       %6 = stablehlo.subtract %c, %c_2 : tensor<i64>
>       %c_3 = stablehlo.constant dense<1> : tensor<i64>
>       %c_4 = stablehlo.constant dense<1> : tensor<i64>
>       %7 = stablehlo.subtract %c_3, %c_4 : tensor<i64>
>       %8 = stablehlo.dynamic_update_slice %0, %cst_1, %6, %7 : (tensor<2x10xf64>, tensor<1x1xf64>, tensor<i64>, tensor<i64>) -> tensor<2x10xf64>
>       stablehlo.return 
>     }) : (tensor<i1>) -> ()
>     %5 = stablehlo.transpose %0, dims = [1, 0] : (tensor<2x10xf64>) -> tensor<10x2xf64>
>     return %5 : tensor<10x2xf64>
>   }
> }
> ```

ah yeah so this is something that needs to be handled by the traced statement. 

You need to record the state of all local variables before the if statement, and after the if statement, to see if any of them had their mlir values change. 

Dynamic update in slice doesn't update in place [it returns a new one], so you need to also return updated values too [in addition to var assignments]

avik-pal:
Looking closer
```
%14 = "stablehlo.if"(%13) ({
        %15 = stablehlo.add %4, %12 : tensor<f64>
        stablehlo.return %15 : tensor<f64>
      }, {
        %15 = stablehlo.subtract %4, %12 : tensor<f64>
        stablehlo.return %15 : tensor<f64>
      }) : (tensor<i1>) -> tensor<f64>
      %c = stablehlo.constant dense<false> : tensor<i1>
      stablehlo.return %12, %cst_1 : tensor<f64>, tensor<f64>
```
`%14` is never used so it seems like I messed up the emission somewhere

avik-pal:
> is this an optimization problem or an emission problem? [cc @Pangoraw ]

This is now fixed. also patched an expression swell that happens if you nest multiple if conditions

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -56,42 +58,98 @@ end
 
 `&&` and `||` are short circuiting operations. In the traced version, we replace them with
 `&` and `|` respectively.
+
+### Different Types in branches
```
mofeing, line 67: "Type instability"? This is almost equal to the "type-stable vs instable" example in the Julia docs

https://docs.julialang.org/en/v1/manual/performance-tips/#Write-%22type-stable%22-functions

mofeing (approved):

wsmoses (approved):

avik-pal:
the setindex issue still needs fixing

mofeing:
> the setindex issue still needs fixing

If we merge this before fixing that, would you mind opening an issue so that we can keep track of that please?

avik-pal:
open #210 to track this, and #211 tracks compiling closures

PR #194, opened 2024-10-25
Add jit macro for simplified testing

wsmoses:


mofeing:
I was thinking about this the other day!

A posible improvement is to interpret directly the linearization, XLA call and delinearization to skip Julia comptime

wsmoses:
perhaps, but compilation here is inherently useful [and mostly for the sake of testing that we can compile a function without writing the double compile and run]

PR #195, opened 2024-10-25
complex DenseElementsAttribute

Pangoraw:


mofeing (approved):

src/mlir/IR/Type.jl
```diff
@@ -85,6 +85,13 @@ Creates a signless integer type of the given bitwidth in the context. The type i
 Type(T::Core.Type{<:Integer}; context::Context=context()) =
     Type(API.mlirIntegerTypeGet(context, sizeof(T) * 8))
 
+"""
+    Type(T::Core.Type{<:Complex}; context=context())
+
+Creates a complex type with the given element type.
+"""
+Type(T::Core.Type{<:Complex}; context=context()) = Type(API.mlirComplexTypeGet(Type(T.parameters[1])))
```
mofeing, line 93: It's better if you don't rely on internals like `.parameters`
```suggestion
Type(::Core.Type{Complex{T}}; context=context()) where {T} = Type(API.mlirComplexTypeGet(Type(T)))
```

mofeing (changes requested):
It's better if you don't rely on internals like `.parameters`

mofeing (approved):

PR #197, opened 2024-10-25
Make libMLIR_h automatically regenerate

Pangoraw:


Pangoraw:
The generation is missing stablehlo header files which are needed before merging

wsmoses:
@mofeing honestly I’m confused why the api calls weren’t here to begin with.

when these were in the Yggdrasil build it should’ve exported every symbol we built into Julia headers so that implies to be that the header rn is wrong somehow…

Pangoraw:
so essentially replace this line: 

https://github.com/EnzymeAD/Reactant.jl/blob/a4f685cad9514a0d31b0df38658dafb5ce46b8c9/src/mlir/MLIR.jl#L12

with 

```julia
include("path/to/Reactant_jll/libMLIR_h.jl")
```
?

mofeing:
Wait, you should just rebuild "libMLIR_h.jl" locally by running:

```
bazel build //:libMLIR_h.jl
```

I just forgot to include that one in the `make-dialects.jl` script, so it regenerates automatically.

Pangoraw:
Ok so the new `make-dialects.jl` should be enough ? it automatically runs on each pr ?

src/mlir/IR/Type.jl
```diff
@@ -90,7 +90,7 @@ Type(T::Core.Type{<:Integer}; context::Context=context()) =
 
 Creates a complex type with the given element type.
 """
-Type(::Core.Type{Complex{T}}; context=context()) where {T} = Type(API.mlirComplexTypeGet(Type(T)))
+Type(T::Core.Type{<:Complex}; context=context()) = Type(API.mlirComplexTypeGet(Type(T(im) |> real |> typeof)))
```
mofeing, line 93: This shouldn't be in this PR right?

deps/ReactantExtra/make-dialects.jl
```diff
@@ -16,7 +8,24 @@ for file in [
     )
     Base.Filesystem.cp(
         joinpath(@__DIR__, "bazel-bin", file),
-        joinpath(dirname(dirname(@__DIR__)), "src", "mlir", "Dialects", file);
+        output_path;
         force=true,
     )
 end
+
+src_dir = joinpath(@__DIR__, "..", "..", "src")
```
mofeing, line 16: It's better if you use `dirname` instead of `".."` so that it works in non-posix systems like Windows.

mofeing (changes requested):
This should be it. About your question, no: it doesn't run on each PR but on periodically every night and by manual dispatch. If you want, you can change the workflow to run on every PR but on 99% of the PRs it will do nothing.

EDIT: You can add a filter to the workflow such that it automatically runs on PRs that modify ReactantExtra.

src/mlir/IR/Type.jl
Pangoraw: #199

mofeing (approved):

PR #169, opened 2024-10-06
Implement `conj`, `conj!` for `TracedRArray`

mofeing:
Fixes #168

# To do

- [x] Implement `conj` for `TracedRNumber`
- [x] Implement in-place `conj!`
- [x] Write tests for `conj`, `conj!`
- [x] Implement `real`, `imag`
- [x] Write tests for `real`, `imag`
- [x] Implement adjoints of `real`, `imag` in Enzyme-JAX https://github.com/EnzymeAD/Enzyme-JAX/pull/145
  - [ ] update Enzyme-JAX commit on ReactantExtra

PR #199, opened 2024-10-27
prevent docs warning

Pangoraw:


mofeing:
Any idea why was giving warnings before? And what warnings was it showing?

Pangoraw:
This is the warning:

```
┌ Warning: Replacing docs for `Reactant.MLIR.IR.Type :: Union{Tuple{Type{Complex{T}}}, Tuple{T}} where T` in module `Reactant.MLIR.IR`
└ @ Base.Docs docs/Docs.jl:243
```

is it because the method key for docs is not the same as for actual methods ?

mofeing:
@Pangoraw I removed this code in #169 because that function was duplicated

PR #202, opened 2024-10-27
update libMLIR_h bindings and fix bindings update step n°2

### LINKED ISSUE #201: Error on regenerating "libMLIR_h.jl" on GitHub runner
The "Regenerate MLIR bindings" is currently failing when regenerating the "libMLIR_h.jl" file. Logs show that it fails on downloading the `GCCBootstrap` artifacts because it fails on creating a temporary directory in `~/.julia/artifacts/` due to the path being read-only.

https://github.com/EnzymeAD/Reactant.jl/actions/runs/11541259802/job/32123044918

```
 Downloading artifact: GCCBootstrap-x86_64-linux-gnu.v4.8.5.x86_64-linux-musl.unpacked
     Failure artifact: GCCBootstrap-x86_64-linux-gnu.v4.8.5.x86_64-linux-musl.unpacked
ERROR: LoadError: IOError: mktempdir("/home/runner/.julia/artifacts"): read-only file system (EROFS)
```

CC @Pangoraw

Pangoraw:
close #201 

- **fetch lazy artifacts in pre-process step**
- **update libMLIR_h.jl**

### DIFF
--- a/.github/workflows/regenerate-mlir-bindings.yml
+++ b/.github/workflows/regenerate-mlir-bindings.yml
@@ -22,7 +22,7 @@ jobs:
         run: |
           import Pkg
           Pkg.add("JuliaFormatter")
-      - run: julia --project=. -e 'using Pkg; pkg"instantiate"'
+      - run: julia --project=. -e 'using Pkg; Pkg.instantiate(); using Clang; Clang.JLLEnvs.get_system_includes()'
         working-directory: ./deps/ReactantExtra
       - run: julia make-bindings.jl
         working-directory: ./deps/ReactantExtra
--- a/src/mlir/libMLIR_h.jl
+++ b/src/mlir/libMLIR_h.jl
@@ -908,6 +908,17 @@ function mlirOpPrintingFlagsElideLargeElementsAttrs(flags, largeElementLimit)
 end
 
 """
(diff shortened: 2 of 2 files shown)


.github/workflows/regenerate-mlir-bindings.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         run: |
           import Pkg
           Pkg.add("JuliaFormatter")
-      - run: julia --project=. -e 'using Pkg; pkg"instantiate"'
+      - run: julia --project=. -e 'using Pkg; Pkg.instantiate(); using Clang; Clang.JLLEnvs.get_system_includes()'
```
mofeing, line 25: I don't understand why this should fix it?

Pangoraw: It downloads the lazy artifact before running inside the bazel sandbox

Pangoraw: New error https://github.com/EnzymeAD/Reactant.jl/actions/runs/11549131512/job/32141659071

mofeing:
@Pangoraw I think we can revert the change of this commit on "regenerate-mlir-bindings.yml" since I was getting the same error by just doing `]update` and `]precompile`.

mmm I believe the problem originates in `bazel` calling `julia` without the `--project` flag.

PR #204, opened 2024-10-28
Refactor regenerate-mlir-bindings.yml workflow

mofeing:
CC @pangoraw

### DIFF
--- a/.github/workflows/regenerate-mlir-bindings.yml
+++ b/.github/workflows/regenerate-mlir-bindings.yml
@@ -26,9 +26,8 @@ jobs:
         shell: julia --project=. --color=yes {0}
         run: |
           using Pkg
-          pkg"instantiate"
-          pkg"update"
-          pkg"precompile"
+          Pkg.instantiate()
+          Pkg.precompile()
           using Clang
           Clang.JLLEnvs.get_system_includes()
         working-directory: ./deps/ReactantExtra
--- a/.github/workflows/regenerate-mlir-bindings.yml
+++ b/.github/workflows/regenerate-mlir-bindings.yml
@@ -23,8 +23,8 @@ jobs:
           import Pkg
           Pkg.add("JuliaFormatter")
       - name: Instantiate dependencies
-        run: julia --project=. --color=yes {0}
-        shell: |
+        shell: julia --project=. --color=yes {0}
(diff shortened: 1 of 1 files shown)


.github/workflows/regenerate-mlir-bindings.yml
```diff
@@ -22,17 +22,22 @@ jobs:
...
           import Pkg
           Pkg.add("JuliaFormatter")
-      - run: julia --project=. -e 'using Pkg; pkg"instantiate"; pkg"update"; pkg"precompile"; using Clang; Clang.JLLEnvs.get_system_includes()'
+      - name: Instantiate dependencies
+        shell: julia --project=. --color=yes {0}
+        run: |
+          using Pkg
+          pkg"instantiate"
+          pkg"update"
+          pkg"precompile"
```
Pangoraw, line 31:
```
┌ Warning: The Pkg REPL mode is intended for interactive use only, and should not be used from scripts. It is recommended to use the functional API instead.
└ @ Pkg.REPLMode /opt/hostedtoolcache/julia/1.10.5/x64/share/julia/stdlib/v1.10/Pkg/src/REPLMode/REPLMode.jl:382
```
Also i think we dont want to update:
```suggestion
          Pkg.instantiate()
          Pkg.precompile()
```

mofeing: Yeah, I've never agreed with that message. I would argue that a script is an interactive use and that you shouldn't use it in a library which uses `Pkg`.
Anyway, we can change it. No problem.

mofeing: About update... I always call update by default. Update does more than the other commands and it has solved problems I had with the dependencies many times.

mofeing:
It seems that the problem is back (with this PR)?

https://github.com/EnzymeAD/Reactant.jl/actions/runs/11555990929/job/32162656313#step:7:318

```
ERROR: LoadError: ArgumentError: Package BinaryBuilderBase [7f725544-6523-48cd-82d1-3fa08ff4056e] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.
```

mofeing:
Up & running now!

PR #205, opened 2024-10-28
feat: support tracing scalars

avik-pal:
realized I need this to correctly compile conditionals with scalars assignments in the branches

```julia
using Reactant

x = (3, [3.14])

function f(x)
    return x[1] * x[2]
end

x_ra = Reactant.to_rarray(x; track_numbers=(Number,))

f2 = @compile f(x_ra)

f2(Reactant.to_rarray((5, [3.14]); track_numbers=(Number,)))
# 1-element ConcreteRArray{Float64, 1} with indices 1:
#  15.700000000000001

x_ra2 = Reactant.to_rarray(x)

f3 = @compile f(x_ra2)

f3(Reactant.to_rarray((5, [3.14])))
# 1-element ConcreteRArray{Float64, 1} with indices 1:
#  9.42
```

### DIFF
--- a/src/ConcreteRArray.jl
+++ b/src/ConcreteRArray.jl
@@ -93,36 +93,32 @@ function to_number(X::ConcreteRNumberType{T}) where {T}
     return data[]
 end
 
-function Base.convert(::Type{T}, x::ConcreteRNumberType{T}) where {T}
-    return to_number(x)
-end
+Base.convert(::Type{T}, x::ConcreteRNumberType{T}) where {T} = to_number(x)
 
-for jlop in (:(Base.isless), :(Base.:+), :(Base.:-), :(Base.:*), :(Base.:/), :(Base.:^))
+for jlop in (:(Base.isless), :(Base.:+), :(Base.:-), :(Base.:*), :(Base.:/), :(Base.:^)),
+    T in (ConcreteRNumber, ConcreteRArray{<:Any,0})
+
     @eval begin
-        function $jlop(x::ConcreteRNumberType{T}, y::ConcreteRNumberType{U}) where {T,U}
-            return $jlop(to_number(x), to_number(y))
-        end
-        function $jlop(x::ConcreteRNumberType{T}, y) where {T}
-            return $jlop(to_number(x), y)
-        end
-        function $jlop(x, y::ConcreteRNumberType{U}) where {U}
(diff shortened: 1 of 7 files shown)


src/Tracing.jl
```diff
@@ -400,8 +415,13 @@ function make_tracer(
 end
 
 function make_tracer(
-    seen, @nospecialize(prev::RT), @nospecialize(path), mode; kwargs...
-) where {RT<:AbstractFloat}
+    seen, @nospecialize(prev::RT), @nospecialize(path), mode; track_numbers=(), kwargs...
+) where {RT<:Number}
+    if mode == ArrayToConcrete
+        length(track_numbers) == 0 && return prev
+        should_convert = any(Base.Fix1(<:, RT), track_numbers)
```
wsmoses, line 422: is there a reason to not just have this be a boolean flag?

avik-pal: In the meeting notes we had this as:

```
x2 = to_rarray(x; track_intscalars=false, track_floatscalars=false) # if scalars (ints) were converted
```

wsmoses: ah yeah fair enough

wsmoses (approved):

src/ConcreteRArray.jl
```diff
@@ -4,11 +4,34 @@ end
...
+    data::T; client=XLA.default_backend[], idx=XLA.default_device_idx[]
+) where {T<:Number}
+    Base.depwarn(
+        "ConcreteRArray(data::Number) is deprecated, use ConcreteRNumber(data) instead",
+        :ConcreteRArray,
+    )
+    return ConcreteRArray(fill(data); client, idx)
+end
+
+const ConcreteRNumberType{T} = Union{ConcreteRArray{T,0},ConcreteRNumber{T}}
```
mofeing, line 34: small aesthetic suggestion: how about `ConcreteRScalar`?

avik-pal: Didn't we decide to not do TracedRScalar to match julia's naming of Number?

mofeing: exactly, that's why you have `ConcreteRNumber` and I'm fine with that.

by `*RScalar` I'm referring to this `Union{*RArray{T,0}, *RNumber{T}}`

PR #206, opened 2024-10-28
fix: unbreak precompilation

avik-pal:
precompile is currently broken due to repeated definition https://buildkite.com/julialang/reactant-dot-jl/builds/1100#0192d4c6-e438-4424-ba7e-9d3aa9aa6858/277-706


Pangoraw:
#199 should fix the problem

mofeing:
@avik-pal don't merge this PR

this is the original function definition that should be correct and the duplicated one was removed in #169

PR #216, opened 2024-11-02
feat: add `sync` to synchonize all result buffers

avik-pal:


src/Compiler.jl
```diff
@@ -691,7 +691,7 @@ function compile_xla(f, args; client=nothing, optimize=true)
     end
 end
 
-function compile(f, args; client=nothing, optimize=true)
+function compile(f, args; client=nothing, optimize=true, block_until_ready=false)
```
wsmoses, line 694: Can this also be passed to the macro?

also we should use a better name like async=fals or sync=true

avik-pal: done

avik-pal: oops broke something

mofeing:
I'm ok with this changes, but I don't think `XLA.synced_buffer` is working correctly (at least in macOS).

wsmoses (approved):

PR #218, opened 2024-11-03
fix complex attr c-api

mofeing:
CC @pangoraw

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -2,6 +2,7 @@
 #include "mlir-c/Support.h"
 
 #include "mlir/CAPI/IR.h"
+#include "mlir/CAPI/Wrap.h"
 #include "mlir/Pass/PassManager.h"
 
 #include "Enzyme/MLIR/Dialect/Dialect.h"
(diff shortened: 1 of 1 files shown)


mofeing:
I'm merging without waiting to tests because currently we have no workflow that checks for ReactantExtra build and I've checked it locally

deps/ReactantExtra/API.cpp
```diff
@@ -66,15 +67,27 @@ using namespace xla;
...
+  };                                                                           \
+  typedef struct name name
+
+DEFINE_C_API_STRUCT(MlirComplexType, const void);
+DEFINE_C_API_METHODS(MlirComplexType, mlir::ComplexType)
+
+#undef DEFINE_C_API_STRUCT
+
+MLIR_CAPI_EXPORTED MlirAttribute mlirComplexAttrDoubleGet(MlirContext ctx, MlirComplexType type, double real, double imag) {
     return wrap(complex::NumberAttr::get(unwrap(type), real, imag));
```
wsmoses, line 82: This seems unwise, shouldn't this have just been cast<mlir::ComplexType>(unwrap(a value of MlirType))

wsmoses: cc @mofeing @Pangoraw

mofeing: I don't know much about the LLVM C++ API and such, so no idea. `complex::NumberAttr::get(unwrap(type), real, imag)` (considering that `type` is of type `MlirType`) was giving problems because it couldn't cast `Type` to `ComplexType` so I improvised looking to some code here and there.

What's the relation between `unwrap` and `cast`? It's not very well explained (or at least I haven't found the docs).

wsmoses: Unwrap takes c API into regular c++ types 

cast lets us convert in the llvm c++ APi a value of one type to a value of another with checks

Issue #190, opened 2024-10-21
Resource exhaustion when compiling many functions


mofeing:
Some of my algorithms change its structure on every iteration, so I need to recompile the function on each iteration. This is sth that works well on Jax. Aside of the huge performance overhead we're now having in Reactant (which may need some refactoring on `Thunk` and `compile` to zero the amount of compilation from the Julia side to improve it), I'm getting the following error after running many compilations.

Some notes:
- It always fails on the same number of iterations, suggesting that each `compile` takes some resource that's it's not releasing and the resource size is constant.
- This resource size might be machine-dependent. I've just tried on macOS aarch64.

![Image](https://github.com/user-attachments/assets/9ea8abf0-75ff-4461-90df-6b53b8613114)

# MWE

```julia
julia> x = rand(2,2)
julia> y = Reactant.to_rarray(x)

julia> for _ in 1:10_000
           f = @compile sum(y)
       end
LLVM ERROR: pthread_create failed: Resource temporarily unavailable

[27943] signal (6): Abort trap: 6
in expression starting at REPL[18]:1
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 175047067 (Pool: 174829757; Big: 217310); GC: 115
fish: Job 1, 'julia +1.10' terminated by signal SIGABRT (Abort)
```

wsmoses:
can you trace in rr or soemthing with a debug jll to ascertain what resource is failing to be allocated?

wsmoses:
I never fully got things setup iirc, but also in principle [SetLogLevel](https://github.com/EnzymeAD/Reactant.jl/blob/ad508b99b4f4fed18d368885d457bcb7ad8f6d08/deps/ReactantExtra/API.cpp#L84C17-L84C28) might force a lot of debug to spew which may be helpful

mofeing:
Related https://github.com/jax-ml/jax/issues/16272

wsmoses:
Yeah we should also disable threading in our mlir contexts too -- I thought we fixed this earlier on the constructor for mlicontext, but maybe only just discussed it

PR #219, opened 2024-11-03
fix resource exhaustion bug #190

mofeing:
Disables multithreaded MLIR contexts as suggested in https://github.com/jax-ml/jax/issues/16272

the test needs to run `@compile` at least a number somewhat above 8000 times in macOS (I put 10,000), so it increases core test time from 2 minutes to... 7 minutes 😅

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -254,6 +254,7 @@ end
 
 function compile_mlir(f, args; kwargs...)
     ctx = MLIR.IR.Context()
+    MLIR.IR.enable_multithreading!(false; context=ctx)
     Base.append!(Reactant.registry[]; context=ctx)
     @ccall MLIR.API.mlir_c.RegisterDialects(ctx::MLIR.API.MlirContext)::Cvoid
     MLIR.IR.context!(ctx) do
@@ -662,6 +663,7 @@ end
 function compile_xla(f, args; client=nothing, optimize=true)
     # register MLIR dialects
     ctx = MLIR.IR.Context()
+    MLIR.IR.enable_multithreading!(false; context=ctx)
     append!(Reactant.registry[]; context=ctx)
     @ccall MLIR.API.mlir_c.RegisterDialects(ctx::MLIR.API.MlirContext)::Cvoid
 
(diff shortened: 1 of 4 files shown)


src/Compiler.jl
```diff
@@ -254,6 +254,7 @@ end
 
 function compile_mlir(f, args; kwargs...)
     ctx = MLIR.IR.Context()
```
wsmoses, line 256: the context constructor takes an optional argument to disable mutlithreadedd, we shouldn't call it in a follow up method

mofeing: do you mean `mlirContextCreateWithThreading` and `mlirContextCreateWithRegistry`? `MLIR.IR.Context` doesn't use them currently

wsmoses: yeah we should call mlirContextCreateWithThreading(false)

wsmoses: ideally actually yeah the registry and threading constructor

mofeing: done

mofeing:
@wsmoses what do we do about the test time problem? should I remove the test?

wsmoses (approved):

wsmoses:
@mofeing "DialectRegistry not defined"

wsmoses:
> @wsmoses what do we do about the test time problem? should I remove the test?

Honestly I'm fine either way. I'm okay not testing this as it's pretty fundamental for setup.

test/compile.jl
```diff
@@ -43,4 +43,17 @@ Base.sum(x::NamedTuple{(:a,),Tuple{T}}) where {T<:Reactant.TracedRArray} = (; a=
...
+        y = Reactant.to_rarray(x)
+        @test try
+            for _ in 1:10_000
+                f = @compile sum(y)
+            end
+            true
+        catch e
+            false
+        end
+    end
```
mofeing, line 58:
```suggestion

    # disabled due to long test time (core tests go from 2m to 7m just with this test)
    # @testset "resource exhaustation bug (#190)" begin
    #     x = rand(2, 2)
    #     y = Reactant.to_rarray(x)
    #     @test try
    #         for _ in 1:10_000
    #             f = @compile sum(y)
    #         end
    #         true
    #     catch e
    #         false
    #     end
    # end
```

PR #224, opened 2024-11-04
refactor: use `@jit` in the tests

avik-pal:


test/basic.jl
```diff
@@ -185,6 +158,12 @@ end
     x = randn(2, 3, 4)
     x_ca = Reactant.ConcreteRArray(x)
 
+    # XXX: @jit doesn't work with `;`
+    # @test @jit(mean(x_ca)) ≈ mean(x)
```
wsmoses, line 162: we should fix this

wsmoses (approved):

test/basic.jl
avik-pal: yes looking into it

PR #225, opened 2024-11-04
feat: make compile and code_hlo macros work with broadcasting

avik-pal:


test/bcast.jl
```diff
@@ -66,11 +66,10 @@ test()
     y_ca = Reactant.ConcreteRArray(y)
 
     @testset "Broadcasting" begin
-        # TODO make `@compile` work with broadcasting syntax
-        @test x .+ y ≈ Reactant.compile(.+, (x_ca, y_ca))(x_ca, y_ca)
-        @test x .- y ≈ Reactant.compile(.-, (x_ca, y_ca))(x_ca, y_ca)
-        @test x .* y ≈ Reactant.compile(.*, (x_ca, y_ca))(x_ca, y_ca)
-        @test x ./ y ≈ Reactant.compile(./, (x_ca, y_ca))(x_ca, y_ca)
+        @test x .+ y ≈ @jit((+).(x_ca, y_ca))
```
wsmoses, line 69: Would this also be able to fix something like @jit(x_ca .+ y_ca)

avik-pal: done

test/basic.jl
```diff
@@ -154,6 +154,15 @@ end
...
 
+@testset "Reactant.@code_hlo broadcasting" begin
+    x = Reactant.ConcreteRArray(randn(Float32, 2, 2))
+    y = Reactant.ConcreteRArray(randn(Float32, 2, 2))
+    res = Reactant.@code_hlo (.+)(x, y)
+    res_repr = sprint(show, res)
+
+    @test contains(res_repr, "stablehlo.add")
+end
+
```
mofeing, line 165: this would be better in `compile.jl` file

there are already to many different things in `basic.jl`

PR #230, opened 2024-11-06
feat: overload `mul!`

avik-pal:
partially addresses https://github.com/LuxDL/Lux.jl/issues/1025


src/TracedRArray.jl
```diff
@@ -381,30 +381,40 @@ for (jlop, hloop, hlocomp, merge) in
...
-    @nospecialize(lhs::TracedRArray{T,2}), @nospecialize(rhs::TracedRArray{T,2})
-) where {T}
-    lhsty = MLIR.IR.type(lhs.mlir_data)
-    rhsty = MLIR.IR.type(rhs.mlir_data)
-    resty = MLIR.IR.TensorType((size(lhs, 1), size(rhs, 2)), eltype(lhsty))
+function LinearAlgebra.mul!(
+    @nospecialize(C::TracedRArray{T1,2}),
+    @nospecialize(A::TracedRArray{T2,2}),
+    @nospecialize(B::TracedRArray{T3,2}),
+) where {T1,T2,T3}
```
wsmoses, line 388: Is it worth overriding the version with alpha and beta?

wsmoses: this is ine too tho, obviously

src/TracedRArray.jl
```diff
@@ -396,15 +408,17 @@ function Base.:*(
...
+            B.mlir_data;
             result_0=resty,
             dot_dimension_numbers=dot_dimension_numbers,
             precision_config=precar,
         ),
         1,
     )
-    return TracedRArray{T,2}((), res, (size(lhs, 1), size(rhs, 2)))
+    res = TracedRArray{T1,2}((), res, size(C))
+    @. C = C * β + res * α
```
wsmoses, line 420: This isn’t quite right since semantically if beta = 0 it must set c to a * b. Normally that seems like the case but if res is undefined memory or contains nan/inf that’s wrong

src/TracedRArray.jl
```diff
@@ -396,15 +408,25 @@ function Base.:*(
...
             precision_config=precar,
         ),
         1,
     )
-    return TracedRArray{T,2}((), res, (size(lhs, 1), size(rhs, 2)))
+    if iszero(β)
+        if isone(α)
+            C.mlir_data = res
+        else
+            C.mlir_data = (TracedRArray{T1,2}((), res, size(C)) .* α).mlir_data
```
wsmoses, line 423: doing a broadcast here probably won't optimize well for now (or rather we need to write optimizations on the batch op. Being able to driectly emit the elementwise mul of a broadcast will likely be a nontrivial perf win as a result

wsmoses (approved):
lgtm, but probably worth not using broadcast internally for perf reasons below

PR #240, opened 2024-11-07
feat: support setindex with views

avik-pal:


test/basic.jl
```diff
@@ -396,6 +396,43 @@ end
...
+    return y
+end
+
+function masking!(x)
+    x[1:2, :] .= 0
+    x[3:4, :] .= 1
+    return x
+end
+
+@testset "setindex with views" begin
```
mofeing, line 412: sorry for being annoying, but apparently there is another function called `Base.setindex` used in `Tuple`s

```suggestion
@testset "setindex! with views" begin
```

PR #247, opened 2024-11-08
feat: add support for `clamp` and `clamp!`

avik-pal:
- [x] needs testing

### DIFF
--- a/src/TracedRNumber.jl
+++ b/src/TracedRNumber.jl
@@ -270,15 +270,17 @@ Base.abs2(x::TracedRNumber{<:Real}) = x^2
 
 Base.log1p(x::TracedRNumber{T}) where {T} = log(x + one(T))
 
-function Base.clamp(x::TracedRNumber{T}, min::Number, max::Number) where {T}
-    min = promote_to(TracedRNumber{T}, min)
-    max = promote_to(TracedRNumber{T}, max)
-    return TracedRNumber{T}(
-        (),
-        MLIR.IR.result(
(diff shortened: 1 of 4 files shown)


src/TracedRNumber.jl
```diff
@@ -270,6 +270,17 @@ Base.abs2(x::TracedRNumber{<:Real}) = x^2
 
 Base.log1p(x::TracedRNumber{T}) where {T} = log(x + one(T))
 
+function Base.clamp(x::TracedRNumber{T}, min::Number, max::Number) where {T}
```
wsmoses, line 273: do we want to have versions where min and/or max are tracedvalues?

avik-pal: That already works

```julia
julia> using NNlib, Reactant

julia> x = Reactant.ConcreteRNumber(4.0)
ConcreteRNumber{Float64}(4.0)

julia> @code_hlo relu6(x)  # calls with TracedRNumber(0) and TracedRNumber(6)
Module:
  func.func @main(%arg0: tensor<f64>) -> tensor<f64> {
    %cst = stablehlo.constant dense<6.000000e+00> : tensor<f64>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
    %0 = stablehlo.clamp %cst_0, %arg0, %cst : tensor<f64>
    return %0 : tensor<f64>
  }
}
```

wsmoses (approved):

Issue #242, opened 2024-11-08
`getindex` assumes contiguous indexing


avik-pal:
```julia
julia> using Reactant

julia> xx = rand(4, 4) |> Reactant.to_rarray
4×4 ConcreteRArray{Float64, 2}:
 0.333112  0.765684  0.808021  0.197128
 0.533312  0.582735  0.454427  0.9106
 0.323209  0.509113  0.918799  0.997749
 0.355191  0.437561  0.620295  0.506551

julia> foo(x) = x[[1, 3, 4], :]
foo (generic function with 1 method)

julia> @code_hlo foo(xx)
ERROR: AssertionError: size(MLIR.IR.type(mlir_data)) == shape
Stacktrace:
  [1] Reactant.TracedRArray{Float64, 2}(paths::Tuple{}, mlir_data::Reactant.MLIR.IR.Value, shape::Tuple{Int64, Int64})
    @ Reactant /mnt/software/lux/Reactant.jl/src/TracedRArray.jl:14
  [2] getindex(::Reactant.TracedRArray{Float64, 2}, ::Vector{Int64}, ::Function)
    @ Reactant /mnt/software/lux/Reactant.jl/src/TracedRArray.jl:103
  [3] foo
    @ ./REPL[17]:1 [inlined]
  [4] (::Tuple{})(none::Reactant.TracedRArray{Float64, 2})
    @ Base.Experimental ./<missing>:0
  [5] (::Reactant.var"#26#35"{Bool, typeof(foo), Vector{ConcreteRArray{Float64, 2}}, Vector{Union{ReactantCore.MissingTracedValue, Reactant.TracedRArray, Reactant.TracedRNumber}}, Tuple{Reactant.TracedRArray{Float64, 2}}})()
    @ Reactant /mnt/software/lux/Reactant.jl/src/utils.jl:148
  [6] block!(f::Reactant.var"#26#35"{Bool, typeof(foo), Vector{ConcreteRArray{Float64, 2}}, Vector{Union{ReactantCore.MissingTracedValue, Reactant.TracedRArray, Reactant.TracedRNumber}}, Tuple{Reactant.TracedRArray{Float64, 2}}}, blk::Reactant.MLIR.IR.Block)
    @ Reactant.MLIR.IR /mnt/software/lux/Reactant.jl/src/mlir/IR/Block.jl:201
[...]

PR #252, opened 2024-11-09
feat: partial NNlib.gather support + better indexing support

### LINKED ISSUE #243: `getindex` assumes static indexing
```julia
julia> foo(x, idx) = x[idx, :]
foo (generic function with 2 methods)

julia> idx = Reactant.to_rarray([1, 2, 3])
3-element ConcreteRArray{Int64, 1}:
 1
 2
 3

julia> @code_hlo foo(xx, idx)
┌ Warning: Performing scalar indexing on task Task (runnable, started) @0x00007f3601e00010.
│ Invocation resulted in scalar indexing of a TracedRArray.
│ This is typically caused by calling an iterating implementation of a method.
│ Such implementations *do not* execute on device, but very slowly on the CPU,
│ and require expensive copies and synchronization each time and therefore should be avoided.
└ @ Reactant /mnt/software/lux/Reactant.jl/src/TracedRArray.jl:60
ERROR: MethodError: no method matching Int64(::Reactant.TracedRNumber{Int64})
The type `Int64` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  (::Type{T})(::T) where T<:Number
   @ Core boot.jl:900
  Int64(::Float16)
   @ Base float.jl:990
  Int64(::Float32)
   @ Base float.jl:990
  ...

Stacktrace:
  [1] (::Reactant.var"#227#232")(i::Reactant.TracedRArray{Int64, 1})
    @ Reactant ./none:0
  [2] iterate
    @ ./generator.jl:48 [inlined]
  [3] collect(itr::Base.Generator{Vector{AbstractVector{Int64}}, Reactant.var"#227#232"})
    @ Base ./array.jl:780
  [4] getindex(::Reactant.TracedRArray{Float64, 2}, ::Reactant.TracedRArray{Int64, 1}, ::Function)
    @ Reactant /mnt/software/lux/Reactant.jl/src/TracedRArray.jl:92
  [5] foo
    @ ./REPL[22]:1 [inlined]
  [6] (::Tuple{})(none::Reactant.TracedRArray{Float64, 2}, none::Reactant.TracedRArray{Int64, 1})
    @ Base.Experimental ./<missing>:0
  [7] (::Reactant.var"#26#35"{Bool, typeof(foo), Vector{ConcreteRArray}, Vector{Union{ReactantCore.MissingTracedValue, Reactant.TracedRArray, Reactant.TracedRNumber}}, Tuple{Reactant.TracedRArray{Float64, 2}, Reactant.TracedRArray{Int64, 1}}})()
    @ Reactant /mnt/software/lux/Reactant.jl/src/utils.jl:148
  [8] block!(f::Reactant.var"#26#35"{Bool, typeof(foo), Vector{ConcreteRArray}, Vector{Union{ReactantCore.MissingTracedValue, Reactant.TracedRArray, Reactant.TracedRNumber}}, Tuple{Reactant.TracedRArray{Float64, 2}, Reactant.TracedRArray{Int64, 1}}}, blk::Reactant.MLIR.IR.Block)
    @ Reactant.MLIR.IR /mnt/software/lux/Reactant.jl/src/mlir/IR/Block.jl:201
  [9] make_mlir_fn(f::Function, args::Vector{ConcreteRArray}, kwargs::Tuple{}, name::String, concretein::Bool; toscalar::Bool, return_dialect::Symbol, no_args_in_result::Bool, construct_function_without_args::Bool)
    @ Reactant /mnt/software/lux/Reactant.jl/src/utils.jl:112
 [10] make_mlir_fn
    @ /mnt/software/lux/Reactant.jl/src/utils.jl:36 [inlined]
 [11] #6
    @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:270 [inlined]
 [12] block!(f::Reactant.Compiler.var"#6#11"{typeof(foo), Vector{ConcreteRArray}}, blk::Reactant.MLIR.IR.Block)
    @ Reactant.MLIR.IR /mnt/software/lux/Reactant.jl/src/mlir/IR/Block.jl:201
 [13] #5
    @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:269 [inlined]
 [14] mmodule!(f::Reactant.Compiler.var"#5#10"{Reactant.MLIR.IR.Module, typeof(foo), Vector{ConcreteRArray}}, blk::Reactant.MLIR.IR.Module)
    @ Reactant.MLIR.IR /mnt/software/lux/Reactant.jl/src/mlir/IR/Module.jl:93
 [15] compile_mlir!(mod::Reactant.MLIR.IR.Module, f::Function, args::Vector{ConcreteRArray}; optimize::Bool)
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:266
 [16] compile_mlir!
    @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:265 [inlined]
 [17] #2
    @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:260 [inlined]
 [18] context!(f::Reactant.Compiler.var"#2#3"{@Kwargs{optimize::Bool}, typeof(foo), Vector{ConcreteRArray}}, ctx::Reactant.MLIR.IR.Context)
 [19] compile_mlir(f::Function, args::Vector{ConcreteRArray}; kwargs::@Kwargs{optimize::Bool})
    @ Reactant.Compiler /mnt/software/lux/Reactant.jl/src/Compiler.jl:258
 [20] top-level scope
    @ /mnt/software/lux/Reactant.jl/src/Compiler.jl:419
 [21] top-level scope
    @ none:1
```

Ideally we should default to `dynamic_slice` and have an optimization pass converting the dynamic_slice to slice if possible

avik-pal:
ideally we should be using `stablehlo.gather`, but this at least lets the currently fallback work

## Features

- Add partial support for `gather!` and `gather`. These are not the most efficient implementations but at least they work now.
- Non-contiguous indexing throws an error instead of giving incorrect results. See #242
- fixes #243. We emit `stablehlo.dynamic_slice` now

### DIFF
--- a/ext/ReactantNNlibExt.jl
+++ b/ext/ReactantNNlibExt.jl
@@ -296,6 +296,17 @@ function NNlib.make_causal_mask(x::AnyTracedRArray; dims::Int=2)
     )
 end
 
+# XXX: reevaluate this manual optimization once
+#      https://github.com/EnzymeAD/Enzyme-JAX/issues/164 is handled
+function NNlib.gather!(
+    dst::TracedRArray{T1,2},
+    src::AnyTracedRArray{T2,2},
+    idxs::Union{AbstractUnitRange{<:Number}},
+) where {T1,T2}
+    dst.mlir_data = src[:, idxs].mlir_data
+    return dst
+end
+
 function NNlib.gather!(
     dst::TracedRArray{T1,2}, src::AnyTracedRArray{T2,2}, idxs::AbstractVector{<:Number}
 ) where {T1,T2}
--- a/ext/ReactantNNlibExt.jl
+++ b/ext/ReactantNNlibExt.jl
@@ -315,13 +315,17 @@ function NNlib.gather!(
     idxs = (Reactant.promote_to(TracedRArray{Int,1}, idxs) .- 1).mlir_data
     slice_sizes = Reactant.promote_to(TracedRArray{Int,1}, [size(src, 1), 1]).mlir_data
 
-    dimension_numbers = """
-        #stablehlo.gather<
-            offset_dims = [0],
-            collapsed_slice_dims = [1],
-            start_index_map = [1],
-            index_vector_dim = 1>"""
-    dimension_numbers = parse(Reactant.MLIR.IR.Attribute, dimension_numbers)
+    #! format: off
+    dimension_numbers = MLIR.API.stablehloGatherDimensionNumbersGet(
+        MLIR.IR.context(),
+        Int64(1), Int64[0],
+        Int64(1), Int64[1],
+        Int64(0), Int64[],
+        Int64(0), Int64[],
+        Int64(1), Int64[1],
+        Int64(1)
+    )
+    #! format: on
 
     res = MLIR.IR.result(
         Reactant.MLIR.Dialects.stablehlo.dynamic_gather(
--- a/ext/ReactantNNlibExt.jl
+++ b/ext/ReactantNNlibExt.jl
@@ -296,14 +296,42 @@ function NNlib.make_causal_mask(x::AnyTracedRArray; dims::Int=2)
     )
 end
 
+function NNlib.gather!(
+    dst::TracedRArray{T1,2}, src::AnyTracedRArray{T2,2}, idxs::AbstractVector{<:Number}
+) where {T1,T2}
+    dims = NNlib.scatter_dims(src, dst, idxs)
+    @assert dims == 1  # scatter_dims lets us do some size checks so we call that function
+    idxs = (Reactant.promote_to(TracedRArray{Int,1}, idxs) .- 1).mlir_data
+    slice_sizes = Reactant.promote_to(TracedRArray{Int,1}, [size(src, 1), 1]).mlir_data
+
+    dimension_numbers = """
+        #stablehlo.gather<
+            offset_dims = [0],
+            collapsed_slice_dims = [1],
+            start_index_map = [1],
+            index_vector_dim = 1>"""
+    dimension_numbers = parse(Reactant.MLIR.IR.Attribute, dimension_numbers)
+
+    res = MLIR.IR.result(
+        Reactant.MLIR.Dialects.stablehlo.dynamic_gather(
+            src.mlir_data, idxs, slice_sizes; dimension_numbers
+        ),
+        1,
+    )
+    dst.mlir_data = res
+    return dst
+end
+
 # XXX: For performance to use `stablehlo.dynamic_gather` or atleast use traced loop
 #      instead of unrolling the loop (the case for AbstractArray can just use
-#      `stablehlo.gather`)
+#      `stablehlo.gather`). See above for the special case implementation that is optimized.
 function NNlib.gather!(
     dst::TracedRArray{T1,N}, src::AnyTracedRArray{T2,N}, idxs::AbstractArray
 ) where {T1,T2,N}
+    @warn "Using fallback implementation of `gather!` for using `stablehlo.dynamic_slice`. \
+           This case is not optimized and will be slow." maxlog = 1
     dims = NNlib.scatter_dims(src, dst, idxs)
-    colons = ntuple(i -> Colon(), dims)
+    colons = ntuple(Returns(Colon()), dims)
     start_sizes = ntuple(i -> size(src, i), dims)
     results = [reshape(src[colons..., Tuple(idxs[k])...], start_sizes..., :) for k in idxs]
     res = reshape(cat(results...; dims=(dims + 1)), size(dst))
--- a/ext/ReactantNNlibExt.jl
+++ b/ext/ReactantNNlibExt.jl
@@ -296,4 +296,19 @@ function NNlib.make_causal_mask(x::AnyTracedRArray; dims::Int=2)
     )
 end
 
+# XXX: For performance to use `stablehlo.dynamic_gather` or atleast use traced loop
+#      instead of unrolling the loop (the case for AbstractArray can just use
+#      `stablehlo.gather`)
+function NNlib.gather!(
+    dst::TracedRArray{T1,N}, src::AnyTracedRArray{T2,N}, idxs::AbstractArray
+) where {T1,T2,N}
+    dims = NNlib.scatter_dims(src, dst, idxs)
+    colons = ntuple(i -> Colon(), dims)
+    start_sizes = ntuple(i -> size(src, i), dims)
+    results = [reshape(src[colons..., Tuple(idxs[k])...], start_sizes..., :) for k in idxs]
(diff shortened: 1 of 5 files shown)


ext/ReactantNNlibExt.jl
```diff
@@ -296,4 +296,19 @@ function NNlib.make_causal_mask(x::AnyTracedRArray; dims::Int=2)
     )
 end
 
+# XXX: For performance to use `stablehlo.dynamic_gather` or atleast use traced loop
+#      instead of unrolling the loop (the case for AbstractArray can just use
+#      `stablehlo.gather`)
+function NNlib.gather!(
+    dst::TracedRArray{T1,N}, src::AnyTracedRArray{T2,N}, idxs::AbstractArray
+) where {T1,T2,N}
+    dims = NNlib.scatter_dims(src, dst, idxs)
```
wsmoses, line 305: can we emit a warning here at least in the interim?

src/TracedRArray.jl
```diff
@@ -65,49 +67,55 @@ Such implementations *do not* execute on device, but very slowly on the CPU,
...
-            limit_indices=MLIR.IR.DenseArrayAttribute([Int64(i) for i in index]),
-            strides=MLIR.IR.DenseArrayAttribute([Int64(1) for i in index]),
-        ),
-        1,
+        MLIR.Dialects.stablehlo.dynamic_slice(a.mlir_data, start_indices; slice_sizes), 1
     )
     res2 = MLIR.IR.result(
         MLIR.Dialects.stablehlo.reshape(
-            res1; result_0=MLIR.IR.TensorType(Int64[], eltype(MLIR.IR.type(res1)))
+            res1; result_0=MLIR.IR.TensorType(Int[], eltype(MLIR.IR.type(res1)))
```
wsmoses, line 78: are we sure this is int? I feel like the c api requires a 64-bit int....

We probably should add 32-bit CI to confirm

src/TracedRArray.jl
```diff
@@ -65,49 +67,55 @@ Such implementations *do not* execute on device, but very slowly on the CPU,
...
 end
 
 function Base.getindex(a::TracedRArray{T,0}) where {T}
     return TracedRNumber{T}((), a.mlir_data)
 end
 
+# XXX: We want to support https://github.com/EnzymeAD/Reactant.jl/issues/242 eventually
 function Base.getindex(a::TracedRArray{T,N}, indices::Vararg{Any,N}) where {T,N}
-    indices = [i isa Colon ? (1:size(a, idx)) : i for (idx, i) in enumerate(indices)]
+    indices = map(enumerate(indices)) do (idx, i)
```
wsmoses, line 92: yeah I think this should be doable with gather. That part I'm less confident we have all the optimization rules to lower into dynamic slice

ext/ReactantNNlibExt.jl
avik-pal: I dont think this is the right way to go. Even for a small testcase (nanoGPT) it takes forever to compile. Let me try to understand the stablehlo gather and get it fixed

wsmoses: 👍 , if you'd like feel free to separate the dynamic_slice stuff into a different PR

avik-pal: optimized the common cases and printing a warning for the other cases.

src/TracedRArray.jl
avik-pal: for safety reverted the change

ext/ReactantNNlibExt.jl
```diff
@@ -296,4 +296,60 @@ function NNlib.make_causal_mask(x::AnyTracedRArray; dims::Int=2)
...
+    idxs = (Reactant.promote_to(TracedRArray{Int,1}, idxs) .- 1).mlir_data
+    slice_sizes = Reactant.promote_to(TracedRArray{Int,1}, [size(src, 1), 1]).mlir_data
+
+    dimension_numbers = """
+        #stablehlo.gather<
+            offset_dims = [0],
+            collapsed_slice_dims = [1],
+            start_index_map = [1],
+            index_vector_dim = 1>"""
+    dimension_numbers = parse(Reactant.MLIR.IR.Attribute, dimension_numbers)
```
wsmoses, line 324: I guess that works, but we should be able to just call the relevant api here, no?

e.g. https://github.com/openxla/stablehlo/blob/4d85f95a737cbd97d769e7449ef8fac8138bae18/stablehlo/integrations/c/StablehloAttributes.h#L78

(this is also why I made the int64 comment before)

ext/ReactantNNlibExt.jl
```diff
@@ -296,4 +296,60 @@ function NNlib.make_causal_mask(x::AnyTracedRArray; dims::Int=2)
...
+    )
+    dst.mlir_data = res
+    return dst
+end
+
+# XXX: For performance to use `stablehlo.dynamic_gather` or atleast use traced loop
+#      instead of unrolling the loop (the case for AbstractArray can just use
+#      `stablehlo.gather`). See above for the special case implementation that is optimized.
+function NNlib.gather!(dst::TracedRArray, src::AnyTracedRArray, idxs::AbstractArray)
+    @warn "Using fallback implementation of `gather!` for using `stablehlo.dynamic_slice`. \
```
wsmoses, line 340: if you'd like you can put this behind a global to make sure it's only printed once (I think other indexing does that).

Though also I'm fine with having it always warn

wsmoses: ah that's what maxlog does!

src/TracedRArray.jl
```diff
@@ -65,49 +67,59 @@ Such implementations *do not* execute on device, but very slowly on the CPU,
 and require expensive copies and synchronization each time and therefore should be avoided."""
     )
 
+    start_indices = [promote_to(TracedRNumber{Int}, i - 1).mlir_data for i in index]
+    slice_sizes = [1 for _ in index]
```
wsmoses, line 71: my int64 worries remains here

src/TracedRArray.jl
```diff
@@ -65,49 +67,59 @@ Such implementations *do not* execute on device, but very slowly on the CPU,
...
+        # XXX: We want to throw error even for dynamic indexing
+        if typeof(a) <: Bool
+            contiguous || error("non-contiguous indexing is not supported")
+        end
+    end
+
+    start_indices = map(indices) do i
+        return promote_to(TracedRNumber{Int}, first(i) - 1).mlir_data
+    end
+    slice_sizes = [length(i) for i in indices]
```
wsmoses, line 110: and perhaps here

wsmoses (approved):

wsmoses:
yeah it looks like we never setup a 32bit jll anyways, but we should still try to make things match when possible

PR #255, opened 2024-11-09
implement `@trace` for

Pangoraw:


wsmoses:
you are the real mvp

Tho I mean everyone here really also is the mvp too <3

wsmoses:
As a high level comment. Is there a good way to add MLIR emission tests?

E.g. It would be great to do something which checks that the hlo generated by a given for loop exactly [or perhaps llvm filechek style] matches some given MLIR

wsmoses:
We have an example of this here for enzyme/llvm https://github.com/EnzymeAD/Enzyme.jl/blob/main/test/codegen/allocator.jl

src/ControlFlow.jl
```diff
@@ -74,9 +74,56 @@ function ReactantCore.traced_if(
     end
 end
 
-function get_region_removing_missing_values(compiled_fn, insertions)
+function ReactantCore.traced_while(cond_fn::CFn, body_fn::BFn, args) where {CFn, BFn}
```
mofeing, line 77: Like above, you're not using `CFn` nor `BFn` right?

```suggestion
function ReactantCore.traced_while(cond_fn, body_fn, args)
```

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -96,15 +96,72 @@ end
...
+        induction,
+        (body_symbols.assignments ∪ body_symbols.references)...,
+    )
+
+    reactant_code_block = quote
+        let $induction = Reactant.promote_to(Reactant.TracedRNumber{Int64}, $(range.args[2])),
+            args = $(all_syms)
+
+            cond_fn = $(all_syms) -> begin
+                $induction < Reactant.promote_to(Reactant.TracedRNumber{Int64}, $limit)
```
wsmoses, line 145: This is the right way to go for now.

However, it does mean that we only support loops like for i in 1:N

The whileop itself doesn't actually care about static sizes, or even about data-dependent control flow (we could iterate equally well with

```
while x < limit
 x = f(x)
 end
```

However to get good performance/differentiability we'd want to do some raising to canonicalize the loop into the i < N form, like you do now. Polygeist does have passes for this where I mentioned earlier, but they don't apply to stablehlo.while at the moment.

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -96,15 +96,72 @@ end
...
+
+    start = range.args[2]
+    step = length(range.args) == 3 ? 1 : range.args[3]
+    limit = range.args[end]
+
+    body_symbols = ExpressionExplorer.compute_symbols_state(quote
+        for _ = $(start):$(step):$(limit)
+            $body
+        end
+    end)
```
mofeing, line 131: do we only support ranges? what about iterators?

lib/ReactantCore/src/ReactantCore.jl
wsmoses: note that doing this will likely require both the "raising" ops [into the more canonical form you see at the moment, as well as things that, for example, get rid of unnecessary loop variables, multiple induction variables, etc

wsmoses: Again this has a lot of what we need, but not for stablehlo: https://github.com/llvm/Polygeist/blob/77c04bb2a7a2406ca9480bcc9e729b07d2c8d077/lib/polygeist/Passes/CanonicalizeFor.cpp#L662

lib/ReactantCore/src/ReactantCore.jl
wsmoses: see my comment above. Simultaneously requires more infra in enzyme-jax land to canonicalize

Pangoraw: The iterator interface is inherently type unstable so it might be trickier to support. Here `@trace for` is just syntactic sugar for the canonical while loop with known number of iterations currently supported in Enzyme-JAX.

lib/ReactantCore/src/ReactantCore.jl
Pangoraw: yeah, this is currently only `@trace for` but it should be relatively straightforward to have `@trace while` by reusing the `ReactantCore.traced_while` implementation from this PR

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -281,6 +338,13 @@ function traced_if(cond, true_fn::TFn, false_fn::FFn, args) where {TFn,FFn}
     return cond ? true_fn(args) : false_fn(args)
 end
 
+function traced_while(cond_fn::CFn, body_fn::BFn, args) where {CFn, BFn}
```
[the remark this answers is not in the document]
Pangoraw: thanks

Pangoraw:
> We have an example of this here for enzyme/llvm https://github.com/EnzymeAD/Enzyme.jl/blob/main/test/codegen/allocator.jl

Nice, it would be great to have FileCheck as part of a LLVM_full_jll or something so we write everything in the Julia test suite.

test/control_flow.jl
```diff
@@ -453,3 +454,50 @@ end
     @test @jit(condition12_compile_test(x_ra, y_ra, z_ra)) ≈
         condition12_compile_test(x, y, z)
 end
+
+function nnorm(x, n)
+    @trace for i in 1:n
```
wsmoses, line 459: can we do some tests with the body modifying a variable in place?

Also can we do a test where i is used outside the for loop, for example

```julia
i = 4
@trace for i in 1:n
   fdasfdsa
end
something(i)
```
and make sure this adheres to julia semantics

test/control_flow.jl
```diff
@@ -453,3 +454,50 @@ end
...
+       x = x * i ./ sum(x)
```
wsmoses, line 460: also some tests with break/continue

[and also returns, tho @avik-pal idr if we support that presently]

Pangoraw: I made it throw in loops as well for now (also for break/continue).

test/control_flow.jl
wsmoses: Can you also add a test with some more interesting ranges. E.g. 5:10. or 10:2:100

wsmoses (approved):
lgtm, but probably wait for @avik-pal to give it the go a head

Pangoraw:
CI seem to be crashing within the new tests, I will investigate

src/ControlFlow.jl
```diff
@@ -74,9 +74,70 @@ function ReactantCore.traced_if(
...
+    # TODO: detect and prevent mutation within the condition
+
+    # We promote all incoming args (is there a better way to do this?)
+    traced_args = [
+        if v isa Number && !(v isa TracedType)
+            Reactant.promote_to(TracedRNumber{typeof(v)}, v)
+        else
+            v
+        end for v in args
+    ]
```
avik-pal, line 89: I am trying to do it in #260, I can handle this there too

avik-pal (approved):

src/ControlFlow.jl
Pangoraw: should i wait for #260 and use ArrayToTracedUnsafe ?

avik-pal: No I think lets go with this. That PR has a segfault that I am trying to debug

PR #258, opened 2024-11-10
feat: more coverage for NNlib functions

avik-pal:


ext/ReactantNNlibExt.jl
```diff
@@ -20,31 +21,34 @@ for (jlop, hloop) in (
...
-# TODO handle non finite cases
 function NNlib.softmax!(out::TracedRArray{T,N}, x::AbstractArray; dims=1) where {T,N}
     max_ = NNlib.fast_maximum(x; dims)
-    #if all(isfinite, max_)
-    @fastmath out .= exp.(x .- max_)
-    #else
-    #    _zero, _one, _inf = T(0), T(1), T(Inf)
-    #    @fastmath @. out = ifelse(isequal(max_,_inf), ifelse(isequal(x,_inf), _one, _zero), exp(x - max_))
-    #end
+    @trace if all(isfinite, max_)
```
wsmoses, line 26: Already putting that to good use!

That said. I kind of wonder how far we are from the original nnlib impl at this point, and whether
it would make sense to put @trace in the NNlib implementation itself:

https://github.com/FluxML/NNlib.jl/blob/02138682a4fc5ca019759218be50e59907d4527c/src/softmax.jl#L60

avik-pal: I think we are mostly there. I am trying to debug an issue with fastmath annotations inside @trace if, but that should mostly do it

wsmoses (approved):

avik-pal:
something https://github.com/EnzymeAD/Reactant.jl/actions/runs/11767109998/job/32775434246?pr=258#step:9:1126 seems to have been broken, let's hold merging this

wsmoses:
> something https://github.com/EnzymeAD/Reactant.jl/actions/runs/11767109998/job/32775434246?pr=258#step:9:1126 seems to have been broken, let's hold merging this

cc @Pangoraw 

So this implies there's reverse mode happening, and we haven't fixed caching infra for it. I presume there's now reverse mode AD of control flow here (which isn't able to be fully removed via optimizations)?

ext/ReactantNNlibExt.jl
```diff
@@ -20,38 +21,44 @@ for (jlop, hloop) in (
...
-    @fastmath out .= x .- max_
-    # else
-    #     _zero, _minf, _inf = T(0), T(-Inf), T(Inf)
-    #     @. out = ifelse(
-    #         isequal(max_, _inf), ifelse(isequal(x, _inf), _zero, _minf), x - max_
-    #     )
-    # end
+    inf_num = Reactant.promote_to(TracedRNumber{T}, Inf)
+    zero_num = Reactant.promote_to(TracedRNumber{T}, 0)
+    @trace if all(isfinite, max_)
```
wsmoses, line 44: For now can you remove the traced if here and above (marking with a todo), then I presume this will work

Issue #189, opened 2024-10-20
Abstract interpretation over `ReactantInterpreter` sometimes skips overriden methods


mofeing:
There are times when interpreting over a `ReactantInterpreter` doesn't correctly replace the overriden methods in `ReactantMethodTable`. For example, https://github.com/EnzymeAD/Reactant.jl/issues/183#issuecomment-2425206624.

After doing a lil research, it seems that this behavior is due to inlining: apparently inlined functions cannot be "interpreted" (@vchuravy correct me if I'm wrong).

Solution is to add a method for `Core.Compiler.inlining_policy` on `ReactantInterpreter` that forbids inlining of the overriden methods.

Some useful links:
- https://github.com/compintell/Mooncake.jl/blob/faa63971859a4758034a1e77e47014b8a0e95130/src/interpreter/abstract_interpretation.jl#L158-L203
- https://github.com/EnzymeAD/Enzyme.jl/blob/9e945a5936f75eaab891361948148342c8f8772d/src/compiler/interpreter.jl#L250-L268
- https://github.com/JuliaLang/julia/blob/1fd7ada972911c181750d104604487a35ae3bea9/test/compiler/AbstractInterpreter.jl#L414-L433

wsmoses:
Oh yeah fair enough we should probably mark these as noinline here .

Cc @gbaraldi @aviatesk

vchuravy:
> After doing a lil research, it seems that this behavior is due to inlining: apparently inlined functions cannot be "interpreted"

Yeah that sounds wrong to me. How are you all executing the code?

mofeing:
> Yeah that sounds wrong to me. How are you all executing the code?

Just `Base.code_ircode` with our interpreter, wrap it into an `OpaqueClosure` and run.

https://github.com/EnzymeAD/Reactant.jl/blob/ad508b99b4f4fed18d368885d457bcb7ad8f6d08/src/utils.jl#L90-L101

The overlayed method for `Enzyme.autodiff` is in here (`@reactant_override` is just an alias for `@overlay ReactantMethodTable`): https://github.com/EnzymeAD/Reactant.jl/blob/ad508b99b4f4fed18d368885d457bcb7ad8f6d08/src/Interpreter.jl#L245-L247

aviatesk:
[...]

PR #259, opened 2024-11-10
Reactant interpreter overload

wsmoses:
hopefully resolves https://github.com/EnzymeAD/Reactant.jl/issues/189


src/utils.jl
```diff
@@ -120,13 +120,7 @@ function make_mlir_fn(
...
         end
 
-        # NOTE an `AbstractInterpreter` cannot process methods with more recent world-ages than it
-        # solution is to use a new interpreter, but we reuse the `code_cache` to minimize comptime in Julia <= 1.10
-        @static if !HAS_INTEGRATED_CACHE
-            interp = ReactantInterpreter(; code_cache=REACTANT_CACHE)
-        else
-            interp = ReactantInterpreter()
-        end
+        interp = ReactantInterpreter()
```
mofeing, line 123: mmm we must keep this for Julia 1.10, because the code cache first arrived to Julia 1.11

mofeing: if we remove this, tracing on Julia 1.10 is gonna slow down a lot

wsmoses: We still have the cache it’s just in the constructor

mofeing:
mmm okay, i see. so with this PR we would be using the interpreter from Enzyme, right?

wouldn't this be problematic? like what if I have a function which I have overlayed for Reactant but then I used that same function in Enzyme without Reactant, or viceversa.

I feel a opposed to removing out interpreter to keeps things clean and separate, but I want to understand it completely before judging this solution.

wsmoses:
> mmm okay, i see. so with this PR we would be using the interpreter from Enzyme, right?
> 

kind of. I revamped the enzyme interpreter to basically be an abstract interpreter with extensions, which is what we do here by providing the custom handle routine. Thus we can successfully override what we want while keeping the other overrides we still need (eg those for custom rules)

mofeing:
okay, let's talk about it on Wednesday's meeting. not against it but just wanna be sure that we don't lose anything.

mofeing (approved):
Okay, I see nothing wrong in this PR but I would like to add one note for the future: The problem is not in our own `ReactantIntepreter` but that **`AbstractIntepreter` currently doesn't support dynamic dispatch**.

Having our own `AbstractInterpreter` is the way to go, but while this is not supported and as we mainly (and almost only) use it for overlaying `Enzyme.autodiff` (which is critical for us), this PR solves the issue of missing method overlaying by moving the problem to Enzyme. We sacrifice method overlaying for other purposes but it's not something we have exploited a lot.

@wsmoses When method overlaying with dynamic dispatch is fixed, I think we should come back to having our `ReactantInterpreter` for better customization (but this can take a while).

wsmoses:
so partially yes, partially no. Longer term I think we will need something like this to make sure we correctly handle custom derivative rules [or at least have some sharing of the "rule override exists"]

mofeing:
> ... correctly handle custom derivative rules [...]

i think this could be handled by Reactant somehow if `AbstractInterpreter`s compose. anyway, this is kind of speculative and in the end, we'll have to adapt to whatever new changes get added to `AbstractInterpreter`.

mmm it looks like this PR breaks stuff and the "Basic grad cos" test is crashing.

wsmoses:
Yeah this Pr needs a release of enzyme and then rebase atop to run, which I’ll do today now that the plan is approved

PR #260, opened 2024-11-10
refactor: rework traced_if to avoid polluting make_mlir_fn

avik-pal:


avik-pal:
https://github.com/EnzymeAD/Reactant.jl/pull/565 newer and nicer attempt

PR #265, opened 2024-11-12
Move ReactantExtra to "lib/"

mofeing:


wsmoses:
Reactant currently places the jll code in the standard place in julia libraries: https://github.com/maleadt/LLVM.jl/tree/master/deps
https://github.com/EnzymeAD/Enzyme.jl/tree/main/deps

avik-pal:
We should keep the directory as is but rename `build_local.jl` to `build.jl`, then Pkg.build will work

```
(Reactant) pkg> ?build 
  build [-v|--verbose] pkg[=uuid] ...

  Run the build script in deps/build.jl for pkg and all of its dependencies in depth-first recursive order. If no packages are given, run the build scripts for all packages in the manifest. The -v/--verbose option redirects build output to stdout/stderr instead of the build.log file. The startup.jl file is disabled
  during building unless julia is started with --startup-file=yes.
```

wsmoses:
I will cc @vchuravy and @maleadt here since both LLVM.jl and Enzyme.jl call it build_local.jl and I don't have more context/reasoning why

mofeing:
> We should keep the directory as is but rename `build_local.jl` to `build.jl`, then Pkg.build will work

nope, that would trigger full compilation of ReactantExtra by any user that install Reactant. the build step is automatically the first time the package is installed on a env. check https://pkgdocs.julialang.org/v1/creating-packages/#Adding-a-build-step-to-the-package

> Reactant currently places the jll code in the standard place in julia libraries: [maleadt/LLVM.jl@master/deps](https://github.com/maleadt/LLVM.jl/tree/master/deps?rgh-link-date=2024-11-12T07%3A00%3A35Z)
[EnzymeAD/Enzyme.jl@main/deps](https://github.com/EnzymeAD/Enzyme.jl/tree/main/deps?rgh-link-date=2024-11-12T07%3A00%3A35Z)

aha, i see. then I will refactor this PR to just some cleaning.

PR #267, opened 2024-11-12
First batch of shortcuts to HLO ops

mofeing:
- [x] implement ops
- [x] test ops
- ~rewrite `TracedR*` methods~


src/Ops.jl
```diff
@@ -0,0 +1,430 @@
...
+) where {T,N}
+    res = MLIR.IR.result(
+        stablehlo.complex(
+            real.mlir_data,
+            imag.mlir_data;
+            result=mlir_type(TracedRArray{T,N}, size(real)),
+            location,
+        ),
+    )
+    return TracedRArray{T,N}((), res, size(real))
```
Pangoraw, line 411: I think this is missing

```suggestion
    return TracedRArray{Complex{T},N}((), res, size(real))
```

src/Ops.jl
```diff
@@ -0,0 +1,430 @@
...
+) where {T}
+    res = MLIR.IR.result(
+        stablehlo.complex(
+            real.mlir_data,
+            imag.mlir_data;
+            result=mlir_type(TracedRArray{T,0}, ()),
+            location,
+        ),
+    )
+    return TracedRNumber{T}((), res)
```
Pangoraw, line 429: Same as above

```suggestion
    return TracedRNumber{Complex{T}}((), res)
```

Pangoraw (commented):
nice!

mofeing:
# Update

I'm done with this first batch of ops. Now I will write some tests and it should be done.

mofeing:
> so CI apparently fails

should be fixed now

I've left some ops without testing due to bugs (mainly #196 but also some weird problem generating the IRCode for `Ops.polygamma`) or complex test configuration (like `send`, `recv`, ...)

wsmoses:
test_broken is fine if it doesn't crash, a commented or disabled test is also fine

mofeing:
then I will leave it commented because some crash

mofeing:
@wsmoses any idea of how to solve macOS-x86?

wsmoses (approved):

src/Ops.jl
```diff
@@ -0,0 +1,923 @@
...
+
+struct Token
+    mlir_data::MLIR.IR.Value
+end
+
+# constant ops
+function constant(
+    x::DenseArray{T,N};
+    location=MLIR.IR.Location(
+        "stablehlo.constant", MLIR.IR.Location(@__FILE__, @__LINE__, 0)
```
wsmoses, line 18: we still should refactor this into a location macro

```
macro location()
return MLIR.IR.Location(FILE, LINE)
end
```

which is then used everywhere (and can be overridden to be that more full back trace as mentioned

mofeing: yes, I have sth in mind but will open another PR

PR #268, opened 2024-11-12
feat: inherit scalar indexing functionality from GPUArraysCore

### LINKED ISSUE #232: Use `GPUArraysCore` for scalar indexing flags

avik-pal:
needs some tests before merging

## Example Usage

```julia
julia> using Reactant

julia> using GPUArraysCore

julia> x_ra = ConcreteRArray(rand(3, 4))
3×4 ConcreteRArray{Float64, 2}:
 0.166621  0.415209   0.23444   0.225489
 0.323775  0.201456   0.885111  0.625804
 0.22719   0.0906565  0.244437  0.98303

julia> x_ra[1]
0.1666208268454895

julia> GPUArraysCore.allowscalar(false)

julia> x_ra[1]
ERROR: Scalar indexing is disallowed.
Invocation of getindex(::ConcreteRArray, ::Vararg{Int, N}) resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] errorscalar(op::String)
   @ GPUArraysCore /mnt/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:151
 [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
   @ GPUArraysCore /mnt/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:124
 [4] assertscalar(op::String)
   @ GPUArraysCore /mnt/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:112
 [5] getindex(a::ConcreteRArray{Float64, 2}, args::Int64)
   @ Reactant /mnt/software/lux/Reactant.jl/src/ConcreteRArray.jl:175
 [6] top-level scope
   @ REPL[5]:1
 [7] top-level scope
   @ none:1

julia> @allowscalar x_ra[1]
0.1666208268454895
```

On CPU no error is ever thrown unless the user manually opts in for no-scalar indexing.

fixes #232


avik-pal:
@mofeing can you check if this helps your case where you saw the scalar indexing warnings

mofeing:
we confirm that this remove the infinite warnings we had in our code. Thanks @avik-pal!

I would approve the PR but it seems like this is breaking the tests?

avik-pal:
The x86 ones are broken since we dont have the binaries in-place.

But I still need to add some tests before merging

src/Reactant.jl
```diff
@@ -110,12 +111,19 @@ function __init__()
 end
 
 function set_default_backend(backend::XLA.Client)
+    if backend === XLA.backends["cpu"]
```
wsmoses, line 114: so this won't quite work because we can end up with both cpu and gpu tensors

src/ConcreteRArray.jl
```diff
@@ -193,16 +192,8 @@ function Base.getindex(a::ConcreteRArray{T}, args::Vararg{Int,N}) where {T,N}
...
-        @warn(
-            """Performing scalar get-indexing on task $(current_task()).
-    Invocation resulted in scalar indexing of a ConcreteRArray.
-    This is typically caused by calling an iterating implementation of a method.
-    Such implementations *do not* execute on device, but very slowly on the CPU,
-    and require expensive copies and synchronization each time and therefore should be avoided."""
-        )
-        getindex_warned[] = true
-    end
+
```
wsmoses, line 195: Can we have this call a function of our own which calls gpuarrays assertscalar if it's loaded

src/Reactant.jl
mofeing: `TracedRArray` doesn't know about the backend accelerator, be CPU, GPU or TPU. actually, HLO dialects neither know about which backend are they gonna run.
@wsmoses correct me if I'm wrong but that step is done later in XLA when compiling HLO to native executable

wsmoses: I mean even if they did [and you're right they don't] we should err for traced

mofeing: I asked to remove for CPU because they don't make much sense in there and they pollute a loooot the stdout.

wsmoses: i think removing for cpu concretearray is fine, but the problem is that it will equally pollute the IR we compile on traced, so we should still warn (or allowscalar)

src/ConcreteRArray.jl
avik-pal: We can, but it is extremely lightweight:

```julia
julia> @time_imports using GPUArraysCore
      0.2 ms  Adapt
      0.4 ms  GPUArraysCore
```

src/Reactant.jl
avik-pal: Changed the behavior to (default CUDA behavior):

1. Allowed by warn in REPL
2. Disallowed with error in scripts
3. Can be locally allowed without warning using `@allowscalar`

mofeing: should we reexport allowscalar?

avik-pal: +1

src/ConcreteRArray.jl
wsmoses: eh okay I'm fine with this then

test/basic.jl
```diff
@@ -16,7 +14,7 @@ using InteractiveUtils
 
     a = Reactant.ConcreteRArray(x)
 
-    c_res = sum(a)
+    c_res = @allowscalar sum(a)
```
wsmoses, line 17: Okay ideally this shouldn't be required. I feel like a load of a concretenumber/tracednumber itself should be automatically allowscalar

avik-pal: These work fine if the backend is CPU but the default implementation of sum will just loop over the indices which fails the GPU ci

wsmoses: Wait really, shouldn’t it fall back to a reduce?

if not this is definitely a bug

wsmoses: Oh sorry this is for the concretearray not traced

yeah we should still eventually make this a reduce, but for another time

wsmoses (approved):

PR #270, opened 2024-11-12
chore: bump ReactantCore version

avik-pal:
We need to keep these in-sync for new functionality


Project.toml
```diff
@@ -39,7 +39,7 @@ LinearAlgebra = "1.10"
 NNlib = "0.9"
 OrderedCollections = "1"
 Preferences = "1.4"
-ReactantCore = "0.1.1"
+ReactantCore = "0.1.2"
```
mofeing, line 42: Do we really need to fix the patch version? shouldn't just putting `0.1` work?

avik-pal: ideally we setup https://discourse.julialang.org/t/psa-add-downgrade-ci-to-better-check-version-compatibility/110063. It ensures that minimum versions are always in an installable state

mofeing: uh I always wondered if sth like this was actually easily doable
would you mind opening an issue for this? I can do it some other day

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -364,7 +364,7 @@ function traced_if(cond, true_fn::TFn, false_fn::FFn, args) where {TFn,FFn}
     return cond ? true_fn(args) : false_fn(args)
 end
 
-function traced_while(cond_fn, body_fn, args) where {CFn,BFn}
```
mofeing, line 367: Can you do the same with `traced_if` just above?

mofeing (approved):

Project.toml
mofeing: small doubt: setting the version to `"0.1.2"` forces Julia to use that specific version or does it allow to use higher versions within the same minor version? e.g. would it be posible then to use v0.1.3?

avik-pal: It allows that version and higher <0.2

avik-pal: Any 0.0.* forces a specific version similar to how enzyme jll is fixed to a specific release

PR #272, opened 2024-11-13
fix: add GPU backend as default if available

avik-pal:


avik-pal:
Lol some of the GPU tests were passing because we forgot to actually run them on GPU :sweat:

wsmoses (approved):

PR #276, opened 2024-11-14
docs: add note on installing cuda version of Reactant on login nodes

avik-pal:


wsmoses:
Hmm why is this needed?

In other words what errs presently

avik-pal:
If you don't have internet and it tries to download the correct binary it will fail.

avik-pal:
same as https://cuda.juliagpu.org/stable/installation/overview/#Precompiling-CUDA.jl-without-CUDA

avik-pal:
login node gets cpu version. compute node tries to get cuda version (correctly so), and it cant download

PR #274, opened 2024-11-14
Use gensym-ed variable in `@compile` and `@jit`

jumerckx:
This should fix #237.

PR #281, opened 2024-11-15
gensym in `@code_hlo`

jumerckx:
I missed the `@code_hlo` in my previous pr (#274).
I've now deduplicated the code for `@compile`, `@jit`, and `@code_hlo`.
Option kwargs passed to `@code_hlo` where handled slightly differently compared to the other two macros, now they're handled the same but maybe someone should check whether `@code_hlo` kwarg handling does what it's supposed to do.

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -430,7 +430,8 @@ end
     @compile f(args...)
 """
 macro compile(args...)
-    return esc(first(compile_call_expr(__module__, args...)))
+    default_options = Dict{Symbol,Any}(:optimize => true, :sync => false)
+    return esc(first(compile_call_expr(__module__, compile, default_options, args...)))
 end
 
 """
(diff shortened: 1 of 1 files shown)


src/Compiler.jl
```diff
@@ -383,54 +383,23 @@ end
...
-            f = Base.Broadcast.BroadcastFunction($(call.args[1]))
-            args = $(call.args[2:end]...)
-            mode = first($(compile_mlir)(f, args; optimize=options.optimize))
-            mode
-        end
-    else
-        error("Invalid function call: $(call)")
-    end
-    return esc(expr)
+    return esc(first(compile_call_expr(__module__, compile_mlir, default_options, args...)))
```
wsmoses, line 394: remove this extra return?

jumerckx: oops

wsmoses (approved):

PR #282, opened 2024-11-16
test for traced_type where array isn't captured by type param

jumerckx:
Test to check that:
1. `traced_type` for `ArrayToConcrete` fails if there are array-typed fields whos type isn't captured by the type parameter list
2. The error isn't a stackoverflow because recursion is properly handled.

### DIFF
--- a/test/tracing.jl
+++ b/test/tracing.jl
@@ -1,5 +1,12 @@
 using Reactant
-using Reactant: traced_type, ConcreteRArray, TracedRArray, ConcreteToTraced, ArrayToConcrete
+using Reactant:
+    traced_type,
+    ConcreteRArray,
+    TracedRArray,
+    ConcreteToTraced,
+    ArrayToConcrete,
+    NoFieldMatchError,
+    TracedTypeError
 using Test
 
 @testset "Tracing" begin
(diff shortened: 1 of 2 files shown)


test/tracing.jl
```diff
@@ -99,6 +99,15 @@ using Test
                 )
             end
         end
+        @testset "mode = ArrayToConcrete" begin
+            struct Node
+                x::Vector{Float64}
+                y::Union{Nothing,Node}
+            end
```
wsmoses, line 106: While we’re doing this perhaps we should make a custom error type and we can test for that?

jumerckx: Can do

jumerckx: I added extra explanation to the error but can get rid of it/change it if wanted?
```
ERROR: NoFieldMatchError: Cannot convert type Node, best attempt Node failed.
This could be because the type does not capture the fieldtypes that should be converted in its type parameters.
Stacktrace:
 [1] traced_type(::Type{Node}, seen::Reactant.OrderedIdDict{Any, Any}, mode::Val{Reactant.ArrayToConcrete})
   @ Reactant ~/Reactant.jl/src/Tracing.jl:174
 [2] top-level scope
   @ ~/Reactant.jl/experiments/main.jl:7
```

wsmoses:
hm tests fail here for some reason

jumerckx:
I'd added a check to not let `Core.apply_type(T.name.wrapper, subParms...)` error in cases where there are no subParms, but the check was not precise enough.
AFAICT if there are no subParms at that point, conversion has certainly failed so the error could be thrown earlier perhaps, but this only happens for invalid types so I don't think it's important to optimize this.

PR #283, opened 2024-11-16
fix: don't print "Module"

avik-pal:
Messes exporting stablehlo see https://github.com/LuxDL/Lux.jl/pull/1088/files#r1845238905


wsmoses:
Can we add a test to make sure this doesn't return?

wsmoses (approved):

PR #287, opened 2024-11-18
fix semantics of `eltype` on `TracedRNumber`

mofeing:
`Base.eltype` on `TracedRNumber` is semantically wrong because `TracedRNumber`, as a `Number`, should return its own type, not the interior type. For example,

```julia
julia> eltype(ComplexF64)
ComplexF64 (alias for Complex{Float64})

julia> eltype(Reactant.TracedRNumber{ComplexF64})
ComplexF64 (alias for Complex{Float64})
```

This behavior is giving some problems when using Yao to get the matrix representation of a parametrized quantum gate when the parameter is traced using `TracedRNumber`.

```julia
julia> using Reactant, YaoBlocks

julia> circ = chain(1, put(1 => H), put(1 => Rz(p)))
nqubits: 1
chain
├─ put on (1)
│  └─ H
└─ put on (1)
   └─ rot(Z, TracedRNumber{ComplexF64}(()))

# before
julia> parameters_eltype(circ)
ComplexF64 (alias for Complex{Float64})

# after
julia> parameters_eltype(circ)
Reactant.TracedRNumber{ComplexF64}
```

### DIFF
--- a/src/ControlFlow.jl
+++ b/src/ControlFlow.jl
@@ -142,10 +142,12 @@ function get_region_removing_missing_values(compiled_fn, insertions)
     return_op = MLIR.IR.terminator(block)
     for (i, rt) in insertions
         if rt isa TracedRNumber
-            attr = MLIR.IR.DenseElementsAttribute(Array{eltype(rt)}(undef, ()))
+            attr = MLIR.IR.DenseElementsAttribute(Array{concrete_eltype(rt)}(undef, ()))
             op = MLIR.Dialects.stablehlo.constant(; value=attr)
         elseif rt isa TracedRArray
-            attr = MLIR.IR.DenseElementsAttribute(Array{eltype(rt)}(undef, size(rt)))
+            attr = MLIR.IR.DenseElementsAttribute(
+                Array{concrete_eltype(rt)}(undef, size(rt))
+            )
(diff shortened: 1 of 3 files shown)


avik-pal (approved):

mofeing:
seems like the error is somewhere else? did we broke sth @wsmoses?

mofeing:
@avik-pal any idea of why this breaks so many tests?

mofeing:
in today's meeting and following what I wrote in https://github.com/EnzymeAD/Reactant.jl/issues/308#issuecomment-2504777208, I see 3 different `eltype` implementations:

- `traced_eltype`: return `TracedRNumber{T}` on a `TracedRArray{T}`. otherwise, call `eltype`.
- `concrete_eltype`: return `ConcreteRNumber{T}` on a `ConcreteRArray{T}`. otherwise, call `eltype`.
- regular `Base.eltype`: return `T` from `RArray{T}`

in general regular `eltype` should be used to fix problems with idiomatic Julia libraries, but the definition of `eltype` on an array is that the type of calling `iterate` on the array should be returned (e.g. calling `x[1]` where `x isa ConcreteRArray{Float64,1}` returns a `Float64`, not a `ConcreteRNumber{Float64}`, and the same should be expected from `TracedRArray`)

mofeing:
closing in favor of #426

PR #292, opened 2024-11-19
chore: specify sources for local dev

avik-pal:
Makes sure that we auto-use the subdirectory package during development. Has no effect when the Project is not directly loaded (and no end user is affected)


wsmoses:
I don’t fully understand this so @gbaraldi can you review this?

mofeing:
@wsmoses I think @avik-pal refers to a new Julia 1.11 feature in Pkg.jl https://pkgdocs.julialang.org/dev/toml-files/#The-%5Bsources%5D-section

PR #294, opened 2024-11-19
feat: define aos_to_soa

avik-pal:
Allows for a hook to ensure that we never end up propagating Array of Traced Values.

Related example for Tracker.jl where propagating Array{<:TrackedReal} leads to terrible performance https://github.com/LuxDL/Lux.jl/blob/cb0900f5350541a45e01af9ef12cc3d3e8df65d3/lib/LuxCore/ext/LuxCoreArrayInterfaceTrackerExt.jl#L8-L16

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -15,6 +15,8 @@ import ..Reactant:
     TracedType
 
 @inline traced_getfield(@nospecialize(obj), field) = Base.getfield(obj, field)
+@inline traced_getfield(@nospecialize(obj::AbstractArray), field) =
+    Base.getindex(obj, field)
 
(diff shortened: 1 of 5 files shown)


src/Compiler.jl
```diff
@@ -15,6 +15,8 @@ import ..Reactant:
     TracedType
 
 @inline traced_getfield(@nospecialize(obj), field) = Base.getfield(obj, field)
+@inline traced_getfield(@nospecialize(obj::AbstractArray), field) =
```
avik-pal, line 18: This seems to break some of the tests. The issue here was that if we pass an `Array{<:ConcreteRNumber}` then getfield will give us a GenericMemoryRef instead of the values

Issue #301, opened 2024-11-22
Incorrect code-generation for `@trace for ...`


avik-pal:
```julia
@macroexpand @trace for i in 1:10
    dx = 1
end
```

```julia
quote
    #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:183 =#
    if any(ReactantCore.is_traced, (dx,))
        #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:184 =#
        begin
            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:160 =#
            let args = (Reactant.promote_to(Reactant.TracedRNumber{Int}, 0), dx)
                #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:161 =#
                cond_fn = ((var"##i#957892", dx)->begin
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:162 =#
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:163 =#
                            local num_iters = div(10 - 1, 1, RoundDown)
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:164 =#
                            local num_iters = Reactant.promote_to(Reactant.TracedRNumber{Int64}, num_iters)
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:167 =#
                            var"##i#957892" < num_iters + 1
                        end)
                #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:169 =#
                body_fn = ((var"##i#957892", dx)->begin
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:170 =#
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:171 =#
                            local step_ = 1
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:172 =#
                            local start_ = 1
                            #= /mnt/software/lux/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:173 =#
                            local i = start_ + var"##i#957892" * step_
[...]

PR #303, opened 2024-11-25
Handle new variables being created inside a traced loop

jumerckx:
RE #301

### DIFF
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -157,16 +157,17 @@ function trace_for(mod, expr)
         Expr(:(=), def, MissingTracedValue()) for
         (def, arg) in zip(defs, potentially_undefined)
     ]
-    assigns = [
-        Expr(:(&&), (Expr(:isdefined, arg), Expr(:(=), def, arg))) for
-        (def, arg) in zip(defs, potentially_undefined)
-    ]
-
     updates = []
+    assigns = []
     for (def, arg) in zip(defs, potentially_undefined)
         append!(updates, (
             quote
                 !($def isa $(MissingTracedValue)) && ($def = $arg)
+            end
+        ).args)
+        append!(assigns, (
+            quote
+                (@isdefined $arg) && ($def = $arg)
             end
         ).args)
     end
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -147,20 +147,28 @@ function trace_for(mod, expr)
             $body
         end,
     )
-    
+
     filter!(∉(SPECIAL_SYMBOLS), body_symbols.assignments)
     filter!(∉(SPECIAL_SYMBOLS), body_symbols.references)
 
     potentially_undefined = setdiff(body_symbols.assignments, body_symbols.references)
     defs = [gensym(arg) for arg in potentially_undefined]
-    inits = [Expr(:(=), def, MissingTracedValue()) for (def, arg) in zip(defs, potentially_undefined)]
-    assigns = [Expr(:(&&), (Expr(:isdefined, arg), Expr(:(=), def, arg))) for (def, arg) in zip(defs, potentially_undefined)]
-    
+    inits = [
+        Expr(:(=), def, MissingTracedValue()) for
+        (def, arg) in zip(defs, potentially_undefined)
+    ]
+    assigns = [
+        Expr(:(&&), (Expr(:isdefined, arg), Expr(:(=), def, arg))) for
+        (def, arg) in zip(defs, potentially_undefined)
+    ]
+
     updates = []
     for (def, arg) in zip(defs, potentially_undefined)
-        append!(updates, (quote
-                            !($def isa $(MissingTracedValue)) && ($def = $arg)
-                        end).args)
+        append!(updates, (
+            quote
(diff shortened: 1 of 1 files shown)


lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -8,6 +8,9 @@ export @trace, MissingTracedValue
 # Traits
 is_traced(x) = false
 
+# Special value to signify that a loop variable is uninitialized
+struct LoopInitializer end
```
avik-pal, line 12: Can we re-use MissingTracedValue here? It has similar behavior for if statements where one branch can have missing variables

jumerckx: Ah I was wondering about this, made the change.

jumerckx:
CI succeeds (?) but running the tests locally, I get a failure for the sinkhorn control flow test.

```
error: expect operands to be compatible with body block arguments but got 'tensor<i64>', 'tensor<5xf32>', 'tensor<5xf32>', 'tensor<10xf32>', 'tensor<10x5xf32>', 'tensor<10xf32>' vs 'tensor<5xf32>', 'tensor<i64>', 'tensor<5xf32>', 'tensor<10xf32>', 'tensor<10x5xf32>', 'tensor<10xf32>'
ERROR: "failed to run pass manager on module"
Stacktrace:
  [1] run!
    @ ~/Reactant.jl/src/mlir/IR/Pass.jl:79 [inlined]
  [2] run_pass_pipeline!(mod::Reactant.MLIR.IR.Module, pass_pipeline::String; enable_verifier::Bool)
    @ Reactant.Compiler ~/Reactant.jl/src/Compiler.jl:256
  [3] run_pass_pipeline!
    @ ~/Reactant.jl/src/Compiler.jl:250 [inlined]
  [4] compile_mlir!(mod::Reactant.MLIR.IR.Module, f::Function, args::Tuple{…}; optimize::Bool)
    @ Reactant.Compiler ~/Reactant.jl/src/Compiler.jl:300
```
The block arguments of `body_fn` seem to be in the wrong order. Is the sinkhorn test run by CI? If yes, my local setup might just not be fully up to date or something.
```julia
cond_fn = ((var"##i#444179", var"##v#444181", ν, μ, K, u) -> begin
    local num_iters = div(10 - 1, 1, RoundDown)
    local num_iters = Reactant.promote_to(Reactant.TracedRNumber{Int64}, num_iters)
    var"##i#444179" < num_iters + 1
end)
body_fn = ((var"##i#444179", var"##v#444181", ν, μ, K, u) -> begin
    local step_ = 1
    local start_ = 1
    local _ = start_ + var"##i#444179" * step_
    begin
        v = ν ./ (K' * u)
        u = μ ./ (K * v)
    end
    !(var"##v#444181" isa ReactantCore.MissingTracedValue) && (var"##v#444181" = v)
    (var"##i#444179" + 1, var"##v#444181", ν, μ, K, u)
end)
```
```mlir
%12:6 = "stablehlo.while"(%11, %10, %1, %0, %7#0, %9) ({
^bb0(%arg9: tensor<i64>, %arg10: tensor<5xf32>, %arg11: tensor<5xf32>, %arg12: tensor<10xf32>, %arg13: tensor<10x5xf32>, %arg14: tensor<10xf32>):
  // ...
  "stablehlo.return"(%64) : (tensor<i1>) -> ()
}, {
^bb0(%arg3: tensor<5xf32>, %arg4: tensor<i64>, %arg5: tensor<5xf32>, %arg6: tensor<10xf32>, %arg7: tensor<10x5xf32>, %arg8: tensor<10xf32>):
  // ...
  "stablehlo.return"(%60, %44#0, %arg5, %arg6, %arg7, %58#0) : (tensor<i64>, tensor<5xf32>, tensor<5xf32>, tensor<10xf32>, tensor<10x5xf32>, tensor<10xf32>) -> ()
}) : (tensor<i64>, tensor<5xf32>, tensor<5xf32>, tensor<10xf32>, tensor<10x5xf32>, tensor<10xf32>) -> (tensor<i64>, tensor<5xf32>, tensor<5xf32>, tensor<10xf32>, tensor<10x5xf32>, tensor<10xf32>)
```

Pangoraw (commented):
I have another possible solution in https://github.com/Pangoraw/Reactant.jl/commit/32853ebb0c25f69dd09956019950027001ffd592. I think there might be edge cases though.

jumerckx:
> I have another possible solution in https://github.com/Pangoraw/Reactant.jl/commit/32853ebb0c25f69dd09956019950027001ffd592. I think there might be edge cases though.

Thanks for the review, I like the simplicity of your approach way better, though. I can close this pr and we can go with your branch.

One consideration I have is for `cond_val(s) = :(isdefined($(mod), $(QuoteNode(s))) ? $s : nothing)`. This works for global variables but doesn't pick up function-local variables, I think?
There exists a special expression type for checking isdefined, it seems:
```
julia> @macroexpand @isdefined var
:($(Expr(:isdefined, :var)))
```

Pangoraw:
>  There exists a special expression type for checking isdefined, it seems:

oh right, I missed that one. thank you! I opened #310. I think it would maybe better to use the `MissingTracedValue` there instead like in your pr to represent undefined values.

jumerckx:
> I think it would maybe better to use the MissingTracedValue
Yeah agreed, it doesn't change much but seems a bit nicer.

jumerckx:
closed in favor of #310

PR #305, opened 2024-11-26
implement `collect` for `TracedRArray`

mofeing:
we had an issue in which some inner function was calling `collect` on a `TracedRArray`, which would error.

this PR implements `collect(::TracedRArray)` as just a copy of the `TracedRArray`

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -160,7 +160,6 @@ Base.Tuple(x::TracedRArray) = ntuple(Base.Fix1(Base.getindex, x), length(x))
 Base.size(x::TracedRArray) = x.shape
 
 Base.copy(A::TracedRArray{T,N}) where {T,N} = TracedRArray{T,N}((), A.mlir_data, size(A))
-Base.collect(A::TracedRArray) = copy(A)
 
 # TODO is there a way to create an unitialized `tensor`? does it show an advantage? maybe `fill`?
 function Base.similar(::TracedRArray, ::Type{T}, dims::Dims{N}) where {T,N}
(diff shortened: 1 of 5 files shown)


src/TracedRArray.jl
```diff
@@ -160,6 +160,7 @@ Base.Tuple(x::TracedRArray) = ntuple(Base.Fix1(Base.getindex, x), length(x))
 Base.size(x::TracedRArray) = x.shape
 
 Base.copy(A::TracedRArray{T,N}) where {T,N} = TracedRArray{T,N}((), A.mlir_data, size(A))
+Base.collect(A::TracedRArray) = copy(A)
```
wsmoses, line 163: Should this reshape it tho? Specifically to a 1dim thing

mofeing: > Should this reshape it tho? Specifically to a 1dim thing

nope, the semantics of Julia is that `collect` of an `Array` just returns a copy. for returning a vectorized version of the array, there is `vec`

```julia
julia> x = rand(2,2)
2×2 Matrix{Float64}:
 0.518931  0.431565
 0.695147  0.4341

julia> collect(x)
2×2 Matrix{Float64}:
 0.518931  0.431565
 0.695147  0.4341

julia> objectid(x)
0xa9678941a4ed8e0c

julia> objectid(collect(x))
0x4ff9bea27617bd1b

julia> vec(x)
4-element Vector{Float64}:
 0.518930826791541
 0.6951470770259726
 0.4315651897756472
 0.43409960060102326
```

wsmoses (approved):
lgtm but add some tests?

PR #306, opened 2024-11-26
fix `YaoBlocks.mat` for rotation gates with `TracedRNumber` parameters

mofeing:
we run into some problems when trying to use a quantum circuit from Yao but whose parameters are traced by Reactant. specially, the LuxurySparse (which Yao uses a lot) arrays are quite problematic when used 

in the future, it might be better to use MLIR's sparse tensor types, but for now this is more than ok.

@wsmoses i know we discussed about creating this package extension here or in YaoBlocks.jl, but right now i'm on a rush and this is faster to get merged. also, in the future some changes in Reactant might fix the problems we had with the sparse arrays and thus, it might make sense to have it here for fast iteration. i'm pro moving it to YaoBlocks (if @Rogger-Luo is ok with it) once Reactant stabilizes.

CC @Todorbsc

### DIFF
--- a/ext/ReactantYaoBlocksExt.jl
+++ b/ext/ReactantYaoBlocksExt.jl
@@ -1,0 +1,34 @@
+module ReactantYaoBlocksExt
+
+function YaoBlocks.mat(
+    ::Type{T}, R::RotationGate{D,Reactant.TracedRNumber{S},<:XGate}
+) where {D,T,S}
+    M = Reactant.broadcast_to_size(zero(T), (2, 2))
+    M[1, 1] = cos(R.theta / 2)
+    M[2, 2] = cos(R.theta / 2)
+    M[1, 2] = -im * sin(R.theta / 2)
+    M[2, 1] = -im * sin(R.theta / 2)
+    return M
+end
+
+function YaoBlocks.mat(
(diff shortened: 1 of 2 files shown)


ext/ReactantYaoBlocksExt.jl
```diff
@@ -0,0 +1,34 @@
+module ReactantYaoBlocksExt
+
+function YaoBlocks.mat(
+    ::Type{T}, R::RotationGate{D,Reactant.TracedRNumber{S},<:XGate}
+) where {D,T,S}
+    M = Reactant.broadcast_to_size(zero(T), (2, 2))
```
wsmoses, line 6: So we know when these can be removed, can you open an issue for what happens when these aren’t here?

mofeing: okay, mentioned in #308

ext/ReactantYaoBlocksExt.jl
```diff
@@ -0,0 +1,34 @@
+module ReactantYaoBlocksExt
+
```
mofeing, line 2:
```suggestion

using Reactant
using YaoBlocks

```

PR #307, opened 2024-11-26
use row major when building attributes

Pangoraw:


mofeing:
so discussing with @wsmoses about this in today's meeting, we have reached the conclusion that the column- vs row-major layout is not forced by MLIR but it's dialect specific (e.g. a Julia MLIR dialect could still use column-major layout).

furthermore, I just found that column-major layouts can be represented by affine maps https://mlir.llvm.org/docs/Dialects/Builtin/#affine-map-layout do we know if affine maps can be used in tensors? they only examples I found are with memrefs. also, is StableHLO compatible with them?

I still believe that this PR is useful, but I would suggest to make the conversion from column- to row-major optional. maybe a kwarg that defaults to `false`?

test/basic.jl
```diff
@@ -593,3 +593,18 @@ end
...
+function f_row_major(x)
+    y = [1 2; 3 4]
+    if x isa Reactant.TracedRArray
+        y = Reactant.promote_to(Reactant.TracedRArray{eltype(x),2}, y)
+    end
+    return x .+ y
+end
+
+@testset "array attributes: row major" begin
+    x = rand(Int, 2, 2)
```
wsmoses, line 606: Minor but can we have this be a fixed constant so it’s easier to debug when it goes wrong

Pangoraw:
> we have reached the conclusion that the column- vs row-major layout is not forced by MLIR but it's dialect specific

I agree but here this is for building shaped builtin attributes through the C API. Row major is needed for the data and shape to be interpreted accordingly. A dialect is then free to reinterpret the attribute as suited within its ops.

> but I would suggest to make the conversion from column- to row-major optional. maybe a kwarg that defaults to false?

I don't really see why one would not want to convert to row major here. We can add it later when/if a potential julia dialect needs transposed attributes.

mofeing:
> I agree but here this is for building shaped builtin attributes through the C API. Row major is needed for the data and shape to be interpreted accordingly. A dialect is then free to reinterpret the attribute as suited within its ops.

i see your point but I'm not sure of all the consequence. how about we discuss it the next meeting?

wsmoses:
Yeah okay I can agree with that logic (since then the bultin attribute conversion is nice — which is distinct from the builtin op semantics).

im okay with this PR

src/mlir/IR/Attribute.jl
```diff
@@ -492,6 +492,9 @@ function Base.fill(::Core.Type{Attribute}, value, shape)
     return Base.fill(value, shaped_type)
 end
 
+to_row_major(x) = permutedims(x, ndims(x):-1:1)
+to_row_major(x::AbstractVector) = x
```
wsmoses, line 496: From the error logs it looks like this also needs a 0-dim specialization

mofeing:
> Yeah okay I can agree with that logic (since then the bultin attribute conversion is nice — which is distinct from the builtin op semantics).
> 
> im okay with this PR

do you know if affine maps can be associated to dense array / dense elements attributes?

wsmoses:
Not attributes directly, but as part of a type

mofeing:
okay, and can be associated to a `tensor`? the only examples I've seen are with `memref`s

mofeing:
> (since then the bultin attribute conversion is nice — which is distinct from the builtin op semantics)

yeah, but we pass the values to `stablehlo.constant` as an attribute. the only change I would like is to make this conversion optional (I don't mind if the default is true or false) so that we can choose if to make the permutedims on the Julia side or on MLIR side.

wsmoses:
I don’t think tensors do, and also that would still be an issue here since it having a different type means that you couldn’t use a memref<affine map> where you needed a regular affine map.

still cool to use, but I think this is necessary regardless

Pangoraw:
Note that the fix here not only affects `stablehlo.constant` but also other ops which have dense attributes such as here:

https://github.com/EnzymeAD/Reactant.jl/blob/45ae14fee01730c3c509b2cb073b0158ea2277bb/ext/ReactantNNlibExt.jl#L111-L113

If we want to try not having the transpose in julia, it should be possible by doing it at the promotion level:

```julia
function promote_to(x)
    cst = stablehlo.constant(reshape(x, :))
    cst = stablehlo.reshape(cst, reverse(size(x)))
    cst = stablehlo.transpose(cst, ndims(x):-1:1)
    return cst
end
```

wsmoses:
> Note that the fix here not only affects `stablehlo.constant` but also other ops which have dense attributes such as here:
> 
> https://github.com/EnzymeAD/Reactant.jl/blob/45ae14fee01730c3c509b2cb073b0158ea2277bb/ext/ReactantNNlibExt.jl#L111-L113
> 
> If we want to try not having the transpose in julia, it should be possible by doing it at the promotion level:
> 
> ```julia
> function promote_to(x)
>     cst = stablehlo.constant(reshape(x, :))
>     cst = stablehlo.reshape(cst, reverse(size(x)))
>     cst = stablehlo.transpose(cst, ndims(x):-1:1)
>     return cst
> end
> ```

Yeah but frankly that’s the more intuitive thing anyways

Pangoraw:
Test errors seem caused by asymmetric padding in convolution, I will investigate

ext/ReactantNNlibExt.jl
```diff
@@ -109,7 +109,7 @@ function NNlib.conv!(
     #! format: on
 
     padding = Reactant.MLIR.IR.DenseElementsAttribute(
-        reshape(collect(padding), (num_spatial_dims, 2))
+        reshape(collect(padding), (2, num_spatial_dims))'
```
mofeing, line 112: don't call `'`/`adjoint` because it will conjugate complex matrices
```suggestion
        transpose(reshape(collect(padding), (2, num_spatial_dims)))
```

ext/ReactantNNlibExt.jl
```diff
@@ -163,7 +163,7 @@ function reduce_window(f, x::AnyTracedRArray{T,N}, pdims; init) where {T,N}
     end
 
     padding = Reactant.MLIR.IR.DenseElementsAttribute(
-        reshape([padding..., 0, 0, 0, 0], (N, 2))
+        reshape([padding..., 0, 0, 0, 0], (2, N))'
```
mofeing, line 166:
```suggestion
        transpose(reshape([padding..., 0, 0, 0, 0], (2, N)))
```

mofeing (changes requested):
okay, then let's merge it like this and revisit in the future if we need it

mofeing (approved):
ahh, `NNlib.padding` just returns a tuple of ints... so it's alright if you call `adjoint`

PR #310, opened 2024-11-27
respect scopping rules in for

Pangoraw:
Alternative to #303.

### DIFF
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -158,8 +158,15 @@ function trace_for(mod, expr)
         external_syms...,
     )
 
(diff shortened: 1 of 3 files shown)


lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -156,8 +156,15 @@ function trace_for(mod, expr)
         :tuple, :(Reactant.promote_to(Reactant.TracedRNumber{Int}, 0)), external_syms...
     )
 
+    cond_val(s) = :(isdefined($(mod), $(QuoteNode(s))) ? $s : nothing)
```
Pangoraw, line 159:
```suggestion
    cond_val(s) = :(@isdefined($s) ? $s : nothing)
```

avik-pal:
Is there anything that is needed before we can get this in?

avik-pal (approved):

PR #313, opened 2024-11-28
Speed up invoke usage

wsmoses:


mofeing:
should we really see a speedup with this changes?

IMHO we should maybe push for more `@nospecialize` in `traced_type`, `make_tracer` and `create_result`. also, we could precompile common methods.

the main issue for this right now is that we have no benchmarks for measuring each component separately, and benchmarking CI has a lot of noise to distinguish if some optimizations have effect or not 🥲

wsmoses:
Not necessarily immediately, but invoke without an arg causes it to re-specialize for each type, which is really bad in this context (and is clearly a bug anyways)

PR #317, opened 2024-12-01
add optional debug info to Ops calls

mofeing:
@wsmoses i made the debug info optional because calling `stacktrace` has a big overhead. for example, `stacktrace` takes around 800-900 $\mu s$ in my laptop, while creating a `Location` takes around 3 $n s$

in order to add the debug info, i've added the `DEBUG_MODE` global var and the `with_debug` method to manage it

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -13,10 +13,7 @@ end
 
 # constant ops
 function constant(
-    x::DenseArray{T,N};
-    location=MLIR.IR.Location(
-        "stablehlo.constant", MLIR.IR.Location(@__FILE__, @__LINE__, 0)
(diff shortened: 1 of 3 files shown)


wsmoses:
Did the passes or jll change.

That probably indicates you’re running an invalid pass

Pangoraw:
> error: 'transform.apply_patterns' op expected children ops to implement PatternDescriptorOpInterface

I updated the transforms in https://github.com/EnzymeAD/Reactant.jl/commit/9e04ba658a25640c870b635ebf4458b6011209d1 but did not update Reactant_jll compat so you might need to update Reactant_jll?

PR #321, opened 2024-12-03
Temporarily disable custom forward/backward diff rules in Enzyme-based interpreter

mofeing:
Let's see if this circumvents the compilation time issue while @vchuravy and @wsmoses fix the issue in Enzyme

### DIFF
--- a/src/Interpreter.jl
+++ b/src/Interpreter.jl
@@ -118,8 +118,8 @@ function set_reactant_abi end
             ReactantCacheToken(),
             nothing,            #=mt=#
             world,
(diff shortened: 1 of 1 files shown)


mofeing:
It seems like this doesn't circumvent the issue 🥲

wsmoses:
sad

wsmoses:
x/ref https://github.com/EnzymeAD/Enzyme.jl/issues/2152 if that is the same place things are beign slow [and we can borrow the same things I was using to debug perf here]

mofeing:
fixed in #341

Issue #311, opened 2024-11-27
Tracking issue for missing tests


mofeing:
- [ ] Ops
  - [ ] after_all
  - [ ] cbrt (depends on #291, spec might misspecified for complex numbers)
  - [ ] constant (depends on #196)
  - cosine on Apple x86_64 (it's a bug on a deprecated platform)
  - [ ] einsum (needs more cases)
  - [ ] iota
  - [ ] logistic
  - [ ] partition_id (depends on #196)
  - [ ] outfeed
  - [ ] recv
  - [ ] replica_id (depends on #196)
  - [ ] rng_bit_generator
  - [ ] send
  - [ ] set_dimension_size
  - [ ] zeta

PR #323, opened 2024-12-04
fix: #196

### LINKED ISSUE #196: crash with parameter less function
```julia
julia> f = Reactant.compile(()) do
           a = ones(Float64, 2, 2)
           b = Reactant.promote_to(Reactant.TracedRArray{Float64,2}, a)
           return b
       end;
       f()

[431649] signal 11 (1): Segmentation fault
in expression starting at REPL[2]:1
llvm_type_rewrite at /cache/build/builder-demeter6-6/julialang/julia-master/src/ccall.cpp:402
emit_llvmcall at /cache/build/builder-demeter6-6/julialang/julia-master/src/ccall.cpp:846
emit_intrinsic at /cache/build/builder-demeter6-6/julialang/julia-master/src/intrinsics.cpp:1265
emit_call at /cache/build/builder-demeter6-6/julialang/julia-master/src/codegen.cpp:5201
emit_expr at /cache/build/builder-demeter6-6/julialang/julia-master/src/codegen.cpp:6199
emit_ssaval_assign at /cache/build/builder-demeter6-6/julialang/julia-master/src/codegen.cpp:5745
emit_stmtpos at /cache/build/builder-demeter6-6/julialang/julia-master/src/codegen.cpp:6023 [inlined]
emit_function at /cache/build/builder-demeter6-6/julialang/julia-master/src/codegen.cpp:9207
jl_emit_code at /cache/build/builder-demeter6-6/julialang/julia-master/src/codegen.cpp:9541

```

glou-nes:
@mofeing, unblocks several tests in #311
Fix #196. 
`Base.llvmcall`  doesn't handle type `NTuple{N,Ptr{Cvoid}}` for N=0. So the Julia LLVM codegen cannot process parameter less call like `@jit Ops.constant([1])`.
Adapt the generated function `ExecutableCall` and LLVM glue code to deal with this case.

### DIFF
--- a/src/XLA.jl
+++ b/src/XLA.jl
@@ -285,14 +285,19 @@ end
 function execute_ir(N, n_outs, fn)
     ptr = sizeof(Int) == sizeof(Int64) ? "i64" : "i32"
     cint = sizeof(Cint) == sizeof(Int64) ? "i64" : "i32"
-    res = """define { [$n_outs x $ptr], [$n_outs x $ptr], i8 } @f($ptr %exec, [$N x $ptr] %inps, [$N x i8] %donated) alwaysinline {
+    args = N > 0 ? ", [$N x $ptr] %inps, [$N x i8] %donated" : ""
+    stores = N > 0 ? """
+store [$N x $ptr] %inps, [$N x $ptr]* %inpa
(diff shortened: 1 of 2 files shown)


mofeing:
awesome!

PR #337, opened 2024-12-07
test: unbreak CUDA CI

avik-pal:


src/Ops.jl
```diff
@@ -485,13 +487,20 @@ function cholesky(
...
         stablehlo.cholesky(
-            x.mlir_data; result=mlir_type(TracedRArray{T,N}, size(x)), lower, location
+            x.mlir_data;
+            result=mlir_type(TracedRArray{T,N}, size(x)),
+            lower=MLIR.IR.Attribute(lower),
+            location,
         ),
     )
-    return TracedRArray{T,N}((), res, size(x))
+    res = TracedRArray{T,N}((), res, size(x))
```
wsmoses, line 498: Oh actually sorry I didn’t mean here ops really should try to be as semantically close to hlo as it can be. I meant more for a Julia cholesky override proper.

we should just ignore the implementation defined part

wsmoses: For the test**

avik-pal: makes sense

src/ConcreteRArray.jl
```diff
@@ -289,15 +289,23 @@ end
...
 function Base.copy(bc::Base.Broadcast.Broadcasted{Broadcast.ArrayStyle{ConcreteRArray}})
-    ElType = Base.Broadcast.combine_eltypes(bc.f, bc.args)
-    if !Base.isconcretetype(ElType)
-        throw(
-            ErrorException(
-                "`copy` on `ConcreteRArray` for non-concrete eltype is not implemented"
-            ),
-        )
-    end
+    fn = Reactant.compile(Broadcast.BroadcastFunction(bc.f), (bc.args...,))
```
wsmoses, line 292: Can we special case the previously handled case? 

I feel like we should use JIT sparingly on the concrete side to avoid people accidentally finding performance cliffs

avik-pal: done. Compiling all non-CPU cases

src/ConcreteRArray.jl
```diff
@@ -289,15 +289,44 @@ end
...
 # TODO replace this copy for `setindex!` maybe? how to copy data to already existing buffer? (i.e. `copyto!`)
 function Base.copy(bc::Base.Broadcast.Broadcasted{Broadcast.ArrayStyle{ConcreteRArray}})
-    ElType = Base.Broadcast.combine_eltypes(bc.f, bc.args)
-    if !Base.isconcretetype(ElType)
-        throw(
-            ErrorException(
-                "`copy` on `ConcreteRArray` for non-concrete eltype is not implemented"
-            ),
-        )
+    foreach(bc.args) do x
```
wsmoses, line 292: Can we do this with loops rather than creating an anonymous function for future compile times to be happier?

src/ConcreteRArray.jl
```diff
@@ -289,15 +289,44 @@ end
...
+end
+
+function Base.mapreduce(
+    @nospecialize(f),
+    @nospecialize(op),
+    @nospecialize(A::ConcreteRArray{T,N});
+    dims=:,
+    init=nothing,
+) where {T,N}
+    mapreduce_fn(x) = Base.mapreduce(f, op, x; dims, init)
```
wsmoses, line 329: also here can we make a specific

```
struct CallMapReduce{Fn, Op, Dims, Init}
    ...
end
```

and @jit that?

again here for the sake of reducing compile times (but this time the reason is so it'll be easier to cache and not recompile the same method in the long term)

test/ops.jl
```diff
@@ -82,14 +82,16 @@ end
 end
 
 @testset "cholesky" begin
-    g(x) = Ops.cholesky(x; lower=true)
+    g1(x) = triu(Ops.cholesky(x))
```
wsmoses, line 85: add a comment describing why we do this for ease in the future, maybe link to issue?

avik-pal:
@wsmoses any other change you think is needed here?

PR #341, opened 2024-12-07
Fix for interpreter

wsmoses:


mofeing:
So it seems like CI in Ubuntu Julia x86_64 has gotten down from ~55min to ~15min!

there still can be some overhead still because previously took ~5-7min (but many PRs with more tests have been merged in between, specially the Ops PR which added a lot of tests)

PR #347, opened 2024-12-09
refactor: use Ops instead of direct stablehlo calls

avik-pal:


src/TracedRArray.jl
```diff
@@ -36,35 +36,16 @@ ReactantCore.is_traced(::TracedRArray) = true
...
 
-function TracedRArray{T,N}(rhs::AbstractArray{T0,N}) where {T0,T,N}
-    attr = MLIR.IR.DenseElementsAttribute(collect(rhs))
-    return TracedRArray{T,N}(
-        TracedRArray{T0,length(size(rhs))}(
-            (), MLIR.IR.result(MLIR.Dialects.stablehlo.constant(; value=attr), 1), size(rhs)
-        ),
-    )
-end
+TracedRArray{T,N}(rhs::AbstractArray{T0,N}) where {T0,T,N} = Ops.constant(collect(rhs))
```
mofeing, line 48: maybe can we implement `Base.convert` here too?

src/TracedRArray.jl
```diff
@@ -266,69 +244,20 @@ function Base.show(io::IOty, X::TracedRArray{T,N}) where {T,N,IOty<:Union{IO,IOC
...
-
-Base.conj!(A::TracedRArray) = A
-function Base.conj!(A::TracedRArray{T,N}) where {T<:Complex,N}
-    A.mlir_data = MLIR.IR.result(
-        MLIR.Dialects.chlo.conj(A.mlir_data; result=mlir_type(TracedRArray{T,N}, size(A))),
-        1,
-    )
+Base.conj!(A::AnyTracedRArray) = A
+function Base.conj!(A::AnyTracedRArray{<:Complex})
+    set_mlir_data!(A, Ops.conj(materialize_traced_array(A)).mlir_data)
```
mofeing, line 252: Please, don't remove the implementations for `Base.conj`. Ops are inheritly "out-place", so just calling them work straightforward.

mofeing (changes requested):

src/TracedRArray.jl
avik-pal: yeah, done!

mofeing (approved):

src/utils.jl
```diff
@@ -27,10 +27,8 @@ function transpose_ty(mlirty)
     return MLIR.IR.TensorType([reverse(size(mlirty))...], eltype(mlirty))
 end
 function transpose_val(val)
```
wsmoses, line 29: okay this one is more of a utility so I'm honestly fine leaving this at a lower level without constructing a tracedrarray since it takes an mlir.value input and does an mlir.value output

wsmoses (approved):

PR #351, opened 2024-12-09
add support to complex expr in compile_call_expr

### LINKED ISSUE #333: Broadcasting Operations: invalid MLIR generated
I don't know if that kind of invalid generation is interesting to fix.
```
using Reactant
Reactant.set_default_backend("cpu")
x = Reactant.ConcreteRArray(randn(Float64, 2, 2))
@code_hlo x .- x .+ x
```

```
module {
  func.func @main(%arg0: tensor<2x2xf64>, %arg1: tensor<2x2xf64>) -> tensor<2x2xf64> {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<2x2xf64>) -> tensor<2x2xf64>
    %1 = stablehlo.transpose %arg1, dims = [1, 0] : (tensor<2x2xf64>) -> tensor<2x2xf64>
    %2 = stablehlo.add %0, %1 : tensor<2x2xf64>
    %3 = stablehlo.transpose %2, dims = [1, 0] : (tensor<2x2xf64>) -> tensor<2x2xf64>
    return %3 : tensor<2x2xf64>
  }
}
```

glou-nes:
fix #333.
Help reducing the discrepancy between @code_typed (and friends) and @code_hlo.
Introduce a closure for complex expression in `Compiler.compile_call_expr`. Avoid invalid MLIR generation for expression like `@jit f(f(a))`. The closure is create only for these complex case.


src/Compiler.jl
```diff
@@ -432,6 +432,38 @@ macro jit(args...)
...
+    args = Set{Symbol}() #collect all the closures argument by walking expression
+    walk = (e::Expr) -> begin
+        Meta.isexpr(e, :.) && return nothing #ignore expression like Main.C
+        i = Meta.isexpr(e, :call) ? 2 : 1
+        for arg in e.args[i:end]
+            arg isa Expr && walk(arg)
+            arg isa Symbol && push!(args, arg)
+        end
+    end
+    walk(expr)
```
Pangoraw, line 446: You can use [ExpressionExplorer.jl](https://github.com/juliapluto/expressionexplorer.jl) here, similarly to the transforms in ReactantCore which creates closures as well.

Pangoraw: IMO this is not a problem to have modules as a param of the closure:

```julia
closure = (f, X, Y, Base) -> f(f(X,Y), Base.pi)
closure(f, X, Y, Base)

File: src/Compiler.jl [hunk 2]
```diff
@@ -432,6 +434,35 @@ macro jit(args...)
...
+    args = ExpressionExplorer.compute_symbols_state(expr).references
+    args = filter(!is_a_module, args)
+    args = tuple(collect(args)...)
+    n_args = length(args)
+    closure_args = if n_args == 1
+        only(args)
+    else
+        Expr(:tuple, args...)
+    end
+    return (Expr(:tuple, args...), Expr(:->, closure_args, Expr(:block, expr)))
```
wsmoses, line 453: This makes an anonymous function, can we make a struct of our own with a callable attr?

glou-nes: I suppose closure are not cache friendly? If I understand correctly, you talk about something like that:
```
struct P P() = $expr end
```

glou-nes: That a better approach anyway because this doesn't capture any environment.

Pangoraw: Structs must be defined at top level so I think that would mean that the macros cannot be used from within functions (to be tested). IIUC, closures just create a single type and instantiate it as a callable struct would.

src/Compiler.jl
Pangoraw: Compile functions should support any arguments, not just reactantrarray.

Pangoraw: ah right. We can use `isdefined(__module__, s) && getproperty(__module__, s) isa Module` in `ìs_a_module` instead of checking the value within `Main` so that it works even if the module is not defined in main

avik-pal:
After this change will `@jit foo(Reactant.to_rarray(rand(2)))` still work or is the expection that this needs to be written as:

```julia
x = Reactant.to_rarray(rand(2))
@jit foo(x)
```

glou-nes:
@avik-pal I forget about this pattern, I will add a case to fallback for this case.

glou-nes:
@avik-pal, I restored the normal behavior and it seems that the CI is right.

glou-nes:
A missing part is kw, they seem to be unsupported now:
```
Ops.pad(x, v; low=[1])

quote
    begin
        var"##f#492373" = Ops.pad
        var"##args#492374" = (x, v; low = [1])
        var"##compiled#492375" = (Reactant.Compiler.compile)(var"##f#492373", var"##args#492374"; sync = false, optimize = true)
    end
    var"##compiled#492375"(var"##args#492374"...)
end

```

wsmoses (approved):

Issue #354, opened 2024-12-09
Trace over Python


mofeing:
There are times where I need to use some Python package but Reactant is unable to trace over it. So I need to find a way to translate it to Julia before and then trace over it.
For example, for my VQE simulation, the circuit is defined in Qiskit so I need to translate the circuit to Yao and then trace over Yao.

Problem is that this translation is not always posible because...

1. there might be libraries in Python which do not have an equivalent in Julia
2. even if there exists an equivalent, the translation might not be perfect
3. coding the alternative yourself can be time and energy consuming

This could be solved if we add a way to trace over PythonCall. I think we can do this by implementing the `ConcreteRArray` and `TracedRArray` types as Python classes, such that they register any operation being done onto them by emitting the MLIR like we do (unlike JAX, which constructs a computational graph first). This MLIR emission can be performed by calling back to Reactant in Julia. And the interception of calls can be done with NumPy's dispatch mechanism.

It doesn't need to be perfect, but just having something would alleviate the interaction between Python and Reactant.

Pd.: My intention is not to reimplement JAX in Python by calling Reactant jajaja. I don't want to start a war against them, but just to solve some integration problems.

wsmoses:
Can we just have trace of pythoncall call Jax and import it with an hlo_call?

mofeing:
> Can we just have trace of pythoncall call Jax and import it with an hlo_call?

I'm not very familiar with `hlo_call` yet, but it could be a possibility. The only problem I see is that users have to set up their Python code to call Jax right? Or can we get in charge of setting up the Jax part?

wsmoses:
In principle we can automate this if the python code is itself jax compatible

mofeing:
So in theory, idiomatic usage of NumPy should be sufficient to make it work with Jax. In practice, I fear that some things like indexing can give us problems as stated in https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#in-place-updates

avik-pal:
xref https://discourse.julialang.org/t/calling-python-function-jit-compiled-with-jax-from-julia-without-overhead/123552

wsmoses:
so in principle once https://github.com/EnzymeAD/Reactant.jl/pull/364 lands, we should be able to overload all uses of `PythonCall.pycall` to do whatever we want. In this case, we want to basically do what @sefffal starting discussing on github with:

```julia
using Reactant
using Reactant: Ops
using Enzyme
[...]

PR #358, opened 2024-12-10
Add Ops.hlo_call(::String, args...)

Pangoraw:


Pangoraw:
idea: we could hash the code and just the call the function if a function with the same hash is already present in the module (preventing duplicate functions).

So that:
```julia
for i in 1:N
  x = Ops.hlo_call(code, x)
end
```
has the function just once (but multiple calls).

mofeing:
that can be done directly with a `Dict{HLOIR,Function}` (or sth like that) and then calling ```julia
get!(dict, code) do code
    compile(code)
end
```

**Comment by Pangoraw** [comment]
> that can be done directly with a Dict{HLOIR,Function} (or sth like that) and then calling

ah right, do we have a place to store such state within a single trace ? Otherwise my idea is to name these function like `main_$(hash(code))` and see if it exists within the module before adding it.

**Comment by mofeing** [comment]
I have a question... What is `hlo_call` returning? A function? Wouldn't it be better to just inject the MLIR code in the current MLIR block?

**Comment by Pangoraw** [comment]
No, it is returning a tuple of tracedarray returned by the call of the function.
_i.e._ `Reactant.@jit Ops.hlo_call(code, x, y)` is a tuple of concrete arrays.

File: src/Ops.jl [hunk 1]
```diff
@@ -1046,4 +1046,72 @@ function compare(
...
+              Reactant.to_rarray(Float32[1, 2, 3]),
+          )
+       )
+(ConcreteRArray{Float32, 1}(Float32[2.0, 4.0, 6.0]),)
+```
+"""
+function hlo_call(code, args...; location=mlir_stacktrace("hlo_call", @__FILE__, @__LINE__))
+    new_mod = parse(MLIR.IR.Module, code)
+    body = MLIR.IR.body(new_mod)
+    fn = MLIR.IR.first_op(body)
```
mofeing, line 1076: what if the first op is not `main`? like what if we the code was traced by us and we added some function barriers.

maybe we can add a kwarg for selecting the target function (and default it to `main`), so we just iterate over the ops doing `first(Iterators.filter(op -> String(IR.attr(op, "sym_name")) == target_fn, OperationIterator(body))`

src/Ops.jl
```diff
@@ -1046,4 +1046,72 @@ function compare(
...
+
+    push!(top_level_block, fn)
+
+    ftype_attr = MLIR.IR.attr(fn, "function_type")
+    ftype = MLIR.IR.Type(ftype_attr)
+
+    MLIR.IR.attr!(fn, "sym_name", MLIR.IR.Attribute(name))
+
+    @assert all(Base.Fix2(isa, Reactant.AnyTracedRArray), args) "all inputs to hlo_call should be reactant arrays"
+    @assert MLIR.IR.ninputs(ftype) == length(args) "invalid number of argument for function $orig_name"
```
mofeing, line 1093:
```suggestion
    @assert MLIR.IR.ninputs(ftype) == length(args) "invalid number of arguments for function $orig_name"
```

test/ops.jl
```diff
@@ -866,3 +866,25 @@ end
...
+    @test Reactant.@jit(
+        Ops.hlo_call(
+            """
+module {
+  func.func @main(%arg0: tensor<3xf32>, %arg1: tensor<3xf32>) -> tensor<3xf32> {
+    %0 = stablehlo.add %arg0, %arg1 : tensor<3xf32>
+    return %0 : tensor<3xf32>
+  }
+}
+""",
```
mofeing, line 885: when using multi-line strings, you can put an indentation without problem. indentation will be applied to the final string relative to the first/last `"""` and the left-most line.

```suggestion
            module {
              func.func @main(%arg0: tensor<3xf32>, %arg1: tensor<3xf32>) -> tensor<3xf32> {
                %0 = stablehlo.add %arg0, %arg1 : tensor<3xf32>
                return %0 : tensor<3xf32>
              }
            }
            """,
```

src/Ops.jl
```diff
@@ -1046,4 +1046,72 @@ function compare(
...
+          )
+       )
+(ConcreteRArray{Float32, 1}(Float32[2.0, 4.0, 6.0]),)
+```
+"""
+function hlo_call(code, args...; location=mlir_stacktrace("hlo_call", @__FILE__, @__LINE__))
+    new_mod = parse(MLIR.IR.Module, code)
+    body = MLIR.IR.body(new_mod)
+    fn = MLIR.IR.first_op(body)
+    MLIR.IR.rmfromparent!(fn)
```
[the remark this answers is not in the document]
Pangoraw: No more in the opposite case
 1. We parse a new module `new_mod`
 2. Extract the func from this mode (`rmfromparent!`)
 3. Add it to the currently traced module `push!(top_level, ...)`

src/Ops.jl
Pangoraw: This is inside the code that was given by the caller. Currently, the expectation is that there is only one function inside the given module.

test/ops.jl
Pangoraw: cool, thanks!

src/Ops.jl
wsmoses: yeah I would instead ideally have a kwargument fn=main, and we extract that fn as the top level one

src/Ops.jl
```diff
@@ -1046,4 +1046,72 @@ function compare(
...
+
+    current_module = MLIR.IR.mmodule()
+    top_level_block = MLIR.IR.body(current_module)
```
wsmoses, line 1080: we also need to mark all fn's as private, as well as make sure to move all fns in the module (e.g. the main function could call something)

src/Ops.jl
```diff
@@ -1046,4 +1046,72 @@ function compare(
...
+
+    orig_name = String(MLIR.IR.attr(fn, "sym_name"))
+    name = orig_name * "_" * string(gensym())
+
+    push!(top_level_block, fn)
+
+    ftype_attr = MLIR.IR.attr(fn, "function_type")
+    ftype = MLIR.IR.Type(ftype_attr)
+
```
wsmoses, line 1089: we should check that type types of the args to hlo_call are the same as the input types of the main function

src/Ops.jl
wsmoses: we have some utilities here: https://github.com/EnzymeAD/Enzyme-JAX/blob/f6587e37ff7298f2a1a273b08c24d69fca7ff30f/src/enzyme_ad/jax/compile_with_xla.cc#L190 and https://github.com/EnzymeAD/Enzyme-JAX/blob/f6587e37ff7298f2a1a273b08c24d69fca7ff30f/src/enzyme_ad/jax/primitives.py#L811 in Enzyme-JaX for explicitly making we can do all the things

Pangoraw: Done, do you know if we can encounter ops other than `func.func` (maybe `gpu.func` in the future?) and what to do with them ?

src/Ops.jl
Pangoraw: added a `func_name::String` kwarg for this

src/Ops.jl
```diff
@@ -1046,4 +1046,120 @@ function compare(
...
+    symbol_attr_name = String(MLIR.API.mlirSymbolTableGetSymbolAttributeName())
+
+    fn = MLIR.IR.lookup(
+        MLIR.IR.SymbolTable(MLIR.IR.Operation(current_module)), name_to_call
+    )
+    if isnothing(fn)
+        new_mod = parse(MLIR.IR.Module, code)
+        body = MLIR.IR.body(new_mod)
+
+        for op in MLIR.IR.OperationIterator(body)
```
wsmoses, line 1099: Oh we should import all the ops, not just func, but it’s okay to be limited to just func as the entry function. Eg if main calls a gpu function that would be fine. Or if a global constant op

test/ops.jl
```diff
@@ -866,3 +866,57 @@ end
...
 end
+
+@testset "hlo_call" begin
+    x = Float32[1.0, 2.0, 50.0]
+    y = Float32[-4.0, 0.001, 2.0]
+    x_reactant = Reactant.to_rarray(x)
+    y_reactant = Reactant.to_rarray(y)
+
+    @test Reactant.@jit(
+        Ops.hlo_call(
```
wsmoses, line 877: Can you add a test with multiple functions in the module.

and can you also add a test with two (different) hlo calls that happen to contain functions of the same name (to make sure we do the symbol rename properly)

Pangoraw:
So hlo_call currently returns a tuple of arrays (one for each result), should we special case `nresults() == 1` to return the only array (no tuple) ?

test/ops.jl
```diff
@@ -866,3 +866,78 @@ end
...
+        )
+    )[1] ≈ x .+ y
+end
+
+function f_repeat(x, y)
+    for _ in 1:3
+        x, = Ops.hlo_call(
+            """
+            module {
+              func.func @my_add(%arg0: tensor<3xf32>, %arg1: tensor<3xf32>) -> tensor<3xf32> {
```
wsmoses, line 897: Can you also add a version of this where the two definitions are different.

just because if we fix caching then we might not actually not emit it twice (and thus not check things)

wsmoses:
> So hlo_call currently returns a tuple of arrays (one for each result), should we special case `nresults() == 1` to return the only array (no tuple) ?

Honestly I think it’s better to always return the tuple. That way folks using it don’t need to special case if there are multiple returns or not

test/ops.jl
Pangoraw: Added a test with the same name but different definitions:

https://github.com/EnzymeAD/Reactant.jl/blob/8eb71cc9f64ecad760765ee30a4c3bc27eba011e/test/ops.jl#L945-L970

wsmoses (approved):
Awesome stuff!

mofeing:
this is amazing and opens a way to do things like #354

PR #364, opened 2024-12-11
Interpreter fixes (prerequisite to CUDA support)

wsmoses:
and also should fix the nested gradient overlay, and ideally improve compile times (but not necessarily time to first @ compile ]


wsmoses:
Currently I want to ensure this doesn't break anything.

Then I will potentially revamp the Enzyme overload to use more of this (and already this should hopefully fix the use of the Enzyme overload in type unstable code), as well as CUDA and perhaps pythoncall stuff

wsmoses:
....and naturally I segfault julia. @aviatesk  @vtjnash @gbaraldi @vchuravy @topolarity do you have any thoughts here?

wsmoses:
Closing in favor of https://github.com/EnzymeAD/Reactant.jl/pull/365

PR #366, opened 2024-12-11
`@trace` function calls

jumerckx:
no caching, multi-return, kwargs, ... yet.


lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -3,6 +3,9 @@ module ReactantCore
 using ExpressionExplorer: ExpressionExplorer
 using MacroTools: MacroTools
 
+using Base.ScopedValues
```
mofeing, line 6: `ScopedValues` where introduced in Julia 1.11, so how about adding the package? i think Base reimplements it and doesn't just reexport, so maybe put an `@static if` based on Julia version?

src/ControlFlow.jl
```diff
@@ -130,6 +130,83 @@ function ReactantCore.traced_while(
...
+        Reactant.Compiler.callcache[][cache_key] = (; f_name, mlir_result_types, traced_result)
+    end
+
+    call_op = MLIR.Dialects.func.call(
+        linear_args;
+        result_0=mlir_result_types,
+        callee=MLIR.IR.FlatSymbolRefAttribute(f_name),
+    )
+
+    seen_results = Reactant.OrderedIdDict()
```
jumerckx, line 190: I first copy a traced structure using `make_tracer`, `TracedSetPath`. This structure can I then fill in by going by mutating the objects in `seen`. This seems to work well but maybe someone has a better approach.

src/Compiler.jl
```diff
@@ -287,12 +289,16 @@ function compile_mlir(f, args; kwargs...)
     end
 end
 
+const callcache = ScopedValue{Dict}()
```
jumerckx, line 292: This seems like a simple way to store the cache. Fancier solutions (i.e. integrating with absint?) are probably possible as well but not sure whether it's worth it to investigate as of now.

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -3,6 +3,9 @@ module ReactantCore
...
+const enable_tracing = ScopedValue{Bool}(false)
```
jumerckx, line 7: I think we simply want to trace function calls everytime in `make_mlir_fn`. I added this scopedvalue that is set in the beginning so it's just a simple check.
If it's okay to use ScopedValues in ReactantCore, ScopedValues probably needs to be added to project.toml for backwards compatibility (?)

lib/ReactantCore/src/ReactantCore.jl
jumerckx: You're a step ahead of me, i just added the same comment myself. Is it fine for ReactantCore to depend on ScopedValues?

jumerckx: I don't think a static if is needed as the ScopedValues package just reexports Base.ScopedValues

lib/ReactantCore/src/ReactantCore.jl
wsmoses: with the interpreter landed, we can have a function within_tracing similar to https://github.com/EnzymeAD/Enzyme.jl/blob/7c0823fa64426a745dae8fc7a50980be0a0a8dc8/lib/EnzymeCore/src/EnzymeCore.jl#L561 https://github.com/EnzymeAD/Enzyme.jl/blob/7c0823fa64426a745dae8fc7a50980be0a0a8dc8/src/compiler/interpreter.jl#L916

which would be nicer than a global

src/Compiler.jl
wsmoses: I think ironically we should do something slightly different here, and we can do something like we do with mmodule (to get the current mlir module): https://github.com/EnzymeAD/Reactant.jl/blob/73899f5b708b7ca87ffee72f597f7ea5696f93ac/src/mlir/IR/Module.jl#L81

If we define the cache scope during the outermost compile / module generation then we're certain that it doesn't get messed up and also can cache calls across subfunctions/etc nicely [which this does too to be clear, but maybe not quite as clean?]

src/ControlFlow.jl
```diff
@@ -130,6 +130,69 @@ function ReactantCore.traced_while(
     end
 end
 
+function ReactantCore.traced_call(f, args...)
+    seen_cache = Reactant.OrderedIdDict()
+    cache_key = (f, make_tracer(
+        seen_cache,
+        args,
+        (),
+        CallCache;
```
wsmoses, line 139: maybe we can call this mode TracedToTypes or something (since really its converting Traced to a Type expression)

wsmoses: This also isn't sufficient, since the cache key will only consider the outermost object as the key.

For example

```julia
mutable struct T
   x
end

X = T(1.0)
cache_key1 = make_key(X)
X.x = 2.0
cache_key2 = make_key(X)
```

Presumably the two keys will be the same since the hash key will be based on the object ID, not the potentially recursive structure.

I think we probably want to make a version of make_tracer that returns a tuple of all of the leaf objects, and we can obviously replace all tracedvalues/numbers with just the type.

src/Tracing.jl
```diff
@@ -382,6 +383,12 @@ function make_tracer(
     if mode == ConcreteToTraced
         throw("Cannot trace existing trace type")
     end
+    if mode == CallCache
+        if !haskey(seen, prev)
```
wsmoses, line 387: Edit, started writing this, but I realize the design has a different issue we need to solve first that point moot (https://github.com/EnzymeAD/Reactant.jl/pull/366/files#r1885172358)

So this is going to present an issue for some code.

For example

```
struct Wrap
   x::AbstractArray
end
```

We will try to set x to the MLIR.Type which is busted.

I think what we want to do here is make a fake abstractarray or even rarray type that solely contains the size and eltype.

src/ControlFlow.jl
jumerckx: Just the leaves is not enough I think. For example, `foo` and `bar` have the same type and the same leaves. But should be cached separately.
```
struct Foo
    x
end

struct Bar
    x
end

foo = Foo(Foo(1))
bar = Foo(Bar(1))
```

wsmoses: Maybe 

If leaf: the object
Otherwise: typeof(object), objects(children…)…

jumerckx: Right, that could work. I'll experiment a bit.

jumerckx: [`68559c5` (#366)](https://github.com/EnzymeAD/Reactant.jl/pull/366/commits/68559c576b239609b5524d96dc3299128eb4ff74) What do you think of this approach. I created a `Cached` type with custom equality and hashing to be used as dictionary keys.
The equality and hashing recursively visit all fields and check type and actual values.

src/Compiler.jl
jumerckx: Added here: [`588cebb` (#366)](https://github.com/EnzymeAD/Reactant.jl/pull/366/commits/588cebb7a9f6034dbb1b5f9c1220a4f37d639d56)

src/ControlFlow.jl
```diff
@@ -157,3 +224,44 @@ function get_region_removing_missing_values(compiled_fn, insertions)
     end
     return region
 end
+
+struct Cached
+    obj
+end
+Base.:(==)(a::Cached, b::Cached) = recursive_equal(a.obj, b.obj)
```
wsmoses, line 231: I'm a bit worried about this key in the context of mutation. For example

```julia
x = [1.0]
cache something with x

x[1] = 2.0
check cache (the key in the set is still based off of the x object, and now would hash to a different thing) 

```

jumerckx: Right, I hadn't accounted for that. So the problem is that the original key is corrupted when it is mutated. This can lead to too many functions being generated but shouldn't lead to incorrect code generation, I believe.
```julia
struct Foo
    x
end

a = rand(10)
b = deepcopy(a)
foo = Cached(Foo(a))

cache = Dict(Cached(Foo(a))=>"cache entry")
cache[Cached(Foo(a))] # "cache entry"
a[1] = 0.
cache[Cached(Foo(a))] # KeyError (as it should)
cache[Cached(Foo(b))] # KeyError!
```
I think it should suffice to add a deepcopy, i.e.:
`Base.:(==)(a::Cached, b::Cached) = recursive_equal(deepcopy(a.obj), deepcopy(b.obj))`?

src/ControlFlow.jl
jumerckx: Trying out extending `make_tracer` to generate a tuple or array containing types and leaf values.
How should we handle self-recursive objects, i.e. `a` contains `b` contains `a`.
The other trace modes return early when they encounter an object for the second time (by consulting `seen`).
But for generating a cache key, that behavior doesn't seem desired because we want the same cache key for:
```julia
a = [1, 2, 3]
b = [1, 2, 3]
# (a, a) should have same cache key as (a, b)
```

@wsmoses the example you gave against `deepcopy` was file handles I believe, does `deepcopy` simply not work for those or are their other issues going on there?

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -19,6 +19,13 @@ const SPECIAL_SYMBOLS = [
     :(:), :nothing, :missing, :Inf, :Inf16, :Inf32, :Inf64, :Base, :Core
 ]
 
+"""
+    within_tracing()
+
+Returns true if within tracing, otherwise false.
+"""
+@inline within_tracing() = false # behavior is overwritten in Interpreter.jl
```
wsmoses, line 27: I thought about this a bit more, and perhaps we want to name this within_compile for the future (especially if we rely on tracing less moving forward).

Thoughts?

jumerckx: Sounds good, will change!

src/ControlFlow.jl
wsmoses: deepcopies themselves are prone to failues (and generally are advised against in julialang/julia). Is the recursive tuple/array generation potentially still on the table?

Project.toml
```diff
@@ -17,6 +17,7 @@ Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 ReactantCore = "a3311ec8-5e00-46d5-b541-4f83e724a433"
 Reactant_jll = "0192cb87-2b54-54ad-80e0-3be72ad8a3c0"
+ScopedValues = "7e506255-f358-4e82-b7e4-beb19740aa63"
```
wsmoses, line 20: presumably we don't need scopedvalues anymore?

src/ControlFlow.jl
jumerckx: I have started implementing this but am not sure how to handle recursive datastructures. I had written about this om a previous comment chain here: https://github.com/EnzymeAD/Reactant.jl/pull/366/#discussion_r1893945631

wsmoses: ooh okay, what if the iddict mapped the structure to the size of the dict at time of insertion. Therefore we can represent the fact that it had a pointer to the 2nd object, or what not

jumerckx: I'm not sure I understand what you're proposing, could you clarify?

Thinking out loud with a more problematic example:
```julia
a.field = a

# vs.

a′.field = b
b.field = a′
```
Assuming `a`, `a′`,  and `b` have the same type: `a` and `a′` should point to the same cache entry while the structure in memory isn't identical.
One way I see to implement equality checks between two objects is to keep track of visited pairs:
```
(a, a′)
(a, b )
(a, a′) <-- already seen, done checking equality!
```
I'm still not completely convinced on the reservation against `deepcopy`, but maybe if I understand what you're proposing I'll see the light :)

wsmoses: So in this case I think it's fine if they cache to different things

So my idea is essentially as follows:

Create a map of object to ID (which is just the index of the time we inserted the object

For each object:
  If traced value, emit that value
  If a primitive type (int etc), emit that object
  Otherwise we are an object
     Check if we've emitted that object before, if so emit a new type Object(id) where id is the value in the map
     Otherwise
        Map[object] = map.size()
        Emit object(id)
     For each subfield
         Emit that field

wsmoses: Actually I think it should be emit object(pointer_from_objref(object)) not the map of

Essentially because you could have something like

If x===global


Inside the code and obviously they won't structurally be equivalent (tho perhaps it's best to have two modes one where it's a structural cache with the id as the map index and the second one where it is fine with globals)

jumerckx:
I did a rebase.
I also added the functionality to `make_tracer`, abusing the `path` argument to store the final vector of types/values-cache key.
I don't quite get the comment about
> If x===global

To me this would imply `x` and `global` *are* structurally equivalent, no?
So maybe I've missed something there. Depending on whether `seen` passed to `make_tracer` is an `IdDict` or `Dict`, caching behavior changes.

wsmoses (approved):
lgtm

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -21,6 +21,13 @@ const SPECIAL_SYMBOLS = [
     :(:), :nothing, :missing, :Inf, :Inf16, :Inf32, :Inf64, :Base, :Core
 ]
 
+"""
+    within_compile()
+
+Returns true if this function is executed in a Reactant compilation context, otherwise false.
+"""
+@inline within_compile() = false # behavior is overwritten in Interpreter.jl
```
giordano, line 29: https://github.com/EnzymeAD/Reactant.jl/actions/runs/13138061355/job/36658135000?pr=366#step:7:168
```
│ 
│     ReactantCore.within_compile :: Tuple{}
│ 
```
Need to add the new docstring to the documentation.

jumerckx: Aha thanks!

PR #369, opened 2024-12-12
feat: add support for the remaining wrapper types

### LINKED ISSUE #242: `getindex` assumes contiguous indexing

### LINKED ISSUE #345: Handling Wrapped Arrays Correctly
- [x] SubArray
- [x] Adjoint
- [x] Transpose
- [x] Diagonal
- [x] WrappedReshapedArray
- [x] PermutedDimsArray
- [x] LowerTriangular
- [x] UnitLowerTriangular
- [x] UpperTriangular
- [x] UnitUpperTriangular
- [x] Tridiagonal
- [x] Symmetric
- [ ] ~WrappedReinterpretArray~

For each of this we need 2 functions:

- `materialize_traced_array` - This unwraps all wrappers and generates a `TracedRArray`
- `set_mlir_data!` - Set the `mlir_data` field for the ancestor

avik-pal:
we need this before linear algebra stuff because most of them end up using these wrappers as return types.

fixes #345 #242 

## Added Wrapper Types

- Symmetric
- UnitLowerTriangular
- LowerTriangular
- UnitUpperTriangular
- UpperTriangular
- Tridiagonal

## Updated Linear Algebra Support

- `diagm` extended for key value pair version and now we use scatter

## New Addition to `Ops`

1. `scatter_setindex`
2. `gather_getindex`

Not calling them `scatter` / `gather` since they are a significantly restrictive subset of the corresponding stablehlo versions but these are also a very common pattern.

## Fixes / Refactor

1. Non-contiguous indexing for both setindex and getindex

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -1447,12 +1447,12 @@ instead.
     #! format: off
     scatter_dimension_numbers = MLIR.API.stablehloScatterDimensionNumbersGet(
         MLIR.IR.context(),
-        0, Int64[],
-        N, collect(Int64, 0:(N - 1)),
-        0, Int64[],
-        0, Int64[],
-        N, collect(Int64, 0:(N - 1)),
-        1
+        Int64(0), Int64[],
+        Int64(N), collect(Int64, 0:(N - 1)),
+        Int64(0), Int64[],
+        Int64(0), Int64[],
+        Int64(N), collect(Int64, 0:(N - 1)),
+        Int64(1)
     )
     #! format: on
 
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -1445,14 +1445,21 @@ instead.
     pushfirst!(update_computation, block)
 
     #! format: off
+    update_window_dims = Int64[]
+    inserted_window_dims = collect(Int64, 0:(N - 1))
+    input_batching_dims = Int64[]
+    scatter_indices_batching_dims = Int64[]
+    scatter_dims_to_operand_dims = collect(Int64, 0:(N - 1))
+    index_vector_dim = Int64(1)
+
     scatter_dimension_numbers = MLIR.API.stablehloScatterDimensionNumbersGet(
         MLIR.IR.context(),
-        Int64(0), Int64[],
-        Int64(N), collect(Int64, 0:(N - 1)),
-        Int64(0), Int64[],
-        Int64(0), Int64[],
-        Int64(N), collect(Int64, 0:(N - 1)),
-        Int64(1)
+        length(update_window_dims), update_window_dims,
+        length(inserted_window_dims), inserted_window_dims,
+        length(input_batching_dims), input_batching_dims,
+        length(scatter_indices_batching_dims), scatter_indices_batching_dims,
+        length(scatter_dims_to_operand_dims), scatter_dims_to_operand_dims,
+        index_vector_dim,
     )
     #! format: on
 
@@ -1486,20 +1493,26 @@ use [`MLIR.Dialects.stablehlo.dynamic_slice`](@ref) instead.
     @assert size(gather_indices, 2) == N
 
     #! format: off
+    offset_dims = Int64[1]
+    collapsed_slice_dims = collect(Int64, 0:(N - 2))
+    operand_batching_dims = Int64[]
+    start_indices_batching_dims = Int64[]
+    start_index_map = collect(Int64, 0:(N - 1))
+    index_vector_dim = Int64(1)
+
     dimension_numbers = MLIR.API.stablehloGatherDimensionNumbersGet(
         MLIR.IR.context(),
-        Int64(1), Int64[1],
-        Int64(N - 1), collect(Int64, 0:(N - 2)),
-        Int64(0), Int64[],
-        Int64(0), Int64[],
-        Int64(N), collect(Int64, 0:(N - 1)),
-        1
+        Int64(length(offset_dims)), offset_dims,
+        Int64(length(collapsed_slice_dims)), collapsed_slice_dims,
+        Int64(length(operand_batching_dims)), operand_batching_dims,
+        Int64(length(start_indices_batching_dims)), start_indices_batching_dims,
+        Int64(length(start_index_map)), start_index_map,
+        Int64(index_vector_dim),
     )
     #! format: on
 
     return reshape(
-        TracedRArray{T,2}(
-            (),
+        TracedRArray{T}(
             MLIR.IR.result(
                 MLIR.Dialects.stablehlo.gather(
                     src.mlir_data,
@@ -1510,7 +1523,6 @@ use [`MLIR.Dialects.stablehlo.dynamic_slice`](@ref) instead.
                 ),
                 1,
             ),
-            (size(gather_indices, 1), 1),
         ),
         size(gather_indices, 1),
     )
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -1473,4 +1473,47 @@ function scatter_setindex(
     )
 end
 
+"""
+    gather_getindex(src, gather_indices)
+
+Uses [`MLIR.Dialects.stablehlo.gather`](@ref) to get the values of `src` at the indices
+specified by `gather_indices`. If the indices are contiguous it is recommended to directly
+use [`MLIR.Dialects.stablehlo.dynamic_slice`](@ref) instead.
+"""
+function gather_getindex(
+    src::TracedRArray{T,N}, gather_indices::TracedRArray{Int64,2}
+) where {T,N}
+    @assert size(gather_indices, 2) == N
+
+    #! format: off
+    dimension_numbers = MLIR.API.stablehloGatherDimensionNumbersGet(
+        MLIR.IR.context(),
+        Int64(1), Int64[1],
+        Int64(N - 1), collect(Int64, 0:(N - 2)),
+        Int64(0), Int64[],
+        Int64(0), Int64[],
+        Int64(N), collect(Int64, 0:(N - 1)),
+        1
+    )
+    #! format: on
+
+    return reshape(
+        TracedRArray{T,2}(
+            (),
+            MLIR.IR.result(
+                MLIR.Dialects.stablehlo.gather(
+                    src.mlir_data,
+                    gather_indices.mlir_data;
+                    dimension_numbers,
+                    slice_sizes=fill(Int64(1), N),
+                    indices_are_sorted=false,
+                ),
+                1,
+            ),
+            (size(gather_indices, 1), 1),
+        ),
+        size(gather_indices, 1),
+    )
+end
+
 end # module Ops
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -1426,7 +1426,7 @@ specified by `scatter_indices` to the values in `updates`. If the indices are co
 is recommended to directly use [`MLIR.Dialects.stablehlo.dynamic_update_slice`](@ref)
 instead.
 """
-function scatter_setindex(
+@noinline function scatter_setindex(
     dest::TracedRArray{T,N},
     scatter_indices::TracedRArray{Int64,2},
     updates::TracedRArray{T,1},
@@ -1480,7 +1480,7 @@ Uses [`MLIR.Dialects.stablehlo.gather`](@ref) to get the values of `src` at the 
 specified by `gather_indices`. If the indices are contiguous it is recommended to directly
 use [`MLIR.Dialects.stablehlo.dynamic_slice`](@ref) instead.
 """
-function gather_getindex(
+@noinline function gather_getindex(
     src::TracedRArray{T,N}, gather_indices::TracedRArray{Int64,2}
 ) where {T,N}
     @assert size(gather_indices, 2) == N
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -1418,4 +1418,59 @@ julia> Reactant.@jit(
     end
 end
 
+"""
+    scatter_setindex(dest, scatter_indices, updates)
+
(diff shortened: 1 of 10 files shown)


avik-pal:
> what do you think about adding a section in the developer reference docs? so that wrapper array authors can add support for Reactant and we also remember how it should be done 😅
> 
> i actually want to add support for Reactant in my DeltaArrays package

That is a good idea, I will add that. 1 caveat of not having "AbstractWrappedArray" in Base Julia is that these new wrapper types won't go through our `AnyTracedRArray` dispatches. Technically, it is not a unique problem for us, and CUDA has been struggling with this for years now.

xref: https://github.com/JuliaLang/julia/issues/51910

src/TracedRArray.jl
```diff
@@ -191,16 +167,40 @@ function Base.getindex(a::WrappedTracedRArray, indices...)
...
+
+    non_contiguous_setindex = false
+    for idxs in indices
+        idxs isa Number && continue
+        contiguous = all(isone, diff(idxs))
+        # XXX: We want to throw error even for dynamic indexing
+        if typeof(contiguous) <: Bool && !contiguous
+            non_contiguous_setindex = true
+            break
+        end
```
avik-pal, line 185: How can we introduce a runtime error in the generated MLIR?

avik-pal:
Also we should revisit this at some point, but the current approach means that the generic dispatches convert a structured array into a dense array so we definitely use worse algorithms than are warranted

avik-pal:
need to add tests

mofeing:
> Also we should revisit this at some point, but the current approach means that the generic dispatches convert a structured array into a dense array so we definitely use worse algorithms than are warranted

yeah, can you leave a note in the docs for this please? IMHO this is ok to leave as default, and we can specialize for performance if we know how to do for the combination of array type and Ops

> That is a good idea, I will add that. 1 caveat of not having "AbstractWrappedArray" in Base Julia is that these new wrapper types won't go through our `AnyTracedRArray` dispatches. Technically, it is not a unique problem for us, and CUDA has been struggling with this for years now.

the way i've solved things like this in Tenet is using Holy traits with a default method that checks the trait (e.g. with `LinearAlgebra.diag`) but here is gonna be hard because we can't override the default method... maybe now that @wsmoses is readding the method overlay mechanism, we can overlay the default method and do sth like this?

```julia
iswrappedtraced(::TracedRArray) = true
iswrappedtraced(::Array) = false
iswrappedtraced(x::Diagonal) = iswrappedtraced(parent(x))
iswrappedtraced(::Type{<:TracedRArray}) = true
iswrappedtraced(::Type{<:Array}) = false
iswrappedtraced(::Type{Diagonal{T,N,A}}) where {T,N,A} = iswrappedtraced(A)

@reactant_override function LinearAlgebra.diag(x::AbstractArray{T,2}, k::Integer=0) where {T}
    if iswrappedtraced(x)
        y = materialize_traced_array(x)
        return diag(y, k)
    else
        # invoke diag(x) on NativeInterpreter
    end
end

function LinearAlgebra.diag(x::TracedRArray{T,2}, k::Integer=0) where {T}
    # here the code we currently have
end
```

the wrapped array lib dev who would like to add support for Reactant would then just need to correctly implement `iswrappedtraced` (suggestions welcome for a better name) and `materialize_traced_array`. furthermore, the dev could implement better methods if they know that this combination of method + array type doesn't need to be materialized (e.g. a Diagonal * Array could be implemented using `dot_general` with batching dims and no contracting dims).

the only problem i see is whether we can invoke methods in the NativeInterpreter from within a custom interpreter. @wsmoses @vchuravy @gbaraldi is that doable?

---

getting out of scope, i thing that sth Julia could have and would fix a loooot of problems like this are extendable unions. problem with abstract types is that you cannot take an external type and tell it "hey, you're now a 'son' of this abstract type because you fulfill this new interface i've defined". even with multiple inheritance there would still be problems. abstract types block composability.

an extendable union would be similar to `abc.ABCMeta` in Python, were you can _register_ new subclasses dynamically https://docs.python.org/3.13/library/abc.html#abc.ABCMeta
problem with this approach is that if not handled with care, i think it could lead to a lot of invalidation (or maybe no, i don't know much about the internals that deal with this).

avik-pal:
We can push an `is_wrapped_array` to ArrayInterface (easier for other packages to use rather than depending on Reactant). Then `is_wrapped_tracked_array` is simply `ancestor(x) isa TracedRArray`

mofeing:
> We can push an `is_wrapped_array` to ArrayInterface (easier for other packages to use rather than depending on Reactant).

super, i like it.

> Then `is_wrapped_tracked_array` is simply `ancestor(x) isa TracedRArray`

`..._traced_...`?

the only remaining problem is then how to call the original method in the native interpreter

src/TracedRArray.jl
mofeing: i thing you can't?

avik-pal: I was thinking of some sort of custom_call which jax is using here https://github.com/openxla/stablehlo/blob/main/docs/tutorials/jax-export.ipynb

```mlir
stablehlo.custom_call @shape_assertion(%1, %0) {api_version = 2 : i32, error_message = "Input shapes do not match the polymorphic shapes specification. Expected value >= 1 for dimension variable 'a'. Using the following polymorphic shapes specifications: args[0].shape = (a, 3, 224, 224). Obtained dimension variables: 'a' = {0} from specification 'a' for dimension args[0].shape[0] (= {0}), . Please see https://jax.readthedocs.io/en/latest/export/shape_poly.html#shape-assertion-errors for more details.", has_side_effect = true} : (tensor<i1>, tensor<i32>) -> ()
```

For tracedRarray indices we should probably always do a `(dynamic_)gather` (and we might be able to write an optimization later to transform that into a `slice` if contiguous.

avik-pal:
> the only remaining problem is then how to call the original method in the native interpreter

Looking at the current implementation, we could possibly do:

```julia
struct UseDefaultInterpreter end # Open to other names as well
```

```diff
--- a/src/Interpreter.jl
+++ b/src/Interpreter.jl
@@ -43,7 +43,8 @@ function set_reactant_abi(
     # the original call
     if f === Reactant.call_with_reactant
         arginfo2 = ArgInfo(fargs isa Nothing ? nothing : fargs[2:end], argtypes[2:end])
-        return abstract_call(interp, arginfo2::ArgInfo, si, sv, max_methods)
+        ret = abstract_call(interp, arginfo2::ArgInfo, si, sv, max_methods)
+        ret isa UseDefaultInterpreter || return ret
     end
 
     return Base.@invoke abstract_call_known(
```

```julia
@reactant_override function LinearAlgebra.diag(x::AbstractArray{T,2}, k::Integer=0) where {T}
    if iswrappedtraced(x)
        y = materialize_traced_array(x)
        return diag(y, k)
    end
    return UseDefaultInterpreter()
end
```

mofeing:
uuu i like it. i'm not very familiar with `set_reactant_abi` but it's that what @wsmoses added for solving dynamic dispatch on abs int?

wsmoses:
if you define the reactant override function as noinline, by default all methods within the override functions will be executed in the native interpreter

wsmoses:
to re-enter the abstract interp override, do `call_with_reactant(f, args...)`. This same behavior is true of anything in Ops, TracedUtils, or is otherwise in this whitelist: https://github.com/EnzymeAD/Reactant.jl/blob/65e9976ac1fc964d78b85298b0e4ca2a1cf74bb8/src/utils.jl#L92

mofeing:
> to re-enter the abstract interp override, do `call_with_reactant(f, args...)`. This same behavior is true of anything in Ops, TracedUtils, or is otherwise in this whitelist:

aha, now i get it

src/Ops.jl
```diff
@@ -1223,4 +1223,102 @@ function hlo_call(
...
 
+"""
+    scatter_setindex(dest, scatter_indices, updates)
+
+Uses [`MLIR.Dialects.stablehlo.scatter`](@ref) to set the values of `dest` at the indices
+specified by `scatter_indices` to the values in `updates`. If the indices are contiguous it
+is recommended to directly use [`MLIR.Dialects.stablehlo.dynamic_update_slice`](@ref)
+instead.
+"""
+function scatter_setindex(
```
wsmoses, line 1234: give it a @noinline ?

wsmoses (approved):

wsmoses (approved):
@avik-pal is there anything more pending here or good to merge?

avik-pal:
> @avik-pal is there anything more pending here or good to merge?

I will add some tests and then we should be good to merge

avik-pal:
Let's add https://github.com/EnzymeAD/Reactant.jl/pull/426 first. We need that to update the dispatches correctly.

wsmoses:
@avik-pal after this [and perhaps the specialfunctions PR] lets cut a release

avik-pal:
tests that should be passing are passing. 

https://github.com/EnzymeAD/Reactant.jl/actions/runs/12537476469/job/34961608153?pr=369#step:9:823 is probably unrelated to this PR

wsmoses:
Yeah that one feels like a jax error, let's merge

PR #375, opened 2024-12-13
No Transpose emission for 0 rank tensor

glou-nes:
We emit transpose for 0 rank tensor.  
```
> i = Reactant.ConcreteRNumber(5)
> @code_hlo optimize=false i<i
module {
  func.func @main(%arg0: tensor<i64>) -> (tensor<i1>, tensor<i64>) {
    %0 = stablehlo.transpose %arg0, dims = [] : (tensor<i64>) -> tensor<i64>
    %1 = stablehlo.compare  LT, %0, %0 : (tensor<i64>, tensor<i64>) -> tensor<i1>
    %2 = stablehlo.transpose %1, dims = [] : (tensor<i1>) -> tensor<i1>
    %3 = stablehlo.transpose %0, dims = [] : (tensor<i64>) -> tensor<i64>
    return %2, %3 : tensor<i1>, tensor<i64>
  }
}
```
Don't emit `stablehlo.transpose` operation for this case, and so simplify reading of unoptimized IR.
```
@code_hlo optimize=false i<i
module {
  func.func @main(%arg0: tensor<i64>) -> tensor<i1> {
    %0 = stablehlo.compare  LT, %arg0, %arg0 : (tensor<i64>, tensor<i64>) -> tensor<i1>
    return %0 : tensor<i1>
  }
}

```


wsmoses (commented):
this should get optimized by enzyme-hlo-opt downstream later regardless, but I'm not opposed if this speeds up compiletime

wsmoses:
@glou-nes can this be rebased (it merge conflicts)?

avik-pal (approved):
test failures are unrelated

PR #353, opened 2024-12-09
precompile first interpretation 

glou-nes:
Partial fix #348
remove a assertion in XLA. Get some result.
Now:
```
using Reactant.XLA
cpu = XLA.CPUClient()
x = Reactant.ConcreteRArray(randn(Float64, 2, 2); client=cpu)
@time @code_hlo sum(x)
5.14s
```
On master:
```
 61.77 s
```

Use `SnoopCompile` to track invalidation, get several invalidation in the abstract interpreter. I already get similar invalidation because of REPL abstract interpreter in Julia 1.12. I don't know how to fix it. Some in the list are probably easier to handle.
```
lock_mi_inference(interp::GPUCompiler.GPUInterpreter, mi::Core.MethodInstance) @ GPUCompiler /depot/packages/GPUCompiler/GnbhK/src/jlgen.jl:405
unlock_mi_inference(interp::GPUCompiler.GPUInterpreter, mi::Core.MethodInstance) @ GPUCompiler /depot/packages/GPUCompiler/GnbhK/src/jlgen.jl:406
add_remark!(interp::GPUCompiler.GPUInterpreter, sv::Core.Compiler.InferenceState, msg) @ GPUCompiler /depot/packages/GPUCompiler/GnbhK/src/jlgen.jl:408
may_compress(interp::GPUCompiler.GPUInterpreter) @ GPUCompiler /depot/packages/GPUCompiler/GnbhK/src/jlgen.jl:413
may_discard_trees(interp::GPUCompiler.GPUInterpreter) @ GPUCompiler /depot/packages/GPUCompiler/GnbhK/src/jlgen.jl:414
verbose_stmt_info(interp::GPUCompiler.GPUInterpreter) @ GPUCompiler /depot/packages/GPUCompiler/GnbhK/src/jlgen.jl:415
abstract_call_known(interp::Enzyme.Compiler.Interpreter.EnzymeInterpreter{Handler}, f, arginfo::Core.Compiler.ArgInfo, si::Core.Compiler.StmtInfo, sv::Union{Core.Compiler.IRInterpretationState, Core.Compiler.InferenceState}, max_methods::Int64) where Handler @ Enzyme.Compiler.Interpreter /depot/packages/Enzyme/6C71q/src/compiler/interpreter.jl:906
[...]

PR #377, opened 2024-12-14
fix: include files if they end with .jl

avik-pal:
https://github.com/LuxDL/Lux.jl/actions/runs/12326671368/job/34407753923?pr=1133#step:7:61 lol look me a few days to figure out what was happening. If coverage is enabled we have .cov files generated which will screw up the import of the file.


mofeing:
Ahhh that's why #353 was failing!! thanks @avik-pal 

cc @glou-nes

mofeing (approved):

Issue #381, opened 2024-12-15
SpecialFunctions meta-issue


wsmoses:
[ ] [SpecialFunctions.gamma](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.gamma)
[ ] [SpecialFunctions.loggamma](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.loggamma)
[ ] [SpecialFunctions.logabsgamma](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.logabsgamma)
[ ] [SpecialFunctions.loggamma1p](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.loggamma1p)
[ ] [SpecialFunctions.logfactorial](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.logfactorial)
[ ] [SpecialFunctions.digamma](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.digamma)
[ ] [SpecialFunctions.invdigamma](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.invdigamma)
[ ] [SpecialFunctions.trigamma](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.trigamma)
[ ] [SpecialFunctions.polygamma](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.polygamma)
[ ] [SpecialFunctions.gamma_inc](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.gamma_inc)
[ ] [SpecialFunctions.gamma_inc_inv](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.gamma_inc_inv)
[ ] [SpecialFunctions.loggammadiv](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.loggammadiv)
[ ] [SpecialFunctions.gammax](https://specialfunctions.juliamath.org/stable/functions_list/#SpecialFunctions.gammax)

PR #384, opened 2024-12-16
`SpecialFunctions` simple functions

glou-nes:
Simple case of #381.


ext/ReactantSpecialFunctions.jl
```diff
@@ -0,0 +1,125 @@
+using Reactant: Reactant, Ops
+using Reactant.TracedUtils
+
+function SpecialFunctions.gamma(x::TracedRNumber{T}) where {T}
+    x = promote_to(TracedRNumber{Float64}, x)
```
wsmoses, line 5: do we need to enforce float64 here or could we use the present type?

ext/ReactantSpecialFunctions.jl
```diff
@@ -0,0 +1,125 @@
...
+
+#utilities...
+
+function SpecialFunctions.erf(x::TracedRNumber{T}) where {T}
+    x = promote_to(TracedRNumber{Float64}, x)
+    return Ops.erf(x)
+end
+
+function SpecialFunctions.erf(x::TracedRNumber{T}, y::TracedRNumber{T}) where {T}
+    return y - x
```
wsmoses, line 80: is this right?

ext/ReactantSpecialFunctions.jl
```diff
@@ -0,0 +1,125 @@
...
+
+#Airy and Related Functions
+
+#Bessel ...
+
+#Elliptic Integrals
+
+function SpecialFunctions.zeta(z::TracedRNumber{T}, s::TracedRNumber{T}) where {T}
+    return Ops.zeta(z,s)
+end
```
wsmoses, line 125: can you add a test file comparing correctness against specialfunctions proper?

wsmoses:
amazing!

ext/ReactantSpecialFunctions.jl
glou-nes: missing erf, thank!

ext/ReactantSpecialFunctions.jl
glou-nes: Probably not `Float64` but several Ops need:
```
op operand #0 must be tensor of f4E2M1FN type or f6E2M3FN type or f6E3M2FN type or f8E3M4 type or f8E4M3 type or f8E4M3FN type or f8E4M3FNUZ type or f8E4M3B11FNUZ type or f8E5M2 type or f8E5M2FNUZ type or f8E8M0FNU type or 16-bit float or 32-bit float or 64-bit float or bfloat16 type values, but got 'tensor<i64>'
```

wsmoses: Maybe only define it for abstractfloat then? Or a reactantfloat if we have one

Project.toml
```diff
@@ -32,6 +32,7 @@ path = "lib/ReactantCore"
 ReactantAbstractFFTsExt = "AbstractFFTs"
 ReactantArrayInterfaceExt = "ArrayInterface"
 ReactantNNlibExt = "NNlib"
+ReactantSpecialFunctions = "SpecialFunctions"
```
avik-pal, line 35: For consistent naming lets make it `ReactantSpecialFunctionsExt`

ext/ReactantSpecialFunctions.jl
```diff
@@ -0,0 +1,125 @@
...
+
+function SpecialFunctions.digamma(x::TracedRNumber{T}) where {T}
+    x = promote_to(TracedRNumber{Float64}, x)
+    return Ops.digamma(x)
+end
+
+# SpecialFunctions.invdigamma
+
+function SpecialFunctions.trigamma(x::TracedRNumber{T}) where {T}
+    return Ops.polygamma(Ops.constant(2.0), x)
```
avik-pal, line 40:
```suggestion
    return Ops.polygamma(Ops.constant(T(2)), x)
```

ext/ReactantSpecialFunctions.jl
avik-pal: Lets do `ReactantFloat = Union{Float16, BFloat16, Float32, Float64}` and update ReactantPrimitive to contain `ReactantFloat`

glou-nes: Sure! I was following `Ops.jl` definition. I can update the file in a following PR.

Project.toml
```diff
@@ -59,3 +60,4 @@ julia = "1.10"
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
+SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
```
avik-pal, line 63: needs to be in weakdeps

ext/ReactantSpecialFunctions.jl
```diff
@@ -0,0 +1,121 @@
+using Reactant: Reactant, Ops
```
avik-pal, line 1: missing a `module ... `, also needs a `using SpecialFunctions`

ext/ReactantSpecialFunctions.jl
wsmoses: ++, though also at one point we should properly add bfloat16 too

glou-nes: I use the following to support the integer which are supported by `SpecialFunctions`. 
```
function SpecialFunctions.gamma(x::TracedRNumber{<:Number})
    return gamma(promote_to(TracedRNumber{Float64}, x))
end

function SpecialFunctions.gamma(x::TracedRNumber{T}) where {T<:ReactantFloat}
    return exp(Ops.lgamma(x))
end
```

avik-pal: you can also do `float(x)` (we have that dispatch in TracedRNumber.jl) to match how SpecialFunctions writes out the dispatches

ext/ReactantSpecialFunctionsExt.jl
```diff
@@ -0,0 +1,132 @@
...
+
+# SpecialFunctions.invdigamma
+
+function SpecialFunctions.trigamma(x::TracedRNumber{T}) where {T}
+    return Ops.polygamma(Ops.constant(T(1)), x)
+end
+
+function SpecialFunctions.polygamma(
+    n::TracedRNumber{T}, x::TracedRNumber{T}
+) where {T<:ReactantFloat}
```
mofeing, line 50: as above, i don't think using a `Union` to constraint the accepted types is the best way. a user can implement their own reduced float type in Julia, declare how to lower to MLIR and it should work.

mofeing (commented):
so regarding using `Union` to constraint types in methods, i don't think it's ~a good~ the best idea. i'm pro the idea by @oxinabox that types in method signatures should be just used for dispatching, not for as type constraints.

the reasons why are:
1. HLO dialects and XLA can (and will) change: checkout for example the bug i informed about `stablehlo.cbrt` on complex numbers, they are probably gonna remove support for it. this forces us to be super cautious of every update on it (which we kinda are already, but we wouldn't need to update the code).
2. it goes a lil bit against composability: a user can come, implement their own reduced float type in Julia, tell Reactant how to lower it to MLIR and it should work. `Union`s have the problem that are not extendable.

this is how i developed Ops in principle

also check it out, that as of the current implementation, there are are some methods that will fail. take `SpecialFunctions.gamma`
```julia
function SpecialFunctions.gamma(x::TracedRNumber{T}) where {T<:ReactantFloat}
    return exp(Ops.lgamma(x))
end
```

it will be correctly dispatched on `TracedRNumber{Float64}`, but what will happen with `TracedRNumber{Int64}` or `TracedRNumber{ComplexF64}`. if `SpecialFunctions.gamma`'s definition is wide enough (like no type constraint, which i'm not sure it does but it's an example), then it will try to trace, won't throw a `MethodError`, and... it will probably do nasty things inside.

IMO it's not bad that it fails on MLIR while it doesn't segfault. and if it fails, you can use the `with_debug` func to get the full Julia + MLIR stacktrace.

glou-nes:
> so regarding using `Union` to constraint types in methods, i don't think it's ~a good~ the best idea. i'm pro the idea by @oxinabox that types in method signatures should be just used for dispatching, not for as type constraints.
> 
> the reasons why are:
> 
>     1. HLO dialects and XLA can (and will) change: checkout for example the bug i informed about `stablehlo.cbrt` on complex numbers, they are probably gonna remove support for it. this forces us to be super cautious of every update on it (which we kinda are already, but we wouldn't need to update the code).
> 
>     2. it goes a lil bit against composability: a user can come, implement their own reduced float type in Julia, tell Reactant how to lower it to MLIR and it should work. `Union`s have the problem that are not extendable.
> 
> 
> this is how i developed Ops in principle
> 
> also check it out, that as of the current implementation, there are are some methods that will fail. take `SpecialFunctions.gamma`
> 
> ```julia
> function SpecialFunctions.gamma(x::TracedRNumber{T}) where {T<:ReactantFloat}
>     return exp(Ops.lgamma(x))
> end
> ```
> 
> it will be correctly dispatched on `TracedRNumber{Float64}`, but what will happen with `TracedRNumber{Int64}` or `TracedRNumber{ComplexF64}`. if `SpecialFunctions.gamma`'s definition is wide enough (like no type constraint, which i'm not sure it does but it's an example), then it will try to trace, won't throw a `MethodError`, and... it will probably do nasty things inside.
> 
> IMO it's not bad that it fails on MLIR while it doesn't segfault. and if it fails, you can use the `with_debug` func to get the full Julia + MLIR stacktrace.

Thank for the answer! Firstly, I want to add that for your example, it supports TracedRNumber{Int64} using the fallback method here: Is it not enough to get composability back?
```
for fn in [:gamma, :loggamma, :digamma, :erf, :erfc]
    @eval(function SpecialFunctions.$fn(x::TracedRNumber{<:Number})
        return $fn(float(x))
    end)
end
```
My major problem here is the difference between `SpecialFunctions` and `StableHLO` semantics. it's really similar to `stablehlo.cbrt` (complex number and integers are not supported with StableHLO) but are supported by SpecialFunctions. I tried to follow SpecialFunctions typing (which is a bit strict using Number and Integer for instance) and error in this file with complex number for instance. I suppose this file handle both semantics to support smooth integration with functions of `SpecialFunctions` and a kind of opaque interface. I understand the idea for Ops need to be completely unconstrained.  It is not better to get a Julia dispatch error than a XLA runtime for that kind of external integration?

test/integration/special_functions.jl
```diff
@@ -0,0 +1,35 @@
...
+        (:erf, 2),
+        (:erfc, 1),
+        (:logerf, 2),
+        (:erfcx, 1),
+        (:logerfc, 1),
+        (:logerfcx, 1),
+    ]
+        x = values[1:n_args]
+        @eval @test @jit(SpecialFunctions.$op(ConcreteRNumber.($x)...)) ≈
+            SpecialFunctions.$op($x...)
```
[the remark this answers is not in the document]
glou-nes: I was using `rand` but get several times singular points. I suppose it can be justified to get both: here to test different types and more complex function and in `ops` to get a good cover of shlo/mhlo operation unit test.

mofeing:
> Thank for the answer! Firstly, I want to add that for your example, it supports TracedRNumber{Int64} using the fallback method here: Is it not enough to get composability back?
> 
> ```
> for fn in [:gamma, :loggamma, :digamma, :erf, :erfc]
>     @eval(function SpecialFunctions.$fn(x::TracedRNumber{<:Number})
>         return $fn(float(x))
>     end)
> end
> ```

okay, but keep in mind that `float` on `Complex` returns a `Complex{<:AbstractFloat}`.

how about this: you remove the type constraint of the method i pointed (or leave it to just `TracedRNumber` or `TracedRNumber{<:Real}`), and you call `float` inside. no need for `Union`s that way.

> My major problem here is the difference between SpecialFunctions and StableHLO semantics. it's really similar to stablehlo.cbrt (complex number and integers are not supported with StableHLO) but are supported by SpecialFunctions

keep in mind that the problem is that StableHLO spec supports complex numbers for `cbrt` but XLA doesn't. and StableHLO spec should be a reflect of XLA so it's a bug in the spec and seems like they are going to remove it from the spec.

> I suppose this file handle both semantics to support smooth integration with functions of SpecialFunctions and is a kind of opaque interface. I understand the idea for Ops need to be completely unconstrained. It is not better to get a Julia dispatch error than a XLA runtime for that kind of external integration?

so i think there are different visions in the project in this sense and probably we need to discuss in the next meeeting, but undoubtely it's better to have a Julia error than a XLA error. my opinion is that when choosing between better errors on the Julia side or better dispatch/composability (at least **in Ops**), we should prefer better dispatch/composability. but there can be different opinions in this aspect.

mofeing:
> a completely generic fallback with `TracedRNumber` is better! Both version, generated runtime error with complex anyway. At least now, it's simpler.
> 
> ```
> error: 'chlo.lgamma' op operand #0 must be tensor of f4E2M1FN type or f6E2M3FN type or f6E3M2FN type or f8E3M4 type or f8E4M3 type or f8E4M3FN type or f8E4M3FNUZ type or f8E4M3B11FNUZ type or f8E5M2 type or f8E5M2FNUZ type or f8E8M0FNU type or 16-bit float or 32-bit float or 64-bit float or bfloat16 type values, but got 'tensor<complex<f64>>'
> ```

you can put a constraint on `<:Real` in that case

ext/ReactantSpecialFunctionsExt.jl
```diff
@@ -0,0 +1,132 @@
...
+    return Ops.lgamma(x)
+end
+
+function SpecialFunctions.loggamma1p(x::TracedRNumber{T}) where {T}
+    return loggamma(1 + x)
+end
+
+function SpecialFunctions.logfactorial(
+    x::TracedRNumber{T}
+) where {T<:Union{Int8,UInt8,Int16,UInt16,Int32,UInt32,Int64,UInt64}}
```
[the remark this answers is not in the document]
wsmoses: I mean this is where I think we should have a ReactantInteger = Union{actual types}, which we can update once if the spec / support changes throughout

ext/ReactantSpecialFunctionsExt.jl
wsmoses: yeah, but we don't want to overload for the user defined union type, but let the default lowering apply.

For example

```julia
struct DoubleFloat{T} <: Real
  x::T
  y::T
end

function SpecialFunctions.gamma(x::DoubleFloat)
   return gamma(x) + gamma(y) * .001
end
```

We don't want their new real type to hit our override and try to call ops. Instead we want it to call the default specialfunctions.gamma here (which will eventually dispatch to our specialfunctions.gamma for the base types)

wsmoses:
@mofeing I disagree with explicitly marking broader support here (e.g. of Number). In particular, we only want to override methods which we know we can implement fully with Ops (aka with stablehlo).

See https://github.com/EnzymeAD/Reactant.jl/pull/384#discussion_r1887189198 for example

ext/ReactantSpecialFunctionsExt.jl
wsmoses: I think this might literally be testable with say https://github.com/JuliaMath/DoubleFloats.jl

wsmoses:
yeah this looks like it needs a rebase (and also using concrete types per above)

glou-nes:
Sure! I will do it today!

ext/ReactantSpecialFunctionsExt.jl
```diff
@@ -0,0 +1,106 @@
...
+using Reactant.TracedRNumberOverrides: float
+
+for fn in [:digamma, :erf, :erfc, (:loggamma, :lgamma)]
+    (fns, fno) = fn isa Tuple ? fn : (fn, fn)
+    @eval(function SpecialFunctions.$fns(x::TracedRNumber{<:Real})
+        return Ops.$fno(float(x))
+    end)
+end
+
+function SpecialFunctions.gamma(x::TracedRNumber{<:Real})
```
wsmoses, line 13: This should be <: ReactantFloat right?

ext/ReactantSpecialFunctionsExt.jl
```diff
@@ -0,0 +1,106 @@
...
+    return exp(Ops.lgamma(float(x)))
+end
+
+function SpecialFunctions.gamma(n::TracedRNumber{<:Integer})
```
wsmoses, line 17: and here ReactantInt?

wsmoses (approved):

wsmoses:
The round ceil floor tests now fail and should be fixed but otherwise LGTM to merge!

wsmoses:
hmm CI still does not seem to like it:

```julia
  Test threw exception
  Expression: #= /home/runner/work/Reactant.jl/Reactant.jl/test/basic.jl:582 =# @jit(floor.(ConcreteRNumber.([5617217728973374046 -3007515435434531812 -5171351284935898551; -157483289222511147 -424687138072795833 -2567171450103491194; 6843301087142890975 1766303807932526665 -9188973971342681186]))) == floor.([5617217728973374046 -3007515435434531812 -5171351284935898551; -157483289222511147 -424687138072795833 -2567171450103491194; 6843301087142890975 1766303807932526665 -9188973971342681186])
  MethodError: no method matching floor(::Reactant.TracedRNumber{Int64})
  
  Closest candidates are:
    floor(::Type{Bool}, ::AbstractFloat)
     @ Base float.jl:389
    floor(::Type{Dates.Date}, ::Union{Dates.Day, Dates.Week, Dates.TimePeriod, Dates.TimeType}, ::Type{P}) where P<:Dates.Period
     @ Dates /opt/hostedtoolcache/julia/1.10.7/x64/share/julia/stdlib/v1.10/Dates/src/rounding.jl:287
    floor(::Missing; sigdigits, digits, base)
     @ Base missing.jl:155
    ...
  
  Stacktrace:
    [1] macro expansion
      @ ~/work/Reactant.jl/Reactant.jl/src/utils.jl:0 [inlined]
    [2] call_with_reactant(::typeof(floor), ::Reactant.TracedRNumber{Int64})
      @ Reactant ~/work/Reactant.jl/Reactant.jl/src/utils.jl:527
    [3] _broadcast_getindex_evalf
```

wsmoses:
@glou-nes can you rebase?

src/TracedRNumber.jl
```diff
@@ -231,6 +229,28 @@ function Base.float(x::TracedRNumber{T}) where {T}
...
+
+function Base.round(A::TracedRNumber{<:ReactantFloat}, ::RoundingMode{R}) where {R}
+    if R == :Nearest
+        Ops.round_nearest_even(A)
+    elseif R == :Up
+        Ops.ceil(A)
+    elseif R == :Down
+        Ops.floor(A)
+    else
+        @error "$R is unsupported"
```
avik-pal, line 246: throw an actual error instead of logging it.

wsmoses:
@glou-nes it looks like some test is segfaulting here, can you isolate it into a MWE to debug?

test/integration/special_functions.jl
```diff
@@ -23,7 +23,7 @@ end
 ]
     for data in ([0.5, 0.6], [2, 4])
         x = data[1:n_args]
-        #@eval @test @≈ @jit(SpecialFunctions.$op(ConcreteRNumber.($x)...)) SpecialFunctions.$op($x...)
+        @eval @test @≈ @jit(SpecialFunctions.$op(ConcreteRNumber.($x)...)) SpecialFunctions.$op($x...)
```
mofeing: just a tip: if you put the function itself instead of the `Symbol`, you won't need the `@eval` so it will be faster

(this testset doesn't take too much time, but take this as a more general tip)

glou-nes:
> @glou-nes it looks like some test is segfaulting here, can you isolate it into a MWE to debug?

It seems that the segfault was from this: Anyway, I removed the second part it's not needed.
```
Base.round(A::TracedRNumber{<:ReactantFloat}) = Ops.round_nearest_even(A)
Base.floor(A::TracedRNumber{<:ReactantFloat}) = Ops.floor(A)
Base.ceil(A::TracedRNumber{<:ReactantFloat}) = Ops.ceil(A)

Base.round(A::TracedRNumber{<:Integer}) = A
Base.floor(A::TracedRNumber{<:Integer}) = A
Base.ceil(A::TracedRNumber{<:Integer}) = A
```

PR #400, opened 2024-12-19
feat: bump jll version and add opt pass

avik-pal:


avik-pal:
@wsmoses did something change in Enzyme in the new JLL? https://github.com/EnzymeAD/Reactant.jl/actions/runs/12405818372/job/34633303834?pr=400#step:9:488

wsmoses:
Oh excellent. Yeah I added a better error message for us accidentally segfaulting sometimes due to asking enzyme to return the wrong number of args from an autodiff call

src/Interpreter.jl
```diff
@@ -283,7 +283,7 @@ function overload_autodiff(
         end
     end
     for (i, act) in enumerate(activity)
-        if act == enzyme_out || (reverse && (act == enzyme_dup || act == enzyme_dupnoneed))
```
avik-pal, line 286: I am not 100% sure if this is the right fix, but it fixes the tests

wsmoses: yeah I think it is

PR #405, opened 2024-12-19
Inference cache

jumerckx:
~~This doesn't work yet! It only caches the top level codeinfo, some other absint method needs to be overwritten to do it properly.
Only tested on Julia 1.11~~
CI passes, and it seems like the cicache is being filled, I don't have a good test to show improvements though.
```julia
f(a, b) = a .* b
a = rand(10)
b = rand(10)

a_ra = Reactant.to_rarray(a)
b_ra = Reactant.to_rarray(b)

g(a, b) = @noinline f(a, b)

@code_hlo(g(a, b))

result = CC.get(CC.code_cache(Reactant.ReactantInterpreter()), Base.method_instance(f, Tuple{Vector{Float64}, Vector{Float64}}), nothing)

```

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -346,7 +346,14 @@ function call_with_reactant_generator(
     method = mi.def
 
     @static if VERSION < v"1.11"
-        ir, rt = CC.typeinf_ircode(interp, method, mi.specTypes, mi.sparam_vals, nothing)
+        # For older Julia versions, we vendor in some of the code to prevent
+        # having to build the MethodInstance twice.
+        result = CC.InferenceResult(mi, CC.typeinf_lattice(interp))
+        frame = CC.InferenceState(result, :no, interp)
+        @assert !isnothing(frame)
+        CC.typeinf(interp, frame)
+        ir = CC.run_passes(frame.src, CC.OptimizationState(frame, interp), result, nothing)
+        rt = CC.widenconst(CC.ignorelimited(result.result))
     else
         ir, rt = CC.typeinf_ircode(interp, mi, nothing)
     end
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -271,6 +294,7 @@ function call_with_reactant_generator(
 )
     @nospecialize
     args = redub_arguments
+    Core.println("in call_with_reactant_generator: ", redub_arguments)
```
wsmoses, line 297: remove the debug stmt?

src/utils.jl
```diff
@@ -320,72 +344,21 @@ function call_with_reactant_generator(
...
-
-    if DEBUG_INTERP[]
-        safe_print("opt.src", opt.src)
-    end
-
-    caller = frame.result
-    @static if VERSION < v"1.11-"
-        ir = Core.Compiler.run_passes(opt.src, opt, caller)
+    @static if VERSION < v"1.11"
+        ir, rt = CC.typeinf_ircode(interp, method, mi.specTypes, mi.sparam_vals, nothing)
```
wsmoses, line 350: ah yeah so this is why I vendored the internals of that code. This causes it to re-run "get methodinstance", which is rather expensive. Can we continue to vendor code which doesn't re-run "get method instance" for 1.10?

jumerckx: Okay

mofeing:
does this fix #353?

jumerckx:
> does this fix https://github.com/EnzymeAD/Reactant.jl/pull/353?

You mean the segfaults or something else? I doubt this will do anything to the segfault but you never know. Is there a MWE or is still only observed in CI?

wsmoses (approved):

wsmoses:
> > does this fix #353?
> 
> You mean the segfaults or something else? I doubt this will do anything to the segfault but you never know. Is there a MWE or is still only observed in CI?

Yeah I think this is unrelated (and just compile time speedups)

PR #407, opened 2024-12-20
Pythoncall

wsmoses:


ext/ReactantPythonCallExt.jl
```diff
@@ -0,0 +1,24 @@
+module ReactantPythonCallExt
+
+using PythonCall
+using Reactant: Reactant, TracedRArray, AnyTracedRArray, MLIR, TracedRNumber
+using ReactantCore: @trace
+
+using PythonCall
+
+const jax = pyimport("jax")
```
mofeing, line 9: putting this here can be problematic. what if the user is using Reactant and PythonCall, but not together? the user most probably hasn't installed jax and this will fail (e.g. me and the VQE sim).

what i've done for Tenet is to add this `pyimport` inside the function. another solution is to use a cache.

wsmoses: resolved

PR #420, opened 2024-12-23
feat: add zero and fill! for ConcreteRArray

avik-pal:


avik-pal:
@wsmoses seems like a recent change is making it always print https://github.com/EnzymeAD/Reactant.jl/actions/runs/12464459169/job/34788567075?pr=420#step:5:942

wsmoses (approved):

wsmoses:
> @wsmoses seems like a recent change is making it always print https://github.com/EnzymeAD/Reactant.jl/actions/runs/12464459169/job/34788567075?pr=420#step:5:942

ah right I added prints to the the lower-kernel pass to debug why it wasn't running. I forgot that it was run every time (not just for the kernel case), in any case it'll be gone in another quick jll bump

PR #427, opened 2024-12-27
CUDA kernels take 3

wsmoses:


test/runtests.jl
```diff
@@ -59,6 +59,7 @@ const REACTANT_TEST_GROUP = lowercase(get(ENV, "REACTANT_TEST_GROUP", "all"))
     end
 
     if REACTANT_TEST_GROUP == "all" || REACTANT_TEST_GROUP == "integration"
+        @safetestset "CUDA" include("integration/cuda.jl")
```
avik-pal, line 62: conditionally run with a `CUDA.functional()`?

wsmoses: So we have that in the test itself, but it’s actually important that we can compile cuda code even without a cuda gpu (since we want to take existing cuda kernels and rewrite them to whatever hardware).

src/utils.jl
```diff
@@ -364,7 +364,10 @@ function call_with_reactant_generator(
         ir, rt = CC.typeinf_ircode(interp, mi, nothing)
     end
 
-    ir, any_changed = rewrite_insts!(ir, interp)
+    if !is_reactant_method(mi::Core.MethodInstance)
```
wsmoses, line 367: @jumerckx as a note. Something that I missed earlier, we previously had a catch all like this that we'll only do rewrite insts if the top level function is not a reactant method, but I think this was accidentally removed by your earlier PR. In any case restored here (and once CUDA lands will actually be tested it is done)

PR #431, opened 2024-12-28
Test YaoBlocks

mofeing:
Needed to check that #426 doesn't break integration with YaoBlocks


mofeing:
> LLVM ERROR: Program used external function 'sincos' which could not be resolved!

ah dammit, we are aware of this error. apparently macos x64 libc doesn't export math symbols. it's the reason why we don't test some numeric ops on macos x64.

should be fixed now by skipping the test on that platform

wsmoses:
eventually we can fix that by manually exporting a sincos ourselves

mofeing:
> eventually we can fix that by manually exporting a sincos ourselves

we could do it by adding a file to the build conditionally, but the question again is if we really want to support macOS x64

wsmoses:
> > eventually we can fix that by manually exporting a sincos ourselves
> 
> we could do it by adding a file to the build conditionally, but the question again is if we really want to support macOS x64

At minimum because selfishly my personal laptop is an Intel Mac it would be nice. But also I can fix that problem in another way too xD

mofeing:
> At minimum because selfishly my personal laptop is an Intel Mac it would be nice. But also I can fix that problem in another way too xD

ahh ups jajajaj

okay then, how could we export those symbols? i don't understand why those symbols are missing in the first place.
do we need to reimplement them manually or it's just that the symbol filtering on mac is filtering too much?

if it's the symbol filter, then #421 should fix it because it refactors the filter to just stop exporting symbols that match the glob pattern `*llvm*`

PR #433, opened 2024-12-29
feat: support Base.stack

### LINKED ISSUE #298: 2nd order AD fails
```julia
using Reactant, Enzyme, Lux, Random, LinearAlgebra

const xdev = reactant_device()
const cdev = cpu_device()

model = Dense(5 => 5, gelu);
ps, st = Lux.setup(Random.default_rng(), model) |> xdev;
potential = StatefulLuxLayer{true}(model, ps, st)

# Currently EnzymeMLIR doesn't support batching so we force chunksize to 1
function ∇potential(potential, x)
    J = reshape(only(Enzyme.jacobian(Forward, potential, x; chunk=Val(1))), :, length(x))
    J_diag = @allowscalar diag(J)
    return reshape(J_diag, size(x))
end

function ∇²potential(potential, x)
    J = reshape(only(
        Enzyme.jacobian(Forward, Base.Fix1(∇potential, potential), x; chunk=Val(1))
    ), :, length(x))
end

x_ra = randn(Float32, 5, 3) |> xdev

@code_hlo ∇²potential(potential, x_ra)
```

A non-minimal example taken from https://github.com/LuxDL/Lux.jl/issues/614

avik-pal:
fixes #298 

- [ ] tests


wsmoses (approved):
Lgtm but probably needs a test

PR #436, opened 2024-12-30
fix: try building with cudnn 9.4

avik-pal:


wsmoses:
So no ci here will do anything we'll need to test this with a PR on yggdrassil

wsmoses (approved):

Issue #393, opened 2024-12-18
AbstractInterpreter and Recursive Calls into Overlay-ed Functions


avik-pal:
See the linked comments for more details.

my understanding was that this should call the non-overlayed version. But here I get 

```
┌ Warning: Directly writing to an array using Random.jl functions inside ReactantInterpreter will generate a constant array in the IR. Use with caution.
└ @ Reactant /mnt/software/lux/Reactant.jl/src/Overlay.jl:84
Unreachable reached at 0x7a64877b6b16

[1417783] signal 4 (2): Illegal instruction
in expression starting at REPL[7]:1
fn2 at ./REPL[6]:4 [inlined]
opaque closure at ./<missing>:0
unknown function (ip: 0x7a64877b6bff)
fn2 at ./REPL[6]:2 [inlined]
call_with_reactant at /mnt/software/lux/Reactant.jl/src/utils.jl:0
#8 at /mnt/software/lux/Reactant.jl/src/TracedUtils.jl:210
block! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Block.jl:201
unknown function (ip: 0x7a64877b6316)
#make_mlir_fn#1 at /mnt/software/lux/Reactant.jl/src/TracedUtils.jl:197
make_mlir_fn at /mnt/software/lux/Reactant.jl/src/TracedUtils.jl:117 [inlined]
#10 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:295 [inlined]
block! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Block.jl:201
#9 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:294 [inlined]
mmodule! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Module.jl:92
unknown function (ip: 0x7a64877b5976)
#compile_mlir!#8 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:291
compile_mlir! at /mnt/software/lux/Reactant.jl/src/Compiler.jl:290 [inlined]
#6 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:285 [inlined]
context! at /mnt/software/lux/Reactant.jl/src/mlir/IR/Context.jl:76
unknown function (ip: 0x7a64877b2a76)
#compile_mlir#5 at /mnt/software/lux/Reactant.jl/src/Compiler.jl:283
[...]

PR #440, opened 2024-12-31
Fix mul overload

wsmoses:


src/Overlay.jl
```diff
@@ -130,14 +130,18 @@ for (cT, aT, bT) in (
             if any(Base.Fix2(isa, TracedRArray) ∘ ancestor, (C, A, B))
                 TracedLinearAlgebra.overloaded_mul!(C, A, B, α, β)
             else
-                LinearAlgebra._mul!(C, A, B, α, β)
+                LinearAlgebra.mul!(C, A, B, α, β)
```
avik-pal, line 133: this will very likely trigger the illegal instruction https://github.com/EnzymeAD/Reactant.jl/issues/393

wsmoses:
@avik-pal it looks like this is hitting some scalar index stuff?

wsmoses:
Yeah we need to change aos to anytracedarray

src/Overlay.jl
wsmoses: There was an absint behavior that we had a while ago but was accidentally removed in a refactor which I re added with the last cuda stuff.

I hope it means we won't hit it

avik-pal: I checked it again in the stack PR, the crash persists

wsmoses: Well darn, okay yeah if CI yells I think we need to make a separate noinline function which is called in the override, which itself calls mul

avik-pal: lets add a test like https://github.com/EnzymeAD/Reactant.jl/blob/main/test/basic.jl#L864 to verify this

wsmoses:
Okay @avik-pal tesfs pass so I’m going to merge

PR #443, opened 2024-12-31
feat: missing BF16 dispatches

avik-pal:


src/mlir/IR/Type.jl
```diff
@@ -210,6 +219,17 @@ Checks whether the given type is an f16 type.
...
 isf16(type::Type) = API.mlirTypeIsAF16(type)
 
+"""
+    isbf16(type)
+
+Checks whether the given type is an bf16 type.
+"""
+function isbf16 end
+
+if isdefined(Core, :BFloat16)
```
wsmoses, line 229: This one doesn’t need to be conditionally defined

wsmoses (approved):

PR #450, opened 2025-01-01
fix: use pip for jax

avik-pal:

PR #451, opened 2025-01-01
Use pip deps for jax

wsmoses:


wsmoses:
Avik is faster than me :P (x/ref https://github.com/EnzymeAD/Reactant.jl/pull/450 )

PR #453, opened 2025-01-01
Revert modularize bazel build to unbreak yggy

wsmoses:
Doing this temporarily for now so things can build.

@mofeing after this, you can re-merge but please test on ygg to confirm


wsmoses:
@mofeing per the jll I'm going to merge this for now [with the new jll] and let you make a new PR un-reverting (and ideally tested against yggy?)

PR #458, opened 2025-01-02
feat: `no_nan` mode and more passes

avik-pal:


test/compile.jl
```diff
@@ -101,3 +101,29 @@ end
...
+
+    fn(x, y) = begin
+        c = x .+ y
+        return c .- y
+    end
+
+    hlo = @code_hlo fn(x_ra, y_ra)
+    @test occursin("subtract", repr(hlo))
+    @test occursin("add", repr(hlo))
+    hlo = @code_hlo no_nan=true fn(x_ra, y_ra)
```
avik-pal, line 126: This one technically changes the semantics of `fn` a little bit. `fn(x, y) === x` starts to hold which wouldn't be the case in julia

src/Compiler.jl
```diff
@@ -113,160 +113,183 @@ function create_result(
...
+
 const transform_passes::String = join(
+    [
+        "enzyme-hlo-generate-td{" * join(transform_passes_list, ';') * "}",
+        "transform-interpreter",
+        "enzyme-hlo-remove-transform",
+    ],
+    ",",
+)
+const no_nan_transform_passes::String = join(
```
wsmoses, line 289: At this point we should just make this a fn with kw args imo

wsmoses (approved):

PR #464, opened 2025-01-03
Fix offsetarrays support

wsmoses:


ext/ReactantOffsetArraysExt.jl
```diff
@@ -0,0 +1,13 @@
+module ReactantOffsetArraysExt
+
+using OffsetArrays: OffsetArray
+using Reactant: Reactant, MLIR, Ops, TracedRArray
+
+function Reactant.traced_type(
+    ::Type{<:OffsetArray{<:Any,N, T}}, seen::ST, ::Val{mode}, track_numbers
+) where {T, N,ST,mode}
+    T2 = Reactant.traced_type(T, seen, Val(mode), track_numbers)
```
wsmoses, line 9: @avik-pal so the reason this is necessary here is because the default conversion of ConcreteRArray breaks this where AFAICT OffsetArray is OffsetArray{ElType, Dim, ArrayType} which for a concreterarray is `OffsetArray{Float64, 2, ConcreRMatrix{Float64}}`. however the conversion of a TracedRArray is `OffsetArray{TracedRNumber{Float64}, 2, TracedRMatrix{Float64}}`.

The following code should always work, however.

PR #469, opened 2025-01-04
Update index.md

x66ccff:
add some tips related to `GC.gc()`

### DIFF
--- a/docs/src/introduction/index.md
+++ b/docs/src/introduction/index.md
@@ -61,7 +61,7 @@ f(input1, input2)
 
 When you encounter OOM (Out of Memory) errors, you can try to clear the cache by using `GC.gc()` between memory-intensive operations.
 
-```@example emptycache
+```julia
 using Reactant
 n = 5_0000_0000
 input1 = Reactant.ConcreteRArray(ones(n))
--- a/docs/src/introduction/index.md
+++ b/docs/src/introduction/index.md
@@ -161,7 +161,7 @@ Some type information was truncated. Use `show(err)` to see complete types.
 After using Julia's built-in `GC.gc()`:
 
 <details>
-  <summary>View Details</summary>
+  <summary>View Outputs</summary>
 
 ```bash
 [ Info: gc... 1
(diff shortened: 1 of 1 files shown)


docs/src/introduction/index.md
```diff
@@ -53,3 +53,29 @@ f = @compile sinsum_add(input1,input2)
 # one can now run the program
 f(input1, input2)
 ```
+
+
+## Tips
+
+### Empty Cache
+
+When you encounter OOM (Out of Memory) errors, you can try to clear the cache by using `GC.gc()` between memory-intensive operations.
```
wsmoses, line 62: perhaps using Julia's builtin `GC.gc()`. Also perhaps adding something like: "Note, this only will free memory which is not currently live. If, for example the result of f was stored in a vector, it would still be live and GC.gc() would not free it."

x66ccff: Got it!

docs/src/introduction/index.md
```diff
@@ -53,3 +53,31 @@ f = @compile sinsum_add(input1,input2)
 # one can now run the program
 f(input1, input2)
 ```
+
+
+## Tips
+
+### Empty Cache
+
```
wsmoses, line 61: sorry one more final nit, can you also paste the error message you got here (to describe what happens if you don't GC). That way if someone comes across the same error message they can search for and find this doc more easily

x66ccff: OK, will fix it tomorrow

docs/src/introduction/index.md
```diff
@@ -53,3 +53,130 @@ f = @compile sinsum_add(input1,input2)
...
+When you encounter OOM (Out of Memory) errors, you can try to clear the cache by using Julia's builtin `GC.gc()` between memory-intensive operations.
+
+Note：This only will free memory which is not currently live. If the result of compiled function was stored in a vector, it would still be live and `GC.gc()` would not free it.
```
mofeing, line 64:
```suggestion
!!! note
    This will only free memory which is not currently live. If the result of compiled function was stored in a vector, it will still be alive and `GC.gc()` won't free it.
```

wsmoses (approved):

x66ccff:
The docs don't render this correctly though. 
image

PR #490, opened 2025-01-07
Properly set within_autodiff (#442)

### LINKED ISSUE #442: `Enzyme.within_autodiff` returns true inside `compile`

jumerckx:
fixes #442
needs Enzyme.jl: https://github.com/EnzymeAD/Enzyme.jl/issues/2254

I had to introduce a new function `call_with_reactant_within_autodiff` to smuggle the `within_autodiff` in the `call_with_reactant_generator` through the `self` argument.
I also tried doing things through `set_reactant_abi` but that didn't seem to suffice (first commit).
Perhaps the extra code in `set_reactant_abi` isn't strictly necessary now so I can try removing it again if wanted.

### DIFF
--- a/src/Interpreter.jl
+++ b/src/Interpreter.jl
@@ -69,7 +69,7 @@ end
 @static if Enzyme.GPUCompiler.HAS_INTEGRATED_CACHE
     struct ReactantCacheToken end
 
-    function ReactantInterpreter(; world::UInt=Base.get_world_counter())
+    function ReactantInterpreter(; world::UInt=Base.get_world_counter(), within_autodiff=false)
         return Enzyme.Compiler.Interpreter.EnzymeInterpreter(
             ReactantCacheToken(),
             REACTANT_METHOD_TABLE,
@@ -78,7 +78,7 @@ end
             false,            #=reverse_rules=#
             false,            #=inactive_rules=#
             false,            #=broadcast_rewrite=#
-            true,             #=defer_within_autodiff=#
+            !within_autodiff, #=defer_within_autodiff=#
             set_reactant_abi,
         )
     end
@@ -86,7 +86,7 @@ else
     const REACTANT_CACHE = Enzyme.GPUCompiler.CodeCache()
 
     function ReactantInterpreter(;
-        world::UInt=Base.get_world_counter(), code_cache=REACTANT_CACHE
+        world::UInt=Base.get_world_counter(), code_cache=REACTANT_CACHE, within_autodiff=false
     )
         return Enzyme.Compiler.Interpreter.EnzymeInterpreter(
             REACTANT_CACHE,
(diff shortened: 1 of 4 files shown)


src/TracedUtils.jl
```diff
@@ -329,9 +330,19 @@ function make_mlir_fn(
         process_linear_args!(linear_args, fnbody, do_transpose, optimize_then_pad, inv_map)
 
         if isempty(kwargs)
-            Reactant.call_with_reactant(f, traced_args...)
+            if within_autodiff
```
wsmoses: I feel like a cleaner way to do this, is not to have a second interpreter. But instead we can create a new global ref set to false, and overlay within_autodiff to lookup that var, and during autodiff set that to true

wsmoses: I'm okay with this though, but if we were to do it in this form, I would probably change call_with_reactant to take a config type var, which stores the current state of whether in autodiff or not (and also we can extend to other things down the line as well)

avik-pal:
```julia
julia> using Enzyme

julia> function error_not_within_autodiff()
           !Enzyme.within_autodiff() && error("Not within autodiff")
           return nothing
       end
error_not_within_autodiff (generic function with 1 method)

julia> fwd_within_autodiff(Mode, RT) = Enzyme.autodiff(Mode, error_not_within_autodiff, RT)
fwd_within_autodiff (generic function with 1 method)

julia> error_not_within_autodiff()
ERROR: Not within autodiff
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] error_not_within_autodiff()
   @ Main ./REPL[5]:2
 [3] top-level scope
   @ REPL[7]:1
 [4] top-level scope
   @ none:1

julia> fwd_within_autodiff(Forward, Const)
()

julia> error_not_within_autodiff()

julia> Enzyme.within_autodiff()
false
```

I am extremely confused why is the 2nd call not throw an error here. Only happens if I call fwd_within_autodiff in between. cc @wsmoses this is in isolation from Reactant

wsmoses:
@vchuravy er wat

src/TracedUtils.jl
jumerckx: Ok, I got rid of the second interpreter like you described

src/Interpreter.jl
```diff
@@ -94,12 +94,12 @@ end
...
             REACTANT_METHOD_TABLE,
             world,
+            false,
             false,            #=forward_rules=#
             false,            #=reverse_rules=#
             false,            #=inactive_rules=#
             false,            #=broadcast_rewrite=#
-            false,            #=within_autodiff_rewrite=#
-            set_reactant_abi,
+            set_reactant_abi,            #=within_autodiff_rewrite=#
```
avik-pal: comment seems off

src/Enzyme.jl
```diff
@@ -426,7 +426,7 @@ function overload_autodiff(
         else
             ntuple(Val(width)) do i
                 Base.@_inline_meta
-                deepcopy(result)
+                return deepcopy(result)
```
avik-pal: revert these changes. possible originated from juliaformatter v2?

avik-pal:
@jumerckx  is this good to go?

jumerckx:
Yes! Will merge after CI has finished

PR #491, opened 2025-01-07
fix: missing scalar indexing check for setindex

avik-pal:


wsmoses (approved):

avik-pal:
needs some test fixes before merging

wsmoses (approved):

PR #494, opened 2025-01-07
fix: incorrect IR for traced RNGs

### LINKED ISSUE #492: Incorrect IR generated
See the IR: https://pastebin.com/wtd7brkX

```julia
/tmp/jl_CDw1qNavoJ.mlir:671:34: error: use of undeclared SSA value name
    %212 = "stablehlo.transpose"(%292) <{permutation = array<i64: 0>}> : (tensor<2xui64>) -> tensor<2xui64>
```

avik-pal:
fixes #492 
fixes https://github.com/LuxDL/Lux.jl/issues/1186

unfortunately I have no clue how to generate a minimal case for this

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -614,11 +614,13 @@ function codegen_flatten!(linear_args, result_stores)
     # resarg_code = Expr[]
 
     for (i, arg) in enumerate(linear_args)
-        paths = ((p for p in arg.paths if p[1] == :args)...,)
+        paths = ((p for p in Reactant.TracedUtils.get_paths(arg) if p[1] == :args)...,)
         path = if length(paths) == 1
             paths[1]
         else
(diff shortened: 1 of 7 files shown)


wsmoses (approved):
The path thing seems like an NFC so I’m confused how it matters but sure, it’s a clean up regardless and apparently it fixes things?

avik-pal:
I think the core issue here was `LuxCore.replicate` did a check with `hasmethod(copy, (x::TracedRNG,))` which for some reason failed even after I added `Base.copy` dispatches. I also forced LuxCore to use the copy instead of the generated function.

The other changes are mostly NFC, but since it was a cleanup I did not revert it

wsmoses:
Got it, yeah feel free to format/merge at your leisure

PR #504, opened 2025-01-09
linearize aliased kernel args

jumerckx:
Can't test end-to-end but the HLO contains the stores.
```julia
using Reactant, CUDA

A = ConcreteRArray(fill(1))

struct MyStruct{T, B}
    b::B
    A::T
end

function f!(tup)
    tup[1].A[] = 2
    return nothing
end

s = MyStruct(10, A)

f(s) = @cuda threads=1 f!((s, s))
@code_hlo optimize=false f(s);
```

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -507,7 +507,6 @@ Reactant.@reactant_overlay @noinline function (func::LLVMFunc{F,tt})(
             if p[1] !== kernelargsym
                 continue
             end
-                        
             # Get the allocation corresponding to which arg we're doing
             alloc = allocs[p[2]][1]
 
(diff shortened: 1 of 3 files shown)


ext/ReactantCUDAExt.jl
```diff
@@ -453,30 +533,20 @@ Reactant.@reactant_overlay @noinline function (func::LLVMFunc{F,tt})(
                     ),
```
jumerckx, line 533: Since I moved argidx++ out of the loop, I'm not sure whether this is still correct because I don't really know what's happening here.

wsmoses:
@jumerckx to make review easier, can you rebase this on main?

jumerckx:
> @jumerckx to make review easier, can you rebase this on main?

Done

wsmoses:
@jumerckx can you add that as a test, then let's merge!

jumerckx:
I'm afk til tomorrow but will add it then.

wsmoses (approved):

wsmoses:
LGTM but I think there's a conflict that needs to be fixed to merge

PR #506, opened 2025-01-10
feat: add rsqrt simplification

avik-pal:


src/Compiler.jl
```diff
@@ -264,6 +264,7 @@ function optimization_passes(; no_nan::Bool=false)
         "if_to_select<1>",
         "dynamic_update_slice_const_prop",
         "dynamic_gather_op_is_not_dynamic<16>",
+        "divide_sqrt_to_multiply_rsqrt<16>",
```
wsmoses, line 267: We also just got log and log1p simplifications too if you want to add at the same time

wsmoses (approved):
lgtm, but the neural network tests fail now (though could just be due to jll bump and not this)

avik-pal:
I will have to take a look here, since master is passing and the rsqrt opt directly affects the normalization tests that fail here

Pangoraw:
Could you also add `select_op_used_within_if` from https://github.com/EnzymeAD/Enzyme-JAX/pull/228 ?

wsmoses:
hm @Pangoraw maybe you can make a second PR for the non-rsqrt change (since apparently there's a bug in that that needs investigation?)

avik-pal:
the rsqrt derivative https://github.com/EnzymeAD/Enzyme-JAX/blob/4e28cbc8523af07a4f0f47972edd022bc9c91192/src/enzyme_ad/jax/Implementations/HLODerivatives.td#L1110 is missing a neg right?

src/Compiler.jl
```diff
@@ -196,6 +196,8 @@ function optimization_passes(; no_nan::Bool=false)
         "slice_pad<1>",
         "dot_reshape_dot<1>",
         "concat_const_prop<1>",
+        "log_const_prop<1>",
+        "log_plus_one_const_prop<1>",
```
Pangoraw, line 200: We don't really need those anymore? they have already been added in #510, or is this on purpose?
```suggestion
```

avik-pal:
we still need to wait for a JLL bump

wsmoses:
landed the jll bump, just merged here, let's see if it passes

Issue #511, opened 2025-01-11
Concatenation of scalar and ConcreteRArray gives a Vector


jaeminoh:
Hi folks,

I discovered that a concatenation of a scalar and a ConcreteRArray does not result in a ConcreteRArray.

```julia

using Reactant

const  xdev = reactant_device()

a = 1.0
b = ones(2) |> xdev

cat(a, b, dims=1) # 3-element Vector{Float64}, not ConcreteRArray{Float64, 1}
```

If we change the order (`cat(b, a, dims=1)`), then it is `ConcreteRArray{Float64, 1}`.

This could be related to the issue mentioned in the link: https://github.com/JuliaGPU/CUDA.jl/issues/1162, where the type of a variable is determined by its first argument.

What’s your plan for this? In my opinion, staying on ConcreteRArray would be more convenient.

wsmoses:
Seems like an easy enough fix. Mind basically writing a PR with this for an RArray? https://github.com/JuliaGPU/GPUArrays.jl/pull/379/files

PR #514, opened 2025-01-12
Make v and hcat with numbers work.

jaeminoh:
Concatenating a number with an RArray does not result in an RArray.

This PR addresses this issue, based on https://github.com/JuliaGPU/GPUArrays.jl/pull/379/files.

The GPU backend encounters a scalar index error when the `@allowscalar` is not used, while the CPU backend functions correctly without it.

(Fixes https://github.com/EnzymeAD/Reactant.jl/issues/511)

### DIFF
--- a/src/stdlibs/Base.jl
+++ b/src/stdlibs/Base.jl
@@ -1,4 +1,4 @@
-@inline Base.vcat(a::Number, b::RArray) =
+@inline Base.vcat(a::Number, b::Union{AnyConcreteRArray, AnyTracedRArray}) =
     @allowscalar(vcat(fill!(similar(b, typeof(a), (1, size(b)[2:end]...)), a), b))
-@inline Base.hcat(a::Number, b::RArray) =
+@inline Base.hcat(a::Number, b::Union{AnyConcreteRArray, AnyTracedRArray}) =
     @allowscalar(hcat(fill!(similar(b, typeof(a), (size(b)[1:end-1]..., 1)), a), b))
--- a/src/stdlibs/Base.jl
+++ b/src/stdlibs/Base.jl
@@ -1,0 +1,4 @@
+@inline Base.vcat(a::Number, b::RArray) =
+    @allowscalar(vcat(fill!(similar(b, typeof(a), (1, size(b)[2:end]...)), a), b))
+@inline Base.hcat(a::Number, b::RArray) =
+    @allowscalar(hcat(fill!(similar(b, typeof(a), (size(b)[1:end-1]..., 1)), a), b))
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -368,6 +368,17 @@ end
         @test y == test_typed_hvncat(x)
         @test eltype(y) === Int
     end
+
+    @testset "Number and RArray" begin
+        a = 1.0
+        b = Reactant.to_rarray(ones(3))
+        c = Reactant.to_rarray(ones(1, 3))
+        @test size(vcat(a, b)) == (4,)
+        @test size(hcat(a, b')) == (1, 4)
+        @test size(hcat(a, c)) == (1, 4)
+        @test size(vcat(a, c')) == (4, 1)
+    end
+
 end
 
 @testset "repeat" begin
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -369,28 +369,32 @@ end
         @test eltype(y) === Int
     end
 
-    @testset "Number and RArray" begin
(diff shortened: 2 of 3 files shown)


mofeing:
thanks for the PR!

just one thing: instead of creating a new file, Base things related to `ConcreteRArray` should go in "ConcreteRArray.jl"

wsmoses:
so intentionally if this applies to both ConcreteRArrays, and TracedRArrays (by virtue of using the base class), I think it's fine in a separate file.

However, we should probably add tests for both then (e.g. inside of a jit and just on a concreterarray)

src/stdlibs/Base.jl
```diff
@@ -0,0 +1,4 @@
+@inline Base.vcat(a::Number, b::RArray) =
```
wsmoses, line 1: Maybe make this Union{AnyConcreteRArray, AnyTracedRArray} (since this will then cover the reshpaed / etc version.

@avik-pal do we have a type for this already?

mofeing:
Ah right. I didn't see it was for `RArray` and I thought it was just for `ConcreteRArray` 🤦. Then alright.

src/stdlibs/Base.jl
avik-pal: no, we should just do an `const AnyRArray = ...`

test/basic.jl
```diff
@@ -368,6 +368,17 @@ end
         @test y == test_typed_hvncat(x)
         @test eltype(y) === Int
     end
+
+    @testset "Number and RArray" begin
+        a = 1.0
+        b = Reactant.to_rarray(ones(3))
+        c = Reactant.to_rarray(ones(1, 3))
+        @test size(vcat(a, b)) == (4,)
```
wsmoses, line 376: Can you test that the return type is as expected ?

also can you do a test with @jit vcat(a,b) etc

src/stdlibs/Base.jl
jaeminoh: I’ve modified the code to use `Union{AnyConcreteRArray, AnyTracedRArray}` instead of `RArray` and included a basic unit test. Could you review it once again?

test/basic.jl
jaeminoh: Here is the test result.
```julia
Test Summary:     | Pass  Total   Time
Number and RArray |    8      8  56.8s
```

test/basic.jl
```diff
@@ -368,6 +368,30 @@ end
         @test y == test_typed_hvncat(x)
         @test eltype(y) === Int
     end
+
+    @testset "Number and RArray" begin
+        a = 1.0
+        b = Reactant.to_rarray(ones(3))
```
wsmoses, line 374: In retrospect probably we should have this be distinct values like [2.,3.,4.] and that way we can confirm the new cat computes the correct result

test/basic.jl
```diff
@@ -369,28 +369,32 @@ end
         @test eltype(y) === Int
     end
 
-    @testset "Number and RArray" begin
-        a = 1.0
-        b = Reactant.to_rarray(ones(3))
-        c = Reactant.to_rarray(ones(1, 3))
-        
+    @testset "Number and RArray" for a in [randn(Float32), randn(Float64)]
```
wsmoses, line 372: Using fixed numbers like 2.0 etc here I think would be preferable for the sake of reproducible tests

test/basic.jl
jaeminoh: I guess the primary goal is to ensure the commutativity (loosely speaking) between `Reactant.to_rarray` and concatenation. So I've modified the testset a bit. Could you check one more time 😀?

```julia
Test Summary:     | Pass  Total   Time
Number and RArray |    8      8  56.9s
Test Summary:     | Pass  Total  Time
Number and RArray |    8      8  0.3s
```

Btw, `ConcrreteRArray(hcat(1.0, [2.0, 3.0]'))` did not work (no matching method).

test/basic.jl
jaeminoh: Is it okay to use `randn` for the body?

wsmoses: if it's just 3 numbers I'd just write some constants

jaeminoh: Test result:

```julia
Test Summary:     | Pass  Total   Time
Number and RArray |    8      8  55.9s
Test Summary:     | Pass  Total  Time
Number and RArray |    8      8  0.2s
```

wsmoses (approved):

PR #517, opened 2025-01-12
XLA Allocator stats

Pangoraw:


deps/ReactantExtra/API.cpp
```diff
@@ -325,6 +325,74 @@ extern "C" PjRtDevice *ClientGetAddressableDevice(PjRtClient *client,
...
+}
+extern "C" int64_t PjRtDeviceGetPeakBytesReserved(PjRtDevice *device) {
+  auto stats = device->GetAllocatorStats();
+  if (!stats.ok())
+    return std::numeric_limits<int64_t>::min();
+  return stats->peak_bytes_reserved;
+}
+extern "C" int64_t PjRtDeviceGetBytesReservableLimit(PjRtDevice *device) {
+  auto stats = device->GetAllocatorStats();
+  if (!stats.ok())
```
wsmoses, line 372: Returning here is invalid since it expects an int return. Maybe throw an error (see myvalueorthrow)

deps/ReactantExtra/API.cpp
```diff
@@ -325,6 +325,74 @@ extern "C" PjRtDevice *ClientGetAddressableDevice(PjRtClient *client,
...
+    return std::numeric_limits<int64_t>::min();
+  return stats->pool_bytes.value_or(std::numeric_limits<int64_t>::min());
+}
+extern "C" int64_t PjRtDeviceGetPeakPoolBytes(PjRtDevice *device) {
+  auto stats = device->GetAllocatorStats();
+  if (!stats.ok())
+    return std::numeric_limits<int64_t>::min();
+  return stats->peak_pool_bytes.value_or(std::numeric_limits<int64_t>::min());
+}
+
```
glou-nes, line 395: I think it's possible to replace this in order to limit GetAllocatorStats calls:
```c++
extern "C" tsl::AllocatorStats getAllocatorStats(PjRtDevice *Device) {
  auto stats = Device->GetAllocatorStats();
  return stats.value();
}
```
Optional are represented with a 8 bytes discriminant. So a Tuple can be used here for instance. 
```julia
struct AllocatorStats
    num_allocs::Int64
    bytes_in_use::Int64
    peak_bytes_in_use::Int64
    largest_alloc_size::Int64
    bytes_limit::Tuple{Int64,Int64}
    bytes_reserved::Int64
    peak_bytes_reserved::Int64
    bytes_reservable_limit::Tuple{Int64,Int64}
    largest_free_block_bytes::Int64
    pool_bytes::Tuple{Int64,Int64}
    peak_pool_bytes::Tuple{Int64,Int64}
end

@ccall mlir_c.getAllocatorStats(device::XLA.Device)::getAllocatorStats
```

wsmoses: I’m okay either way, but if you want to go this route, have it return the result via a pointer in the first arg and have the function return void (to avoid ABI issues with struct returning functions)

deps/ReactantExtra/API.cpp
Pangoraw: I made it throw 👍

deps/ReactantExtra/API.cpp
Pangoraw: How should the `Tuple{Int,Int}` be interpreted for `std::optional<int64_t>` ?

glou-nes: The first one is the value, the second is the discriminant:
```
get_value(t::Tuple{Int,Int}) = t[2] % 2 == 1 ? t : nothing
```

wsmoses: I'm slightly concerned with directly doing this (relying on the ABI of std optional), could we instead make our own struct which explicitly contains the ints (which we unwrap in c++)

mofeing: for this kind of things, i use a `std::tuple`. you could use a `std::tuple<bool,int>` where the first value says if the value is valid or not.

Pangoraw: I think using `typemin(Int64)` as a sentinel like I did here should be ok and C friendly ?

mofeing: @glou-nes reading it more closely, don't rely on passing that `tsl::AllocatorStats` directly. you have no guarantee it will work.

you can instead:
1. save it into a pointer (e.g. using a copy/move constructor with new) and pass the pointer
2. create your own C-struct and pass that, without C++ objects in it

Pangoraw: I updated the call to make a single call to `GetAllocatorStats`.

wsmoses:
@Pangoraw jll is now landed if you want to try btw

Pangoraw:
```julia
julia> stats = Reactant.XLA.allocatorstats()
       stats.num_allocs, stats.bytes_in_use
(0, 0)

julia> x = Reactant.to_rarray(randn(Float32, 1000));

julia> stats = Reactant.XLA.allocatorstats()
       stats.num_allocs, stats.bytes_in_use
(1, 4096)

julia> x = nothing; GC.gc(true)

julia> stats = Reactant.XLA.allocatorstats()
       stats.num_allocs, stats.bytes_in_use
(1, 0)
```

mofeing:
is this working on CPU-only too?

Pangoraw:
unfortunately, no

```
ERROR: UNIMPLEMENTED: GetAllocatorStats is not supported

PR #523, opened 2025-01-13
Automatic function call insertion

jumerckx:
on top of: https://github.com/EnzymeAD/Reactant.jl/pull/366, currently has an error:
```julia
Reactant.DEBUG_INTERP[] = true
Reactant.TOGGLE_TRACECALLS[] = true

@noinline _call1(a, b) = a
function call1(a, b)
    x = _call1(a, b)
    y = _call1(a, b)
    return _call1(x, y)
end

a = rand(2, 3)
b = rand(2, 3)
a_ra = Reactant.to_rarray(a)
b_ra = Reactant.to_rarray(b)

@compile(call1(a_ra, b_ra))
```

```
ERROR: StackOverflowError:
Stacktrace:
     [1] make_typealias(x::Type)
       @ Base ./show.jl:644
     [2] show_typealias(io::IOBuffer, x::Type)
       @ Base ./show.jl:805
     [3] _show_type(io::IOBuffer, x::Type)
       @ Base ./show.jl:970
     [4] show(io::IOBuffer, x::Type)
       @ Base ./show.jl:965
     [5] show_typeparams(io::IOBuffer, env::Core.SimpleVector, orig::Core.SimpleVector, wheres::Vector{TypeVar})
       @ Base ./show.jl:722
     [6] show_datatype(io::IOBuffer, x::DataType, wheres::Vector{TypeVar})
       @ Base ./show.jl:1181
     [7] show_datatype
       @ ./show.jl:1089 [inlined]
     [8] _show_type(io::IOBuffer, x::Type)
       @ Base ./show.jl:973
     [9] show(io::IOBuffer, x::Type)
       @ Base ./show.jl:965
    [10] _show_default(io::IOBuffer, x::Any)
       @ Base ./show.jl:486
    [11] show_default
       @ ./show.jl:482 [inlined]
    [12] show
       @ ./show.jl:477 [inlined]
    [13] print(io::IOBuffer, x::Base.Generator{Vector{Pair{Any, Any}}, Reactant.var"#1#3"})
       @ Base ./strings/io.jl:35
    [14] print_to_string(xs::Base.Generator{Vector{Pair{Any, Any}}, Reactant.var"#1#3"})
       @ Base ./strings/io.jl:148
    [15] string
       @ ./strings/io.jl:189 [inlined]
    [16] safe_print
       @ ~/Reactant.jl/src/utils.jl:448 [inlined]
    [17] OrderedDict
       @ ~/.julia/packages/OrderedCollections/5e4BO/src/ordered_dict.jl:27 [inlined]
    [18] call_with_reactant(::Type{OrderedCollections.OrderedDict{…}}, ::Base.Generator{Vector{…}, Reactant.var"#1#3"})
       @ Reactant ~/Reactant.jl/src/utils.jl:0
    [19] OrderedIdDict
       @ ~/Reactant.jl/src/OrderedIdDict.jl:8 [inlined]
    [20] OrderedIdDict
       @ ~/Reactant.jl/src/OrderedIdDict.jl:16 [inlined]
    [21] OrderedIdDict
       @ ~/Reactant.jl/src/OrderedIdDict.jl:15 [inlined]
    [22] traced_call
       @ ~/Reactant.jl/src/ControlFlow.jl:134 [inlined]
    [23] traced_call(none::typeof(memoryref), none::Tuple{Memory{UInt64}})
       @ Reactant ./<missing>:0
    [24] GenericMemory
       @ ./boot.jl:514 [inlined]
    [25] Array
       @ ./boot.jl:578 [inlined]
    [26] getindex
       @ ./array.jl:400 [inlined]
    [27] OrderedIdDict
       @ ~/Reactant.jl/src/OrderedIdDict.jl:16 [inlined]
    [28] OrderedIdDict
       @ ~/Reactant.jl/src/OrderedIdDict.jl:15 [inlined]
    [29] traced_call
       @ ~/Reactant.jl/src/ControlFlow.jl:134 [inlined]
    [30] call_with_reactant(::typeof(ReactantCore.traced_call), ::typeof(memoryref), ::Memory{UInt64})
       @ Reactant ~/Reactant.jl/src/utils.jl:0
    [31] traced_call_with_reactant(f::Function, args::Memory{UInt64})
       @ Reactant ~/Reactant.jl/src/utils.jl:19
    [32] Array
       @ ./boot.jl:579 [inlined]
    [33] Array
       @ ./boot.jl:601 [inlined]
    [34] OrderedDict
       @ ~/.julia/packages/OrderedCollections/5e4BO/src/ordered_dict.jl:23 [inlined]
    [35] OrderedCollections.OrderedDict{UInt64, Any}()
       @ Reactant ./<missing>:0
    [36] GenericMemory
       @ ./boot.jl:516 [inlined]
    [37] Array
       @ ./boot.jl:578 [inlined]
    [38] Array
       @ ./boot.jl:591 [inlined]
    [39] zeros
       @ ./array.jl:589 [inlined]
    [40] zeros
       @ ./array.jl:585 [inlined]
    [41] OrderedDict
       @ ~/.julia/packages/OrderedCollections/5e4BO/src/ordered_dict.jl:23 [inlined]
    [42] call_with_reactant(redub_arguments#232::Type{OrderedCollections.OrderedDict{UInt64, Any}})
       @ Reactant ~/Reactant.jl/src/utils.jl:0
    [43] traced_call
       @ ~/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:397 [inlined]
    [44] traced_call(none::Type{OrderedCollections.OrderedDict{UInt64, Any}}, none::Tuple{})
       @ Reactant ./<missing>:0
    [45] traced_call
       @ ~/Reactant.jl/lib/ReactantCore/src/ReactantCore.jl:397 [inlined]
    [46] call_with_reactant(::typeof(ReactantCore.traced_call), ::Type{OrderedCollections.OrderedDict{UInt64, Any}})
       @ Reactant ~/Reactant.jl/src/utils.jl:0
    [47] traced_call_with_reactant(::Type)
       @ Reactant ~/Reactant.jl/src/utils.jl:0
    [48] OrderedDict
       @ ~/.julia/packages/OrderedCollections/5e4BO/src/ordered_dict.jl:27 [inlined]
    [49] OrderedCollections.OrderedDict{UInt64, Any}(none::Base.Generator{Vector{Pair{Any, Any}}, Reactant.var"#1#3"})
       @ Reactant ./<missing>:0
--- the above 33 lines are repeated 2284 more times ---
...
```
With debug printing eventually just repeating:
```
...
"fn arg[1] traced_call"
"fn arg[2] memoryref"
"fn arg[3] (UInt64[],)"
"fn arg[1] memoryref"
"fn arg[2] Pair{Any, Any}[]"
"fn arg[1] OrderedCollections.OrderedDict{UInt64, Any}"
"fn arg[2] Base.Generator{Vector{Pair{Any, Any}}, Reactant.var\"#1#3\"}(Reactant.var\"#1#3\"(), Pair{Any, Any}[])"
"fn arg[1] traced_call"
"fn arg[2] OrderedCollections.OrderedDict{UInt64, Any}"
"fn arg[3] ()"
"fn arg[1] OrderedCollections.OrderedDict{UInt64, Any}"
...
```

As if there's a cycle in the callgraph?


src/utils.jl
```diff
@@ -257,11 +274,12 @@ function rewrite_inst(inst, ir, interp, RT, guaranteed_error)
                 end
             end
         elseif Base.invokelatest(should_rewrite_ft, ft)
+            new_f = (!allow_tracing || ft <: typeof(ReactantCore.traced_call)) ? call_with_reactant : traced_call_with_reactant
```
wsmoses, line 277: So I see what's happening here, but I feel like we should do this a bit differently.

Presently you're rewriting all calls to be traced_call_with_reactant.

What if, instead, we modified the codeinfo and/or opaque closure within call_with_reactant. That way we don't have an extra level of indirection (that may cause issues)

wsmoses: We already have the methodinstance/codeinfo on the inside (and argtypes themselves) so we could even do the equivalent of make_Tracer into a compile-time recusion (aka generate the equivalent of

```
key = (arg1.x.y, arg2.z, arg3[4], ...)
```

during the generated function, so then the (relatively expensive) make_tracer isn't called every function call (which is already quite expensive).

wsmoses: Also therefore if a function just is foo(TracedArray, ) we literally don't even need to do a key check!

wsmoses: yeah it depends on the type, maybe we can talk about it tomorrow/over the weekend

wsmoses:
first seems cleaner to me

jumerckx:
When tracing through this function:
https://github.com/EnzymeAD/Reactant.jl/blob/467559dacc1f01db63771bdfaf58eb5a4714a312/src/TracedRArray.jl#L772C1-L773C1
Somewhere in the callgraph of `collect(TracedUtils.get_mlir_data.(X))` is a call to something similar to:
```jl
function f(X::TracedRArray)
	return (X.get_mlir_data, ) # MLIR values, not TracedRArrays -> Considered "not part of" the Reactant tracing
end
```
i.e., a function that takes a tracedrarray and returns a tuple of MLIR values.

The automatic call insertion currently maps all `TracedRArray` arguments onto mlir function arguments. Similarly, `TracedRArray` return values are updated once returned to the caller scope, to contain the result of the `func.call` instead of the mlir value from the callee scope.

Here, however, this leads to buggy code:
`X` is mapped onto `block arg 0` in the callee, but when the MLIR value is returned, it is not a TracedRArray, but an MLIR value—part of Julia land.
As a result, the return result is not mapped back on the `func.call` result but instead keeps referring to block arg from the callee scope.

A straightforward solution would be to have an option to disable Reactant tracing as discussed in the call today. This requires the user code to be changed, though, so I'm wondering if there's a possibility to handle this case automatically?
Curious to hear what your thoughts are on this, @wsmoses.

wsmoses:
I'd consider there to be two classes of functions:
* User and helper code that uses tracedrarray's/etc
* Ops, and other manual code that takes mlir values

We should handle the first and not the second (and I'm surprised we didn't mark that existing function as noinline and not to be re-interpreter'd on the inside)

jumerckx:
> and I'm surprised we didn't mark that existing function as noinline and not to be re-interpreter'd on the inside

Right, so the solution is just to move part of that function to an Op?

wsmoses:
yeah like Ops.concat or something

jumerckx:
This pr has unearthed a problem that's also present for the manual call tracing (`@trace f(args...)`).
We now cache the result of `make_mlir_fn`, including `traced_result` which contains TracedRArrays where we update the `mlir_data` inside.
The problem is that the objectid's of these refurbished TracedRArrays don't match those of the call arguments even though they contain the same data.
This leads to the "same" arg appearing multiple times in `linear_results` from the caller.
This test where we verify pointer equality between argument and result fails because of this:
https://github.com/EnzymeAD/Reactant.jl/blob/ae6e60c923cce2b3932b4a90574d75817058e235/test/indexing.jl#L312C1-L313C1

My current thinking to solve this is to detect these duplicate args and somehow replace the offending result with its up-to-date arg version in the `traced_result`. I'm still working out how this replacement can happen, though.

Any thoughts on this are appreciated! (@wsmoses)

jumerckx:
> EDIT: a full fancy replacement might not be necessary, just making sure that the duplicate arg isn't added and the paths in the refurbished result are correctly set might suffice...

This seems to have worked for that test at least! Will have to check if I broke anything else in the process...

jumerckx:
I will probably retire this pr and open a new one, but I'm running into an issue that I'd love some feedback on.
When code contains a call to e.g. `push!`, I'm wondering how function caching can work.
```jl
A = []
x::TracedRArray = ...

push!(A, x)

use(A) # <-- A is not mutated if the call wasn't explicitly executed?
```

`push!` mutates the outer Julia array, not some TracedRArray directly. So somehow the function cache needs to store how to mutate this Julia object to have the same effect as an actual call to `push!`. I have not yet found a way to do this, do you happen to have thoughts on this @wsmoses?

PR #535, opened 2025-01-15
feat: implement sort

avik-pal:
upstream issues: https://github.com/EnzymeAD/Enzyme-JAX/issues/239


src/Ops.jl
```diff
@@ -1557,4 +1557,119 @@ use [`MLIR.Dialects.stablehlo.dynamic_slice`](@ref) instead.
...
+
+    MLIR.IR.activate!(fnbody)
+    result = try
+        for (i, arg) in enumerate(linear_args)
+            raw_arg = MLIR.IR.argument(fnbody, i)
+            Reactant.TracedUtils.set_mlir_data!(arg, raw_arg)
+        end
+        # XXX: call_with_reactant is not working here?
+        # ERROR: type Nothing has no field stmts
+        # res = Reactant.call_with_reactant(f, linear_args...)
```
wsmoses, line 1631: we really do need to do call with reactant here (otherwise f is called outside the absint)

avik-pal: it was the missing noinline that caused the trouble. fixed it!

Project.toml
```diff
@@ -9,6 +9,7 @@ CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
+Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
```
avik-pal, line 12: I tried not using functors [`9ef1fff` (#535)](https://github.com/EnzymeAD/Reactant.jl/pull/535/commits/9ef1fffa4e1d480a76b3e315ccd26a943dd14942) but it gets quite messy and I couldn't figure out how to reconstruct arguments with corrected batch dims

avik-pal:
This is going to be a fun rebase :sweat:

wsmoses:
oof yeah, the "lets speed up to_rarray" just got in (and cut down time from minutes to seconds for big cases I tested on)

Issue #548, opened 2025-01-17
MethodError for `setindex!` in `sum!`


glwagner:
While trying to time-step an `Oceananigans.Models.HydrostaticFreeSurfaceModel`:

```
ERROR: LoadError: MethodError: setindex!(::SubArray{Reactant.TracedRNumber{Float64}, 3, Reactant.TracedRArray{Float64, 3}, Tuple{UnitRange{Int64}, UnitRange{Int64}, UnitRange{Int64}}, false}, ::Reactant.TracedRNumber{Float64}, ::Int64, ::Int64, ::Int64) is ambiguous.

Candidates:
  setindex!(V::SubArray{T, N}, x, I::Vararg{Int64, N}) where {T, N}
    @ Base subarray.jl:328
[...]

PR #549, opened 2025-01-17
add int override

wsmoses:
Fixes https://github.com/EnzymeAD/Reactant.jl/issues/548


wsmoses:
hm @avik-pal it looks like the overlay broke some of the existing tests =/

mofeing:
this should fix #550 too

PR #560, opened 2025-01-18
[CI] Format generated files twice to work around JuliaFormatter bug

giordano:
While we evaluate a move to Runic, this should resolve the problem that the formatting job and bindings generation job conflict with each other due to an [upstream bug in JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl/issues/897).


wsmoses (approved):
lol re double format.

PR #575, opened 2025-01-19
profiler: Add option to generate perfetto url

### LINKED ISSUE #572: [Profiling] Add option to autogenerate perfetto and/or tensorboad url
Image

Profiling apis are super useful (thanks @Pangoraw). JaX seems to have an option for autocreating the perfetto url (see https://github.com/jax-ml/jax/blob/b0117366686ab084d38ad2657d9a2ae3a581ca7e/jax/_src/profiler.py#L180 ) which would be really nice here too

Pangoraw:
Closes #572.


giordano:
https://github.com/EnzymeAD/Reactant.jl/actions/runs/12852483302/job/35834519904?pr=575#step:7:167
```
[ Info: CheckDocument: running document checks.
│ 
│     Reactant.Profiler.with_profiler :: Tuple{Any, String}
│ 
 ```

File: src/Reactant.jl [hunk 1]
```diff
@@ -249,6 +249,8 @@ include("Tracing.jl")
 include("Compiler.jl")
 include("Profiler.jl")
 
+const with_profiler = Profiler.with_profiler
```
mofeing, line 252: just curious, why not `import .Profiler: with_profiler`?

giordano: I assume it's more making Reactant.with_profiler autocomplete in the REPL, simply importing it wouldn't do it

mofeing: 😮 that looks like a autocompleter bug right?

giordano: Not sure, there's technically no symbol `Reactant.with_profiler` unless you define it like here, there would be only `Reactant.Profiler.with_profiler` imported in the `Reactant` scope.

wsmoses: yeah I feel like the import would be cleaner?

wsmoses (approved):
LGTM but maybe change to import per below

src/Reactant.jl
mofeing: but `names(Reactant; imported=true)` should return `with_profiler` too if imported

Pangoraw: Yeah this is for auto complete. I think using or import did not have auto complete unfortunately.

PR #582, opened 2025-01-20
Profiler annotations & tutorial

### LINKED ISSUE #581: Profiling Tutorial
@Pangoraw if you could add some breif examples on how to use / export to web services / tensorboard (and say generated things like in https://github.com/EnzymeAD/Enzyme-JAX/issues/256) that would be really useful for folks

Pangoraw:
Closes #581.


docs/src/tutorials/images/perfetto.png
mofeing, line 1: are these screenshots by us or are they taken from somewhere?

Pangoraw: by us, those are from the example

wsmoses (approved):
Fantastic! Maybe also add a line about the allocator stats in the tutorial too?

wsmoses (approved):

PR #587, opened 2025-01-20
feat: multi GPU support

avik-pal:
```julia
julia> y_ra = ConcreteRArray(rand(10, 4); idx=5)
10×4 ConcreteRArray{Float64, 2}:
 0.331353  0.63357    0.241808    0.378997
 0.217468  0.125116   0.00109667  0.140413
 0.849274  0.468709   0.21194     0.383364
 0.875864  0.721931   0.315834    0.786151
 0.368286  0.314255   0.120778    0.819903
 0.662467  0.402153   0.996147    0.432376
 0.92015   0.0209533  0.175266    0.0502965
 0.355591  0.622008   0.71689     0.395723
 0.960684  0.708643   0.915696    0.623402
 0.389559  0.883376   0.383502    0.33884

julia> x_ra = ConcreteRArray(rand(10, 4); idx=2)
10×4 ConcreteRArray{Float64, 2}:
 0.411537  0.820146   0.72804    0.485311
 0.894575  0.990613   0.0608764  0.599667
 0.11137   0.306586   0.934523   0.661384
 0.51723   0.761962   0.360649   0.876964
 0.203228  0.563193   0.483365   0.691988
 0.995663  0.880207   0.0743423  0.988633
 0.372236  0.711898   0.423323   0.509736
 0.353636  0.0939618  0.365339   0.435227
 0.954899  0.757799   0.701784   0.451779
 0.885934  0.46312    0.246529   0.539774

julia> @jit .+(x_ra, y_ra)
2025-01-20 18:42:00.317253: I external/xla/xla/service/llvm_ir/llvm_command_line_options.cc:50] XLA (re)initializing LLVM with options fingerprint: 11040064543056471292
E0120 18:42:01.086668  124306 pjrt_stream_executor_client.cc:3045] Execution of replica 0 failed: INVALID_ARGUMENT: Buffer passed to Execute() as argument 0 to replica 0 is on device cuda:2, but replica is assigned to device cuda:0.
ERROR: INVALID_ARGUMENT: Buffer passed to Execute() as argument 0 to replica 0 is on device cuda:2, but replica is assigned to device cuda:0.

Stacktrace:
 [1] reactant_err(msg::Cstring)
   @ Reactant.XLA ~/research/reactant/Reactant.jl/src/XLA.jl:131
 [2] macro expansion
   @ ~/research/reactant/Reactant.jl/src/XLA.jl:464 [inlined]
 [3] ExecutableCall
   @ ~/research/reactant/Reactant.jl/src/XLA.jl:441 [inlined]
 [4] macro expansion
   @ ~/research/reactant/Reactant.jl/src/Compiler.jl:925 [inlined]
 [5] (::Reactant.Compiler.Thunk{Base.Broadcast.BroadcastFunction{…}, Symbol("##Base.Broadcast.BroadcastFunction(+)_reactant#234"), Tuple{…}, true})(::ConcreteRArray{Float64, 2}, ::ConcreteRArray{Float64, 2})
   @ Reactant.Compiler ~/research/reactant/Reactant.jl/src/Compiler.jl:1060
 [6] top-level scope
   @ ~/research/reactant/Reactant.jl/src/Compiler.jl:616
 [7] top-level scope
   @ none:1
```


src/XLA.jl
```diff
@@ -54,22 +54,48 @@ function CPUClient(asynchronous=false, node_id=0, num_nodes=1; checkcount=true)
...
-        platform,
-        refstr,
-    )
+function GPUClient(
+    node_id=0, num_nodes=1, platform="gpu"; num_allowed_devices::Union{Cint,Nothing}=nothing
+)
+    allowed_devices = try
+        if haskey(ENV, "CUDA_VISIBLE_DEVICES")
+            return unique!(parse.(Cint, strip.((split(ENV["CUDA_VISIBLE_DEVICES"])))))
+        else
```
wsmoses, line 63: I'm not certain here, but I feel like we're not supposed to make a separate client per device, but we create a separate device like

https://github.com/EnzymeAD/Reactant.jl/blob/cca721dad2218cee45112e61a971f77a44e57e15/src/XLA.jl#L57
and/or https://github.com/EnzymeAD/Reactant.jl/blob/cca721dad2218cee45112e61a971f77a44e57e15/src/XLA.jl#L91C7-L91C25
from https://github.com/EnzymeAD/Reactant.jl/blob/cca721dad2218cee45112e61a971f77a44e57e15/src/ConcreteRArray.jl#L4
eventually hitting https://github.com/EnzymeAD/Reactant.jl/blob/cca721dad2218cee45112e61a971f77a44e57e15/src/XLA.jl#L342

Not certain though

avik-pal: The return was accidental. We are creating a single client. But we need to refactor the concrete array dispatches a bit to make sure we end up on the same device as the original array

wsmoses: oh sorry yeah misread quickly

wsmoses: re device count / etc we can always call something like https://github.com/EnzymeAD/Reactant.jl/blob/63d407a3962c46aaf8c011adf95f3aa83ee1e309/deps/ReactantExtra/API.cpp#L466 but also we should explicitly use something like https://github.com/openxla/xla/blob/03260eb1b76d0d41c67ff98d4f1c3bd5af9dbadc/xla/python/pjrt_ifrt/pjrt_client.h#L220 (idk if that adheres to CUDA_VISIBLE_DEVICES, but I would hope so)

wsmoses: actually we already have it https://github.com/EnzymeAD/Reactant.jl/blob/63d407a3962c46aaf8c011adf95f3aa83ee1e309/deps/ReactantExtra/API.cpp#L364 and/or https://github.com/EnzymeAD/Reactant.jl/blob/63d407a3962c46aaf8c011adf95f3aa83ee1e309/deps/ReactantExtra/API.cpp#L368

avik-pal: > actually we already have it

we need the device_count to create the PJRT client first

> idk if that adheres to CUDA_VISIBLE_DEVICES, but I would hope so

This I figured out is directly handled by cuda, so we just need to provide a counter for this.

wsmoses: hm I don't think so. If we pass in null (as now), MakeGPUClient sets allowed_devices to the null optional : https://github.com/EnzymeAD/Reactant.jl/blob/63d407a3962c46aaf8c011adf95f3aa83ee1e309/deps/ReactantExtra/API.cpp#L285

I presume this means get all devices (permitted by cuda)?

avik-pal: > re device count / etc we can always call something like

This would work. But I can't seem to build the cuda jll locally :sweat:

Issue #380, opened 2024-12-15
[BUG]: Segfault with large amount of data


x66ccff:
```julia

using Reactant
using ProgressMeter

# 定义一个 Julia 函数
function julia_func(x, y)
    return sin.(x) .+ y
end

in_dim = 1000000
n = 10000  # 总迭代次数

# 测量编译时间
println("Compiling the function:")
compilation_time = @elapsed begin
    global compiled = @compile julia_func(
        Reactant.ConcreteRArray(ones(in_dim)),
        Reactant.ConcreteRArray(ones(in_dim))
    )
end

println("Compilation time: $(compilation_time) seconds")

random_x = Vector{Reactant.ConcreteRArray{Float64, 1}}(undef, n)
random_y = Vector{Reactant.ConcreteRArray{Float64, 1}}(undef, n)

# 生成随机数并显示进度
println("Generating random numbers:")
p_rand = Progress(n; dt=0.5, barglyphs=BarGlyphs("[=> ]"), barlen=50, color=:yellow)

for i in 1:n
    random_x[i] = Reactant.ConcreteRArray(randn(in_dim))
    random_y[i] = Reactant.ConcreteRArray(randn(in_dim))
    next!(p_rand)
end
finish!(p_rand)

# 进行计算并显示进度
println("Processing calculations:")
p_calc = Progress(n; dt=0.5, barglyphs=BarGlyphs("[=> ]"), barlen=50, color=:cyan)

for i = 1:n
    result = compiled(random_x[i], random_y[i])

    next!(p_calc)
end

finish!(p_calc)

```

```
2024-12-15 23:55:31.119553: I external/xla/xla/tsl/framework/bfc_allocator.cc:1109] InUse at 7dea46a8f400 of size 8000000 next 2347
2024-12-15 23:55:31.119571: I external/xla/xla/tsl/framework/bfc_allocator.cc:1109] InUse at 7dea47230600 of size 8000000 next 2348
2024-12-15 23:55:31.119591: I external/xla/xla/tsl/framework/bfc_allocator.cc:1109] InUse at 7dea479d1800 of size 8000000 next 2349
2024-12-15 23:55:31.119608: I external/xla/xla/tsl/framework/bfc_allocator.cc:1109] InUse at 7dea48172a00 of size 8000000 next 2350
[...]

PR #589, opened 2025-01-21
feat: expose gpu memory allocation options

avik-pal:
introduces `XLA_REACTANT_GPU_MEM_FRACTION` and `XLA_REACTANT_GPU_PREALLOCATE`


mofeing:
Nice! It partially solves #380

deps/build_local.jl
```diff
@@ -80,10 +80,24 @@ end
...
+bazel_cmd = try
+    run(Cmd(`bazelisk --version`))
+    "bazelisk"
+catch
+    try
+        run(Cmd(`bazel --version`))
+        "bazel"
+    catch
+        error("Could not find `bazel` or `bazelisk` in PATH!")
+    end
```
giordano, line 92: Use `Sys.which` instead?

PR #611, opened 2025-01-24
feat: forward more base ops to chlo

avik-pal:


wsmoses (approved):
lgtm but maybe add a range test?

PR #604, opened 2025-01-24
Fix dense elements attribute in `Enzyme.autodiff` #593

mofeing:
Fixes #593

(I still have a problem with rev-rule that i wasn't having before but that's probably on my side)

PR #613, opened 2025-01-25
Bump Enzyme-JAX

mofeing:
We need this commit in order to test #604


wsmoses:
dup of https://github.com/EnzymeAD/Reactant.jl/pull/612

PR #622, opened 2025-01-26
feat: build the shardy dialect

avik-pal:


wsmoses (approved):
Lg but also FYI we'll have to make sure we register it to our mlircontext to use it

wsmoses:
Also FYI we shouldnt need the http archive since it should already be a dep of xla itself

deps/ReactantExtra/API.cpp
```diff
@@ -701,6 +702,8 @@ extern "C" void InitializeRegistryAndPasses(MlirDialectRegistry creg) {
   mlir::registerNVVMDialectImport(registry);
   mlir::LLVM::registerInlinerInterface(registry);
 
+  mlir::sdy::registerAllDialects(registry);
```
avik-pal, line 705: @wsmoses this should be enough right

wsmoses: Double check if there’s a loadDialect call somewhere too otherwise lgtm

wsmoses (approved):

PR #624, opened 2025-01-26
chore: missing upstream optimization passes

avik-pal:


src/Compiler.jl
```diff
@@ -297,6 +297,8 @@ function optimization_passes(; no_nan::Bool=false, sroa::Bool=false)
         "replace_neg_add_with_subtract",
         "log_const_prop<1>",
         "log_plus_one_const_prop<1>",
+        "while_deadresult<1>",
```
avik-pal, line 300: https://github.com/EnzymeAD/Enzyme-JAX/commit/47db9dda28d30c618c3bf59d4406d2def0162f24#diff-3ea4c99365467e031be723fefaf5b74ab16bc9b011c2b718bfeb29a38db1fdb6R801

why doesn't it pick this up?

wsmoses:
That pass was asserting so I reverted the commit for the time being, cc @ftynse

src/Compiler.jl
```diff
@@ -297,6 +297,8 @@ function optimization_passes(; no_nan::Bool=false, sroa::Bool=false)
...
+        "while_simplify<1>",
     ]
```
avik-pal, line 302:
```suggestion
        "binop_const_simplify",
        "transpose_broadcast_in_dim_to_broadcast_in_dim",
        "not_select_simplify",
        "common_compare_expression_rewrite",
        "normalize_loop",
        "compare_select_simplify",
        "while_simplify<1>",
    ]
```

src/Compiler.jl
```diff
@@ -298,6 +298,13 @@ function optimization_passes(; no_nan::Bool=false, sroa::Bool=false)
         "replace_neg_add_with_subtract",
         "log_const_prop<1>",
         "log_plus_one_const_prop<1>",
+        "binop_const_simplify",
+        "transpose_broadcast_in_dim_to_broadcast_in_dim",
+        "not_select_simplify",
+        "common_compare_expression_rewrite",
+        "normalize_loop",
```
avik-pal, line 305:
```suggestion
```

Issue #614, opened 2025-01-25
InvalidIRError in CUDA integration tests on non-CUDA machines


giordano:
After the upgrade to Julia v1.11.3 we're getting
```
Square Kernel: Error During Test at /home/runner/work/Reactant.jl/Reactant.jl/test/integration/cuda.jl:26
  Got exception outside of a @test
  InvalidIRError: compiling MethodInstance for Main.var"##CUDA#245".square_kernel!(::ReactantCUDAExt.CuTracedArray{Int64, 1, 1, (64,)}, ::ReactantCUDAExt.CuTracedArray{Int64, 1, 1, (64,)}) resulted in invalid LLVM IR
  Reason: unsupported dynamic function invocation (call to throw_boundserror() @ CUDA ~/.julia/packages/CUDA/1kIOw/src/device/quirks.jl:51)
  Stacktrace:
   [1] #throw_boundserror
     @ ~/.julia/packages/CUDA/1kIOw/src/device/quirks.jl:53
   [2] checkbounds
     @ ./abstractarray.jl:699
   [3] #arrayref
     @ ~/work/Reactant.jl/Reactant.jl/ext/ReactantCUDAExt.jl:62
   [4] getindex
     @ ~/work/Reactant.jl/Reactant.jl/ext/ReactantCUDAExt.jl:151
   [5] square_kernel!
     @ ~/work/Reactant.jl/Reactant.jl/test/integration/cuda.jl:8
```
See e.g. [Julia 1.11 - integration - ubuntu-20.04 - x64 - packaged libReactant - assertions=false - push](https://github.com/EnzymeAD/Reactant.jl/actions/runs/12958842182/job/36151452487#step:9:888)

I'm going to skip these tests to keep CI green, but this issue still needs to be resolved.

wsmoses:
@vchuravy @maleadt I'm not sure why but I think CUDA.jl on 1.11 is hitting an error (in its error handling).

In particular Julia is deciding to emit a type unstable invoke to CUDA.throw_boundserror() [no arguments], instead of inlining it and enabling the subsequent LLVM code to be converted to an exit asm instruction.

```julia
# Using Julia 1.11

using CUDA, Reactant, Test, Revise, GPUCompiler, Cthulhu

using Reactant
using Test
using CUDA

function square_kernel!(x, y)
    i = threadIdx().x
    x[i] *= y[i]
    sync_threads()
    return nothing
end

# basic squaring on GPU
function square!(x, y)
[...]

PR #626, opened 2025-01-26
Add bound optimizations

wsmoses:


giordano:
Can this help with #614?

wsmoses:
not that's unrelated. The issue you link to is that for some reason cuda.jl thinks the kernel has a type unstable call (and throws an error about)

PR #632, opened 2025-01-26
make `similar` return empty tensors.

jumerckx:
(Instead of creating a dummy traced array containing no mlir_data)
https://github.com/EnzymeAD/Reactant.jl/pull/523#issuecomment-2606965530

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -373,9 +373,8 @@ Base.collect(x::TracedRArray) = copy(x) # XXX: Is this correct?
 
 Base.copy(A::TracedRArray{T,N}) where {T,N} = TracedRArray{T,N}((), A.mlir_data, size(A))
 
-# TODO is there a way to create an unitialized `tensor`? does it show an advantage? maybe `fill`?
 function Base.similar(::TracedRArray, ::Type{T}, dims::Dims{N}) where {T,N}
-    return Ops.constant(zeros(unwrapped_eltype(T), dims))
+    return Ops.fill(zero(unwrapped_eltype(T)), dims)
 end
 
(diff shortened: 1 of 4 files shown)


src/TracedRArray.jl
```diff
@@ -554,14 +554,14 @@ function Base.similar(
     ::Broadcasted{AbstractReactantArrayStyle{N}}, ::Type{T}, dims
 ) where {T<:Reactant.ReactantPrimitive,N}
     @assert N isa Int
-    return TracedRArray{T,length(dims)}((), nothing, map(length, dims))
+    return Ops.constant(zeros(unwrapped_eltype(T), dims))
```
wsmoses, line 557: This looks reasonable to me, but this forces us to create an array of size dims which might be expensive.

Since we know all values are going to be the same we can use the MLIR splat attribute (see https://github.com/EnzymeAD/Reactant.jl/blob/59b1856a39d803fec1ecc94c906987f0bee3d0e1/src/mlir/libMLIR_h.jl#L3962 )

Can we instead make an Ops.fill(value, dims) and use that here instead (and elsewhere)

mofeing: try out the MLIR splat attribute before merging. We've had segfaults with the C-API for creating MLIR splat attributes.

wsmoses: I mean either way we can make an Ops.fill and use it here (and elsewhere), then figure out the best impl inside

jumerckx: This is now added. I looked for other spots to use fill instead of full materialization but didn't find anything yet.

wsmoses (approved):

PR #638, opened 2025-01-27
Add IR dumping

wsmoses:


wsmoses:
@giordano the CUDA jll is broken


ERROR: LoadError: InitError: could not load library "/root/.cache/julia-buildkite-plugin/depots/0190ad31-dfb1-4a4c-ac5b-02d05277ba6c/artifacts/379d34d3179cb57eb475af426a954f27f35d6f4c/lib/libReactantExtra.so"
--
  | /root/.cache/julia-buildkite-plugin/depots/0190ad31-dfb1-4a4c-ac5b-02d05277ba6c/artifacts/379d34d3179cb57eb475af426a954f27f35d6f4c/lib/libReactantExtra.so: undefined symbol: descriptor_table_xla_2fstream_5fexecutor_2fcuda_2fcuda_5fcompute_5fcapability_2eproto

giordano:
Sigh. Maybe we had to use the cuda capability proto stuff always and not just for non-CUDA? This should have been caught by the linker though, but sometimes on Linux unresolved symbols are ignored, there should be a flag to disallow them completely (that's what macOS and Windows do by default).

giordano:
Basically, this may be resolved by #647

PR #640, opened 2025-01-27
[ReactantExtra] Bump XLA version

giordano:
With previous commit I was running into [fun compilation errors on macOS](https://github.com/JuliaPackaging/Yggdrasil/pull/10313#issuecomment-2616420694).


giordano:
Hold on, this now breaks building on aarch64-linux :sob:

giordano:
Ok, this seems to be working for a couple of platforms I built locally, let's see how it goes in Yggdrasil.

PR #645, opened 2025-01-28
feat: add dispatch for KA get_backend

avik-pal:


vchuravy:
Looks about right for me

ext/ReactantCUDAExt.jl
```diff
@@ -3,10 +3,13 @@ module ReactantCUDAExt
 using CUDA
 using Reactant: Reactant, TracedRArray, AnyTracedRArray, MLIR, TracedRNumber
 using ReactantCore: @trace
+using KernelAbstractions: KernelAbstractions
 using Libdl
 
 using Adapt
 
+KernelAbstractions.get_backend(::AnyTracedRArray) = CUDABackend()
```
wsmoses, line 11: should we move this to a separate ReactantKAext, or does it require both CUDA and KA here?

avik-pal:
Cuda backend is from cuda.jl

wsmoses:
@avik-pal looks good to me, but some tests fail rn. Can you only run KA tests when cuda is available? (once we add a cpu backend to Enzyme-JaX for kernel lowering we can re-enable)

ext/ReactantCUDAExt.jl
```diff
@@ -3,10 +3,13 @@ module ReactantCUDAExt
...
+
```
wsmoses, line 12: @avik-pal can we also add this for concreterarray?

PR #646, opened 2025-01-28
Use `xla/stream_executor/cuda:cuda_compute_capability_proto_cc_impl` only on non CUDA

giordano:


giordano:
Tested locally, should be good to go.

PR #647, opened 2025-01-29
CPU backend

wsmoses:
Naturally, requires jll update.

But enables us to run GPU kernels on CPU (without a cuda device)


wsmoses:
cc @giordano

src/Compiler.jl
```diff
@@ -975,12 +978,24 @@ function compile_xla(f, args; client=nothing, optimize=true, no_nan=false, devic
...
+        backend = XLA.ClientGetPlatformName(client)
+    else
+        backend = XLA.ClientGetPlatformName(XLA.default_backend[])
+    end
+    if backend == "CUDA"
+        backend = "GPU"
+    elseif backend == "CPU"
+        backend = "cpu"
+    end
+    @show backend
```
giordano, line 991: Leftover debug print?

src/Compiler.jl
```diff
@@ -456,7 +456,10 @@ function compile_mlir!(mod, f, args; optimize::Union{Bool,Symbol}=true, no_nan::
     if isdefined(Reactant_jll, :ptxas_path)
         toolkit = Reactant_jll.ptxas_path[1:(end - length("/bin/ptxas"))]
     end
-    if DEBUG_KERNEL[]
+
+    if backend == "cpu"
+        kern = "lower-kernel{openmp=false backend=cpu},symbol-dce"
```
giordano, line 461: Out of curiosity, now that I see the reference to openmp, what's used for multithreading on the CPU?

giordano (approved):
Sounds cool!

src/Compiler.jl
wsmoses: "who even needs to have threads"

(it should be able to be turned on and we'll use openmp but I want to first run end to end before debugging symbol resolution for kmpc_for)

src/Compiler.jl
wsmoses: or rather "curreently being debugged print" :P

src/Compiler.jl
wsmoses: see https://github.com/EnzymeAD/Enzyme-JAX/pull/290 for context

src/Compiler.jl
```diff
@@ -429,7 +429,7 @@ end
 
 const DEBUG_KERNEL = Ref{Bool}(false)
 
-function compile_mlir!(mod, f, args; optimize::Union{Bool,Symbol}=true, no_nan::Bool=false)
+function compile_mlir!(mod, f, args; optimize::Union{Bool,Symbol}=true, no_nan::Bool=false, backend="gpu")
```
avik-pal, line 432: `backend` should be queried from the current client?

avik-pal: the part in code_hlo should probably be done in this function

wsmoses: The problem is this function doesn’t have a client so it has to be done above

giordano:
On macOS:
```console
$ nm /Users/mose/.julia/artifacts/f3d3555e5fc223023219f21123197e90fa208f2d/lib/libReactantExtra.dylib | grep ClientGetPlatformName
000000000000472c t _ClientGetPlatformName
```
Lowercase `t` means that the symbol isn't visible (it's local, non-external).

wsmoses:
@avik-pal was that fn not marked exported?

Project.toml
```diff
@@ -72,7 +72,7 @@ PythonCall = "0.9"
 Random = "1.10"
 Random123 = "1.7"
 ReactantCore = "0.1.4"
-Reactant_jll = "0.0.52"
+Reactant_jll = "0.0.56"
```
avik-pal, line 75:
```suggestion
Reactant_jll = "0.0.58"
```

PR #667, opened 2025-01-31
KA ext

wsmoses:
Needs https://github.com/JuliaGPU/GPUCompiler.jl/pull/660 for mac at least


Project.toml
```diff
@@ -54,8 +53,8 @@ ReactantYaoBlocksExt = "YaoBlocks"
 AbstractFFTs = "1.5"
 Adapt = "4.1"
 ArrayInterface = "7.17.1"
-CEnum = "0.5"
-CUDA = "5.5"
+CEnum = "0.4, 0.5"
+CUDA = "5.5, 5.6"
```
giordano, line 57: the `5.6` is redundant, `5.5` already means `[5.5, 6)`, the union with the set represented by `5.6` (`[5.6, 6)`) is just the same.

Issue #526, opened 2025-01-13
Precompiling `Reactant` errors in GPU-related symbol


Todorbsc:
CC @mofeing 

I got this error while trying to update the version of `Reactant.jl` to a branch:
```julia
pkg> add Reactant#ss/split-invoke-call-rewrite
    Updating git-repo `https://github.com/EnzymeAD/Reactant.jl.git`
   Resolving package versions...
    Updating `<project-path>/Project.toml`
  [3c362404] ~ Reactant v0.2.18 `https://github.com/EnzymeAD/Reactant.jl.git#main` ⇒ v0.2.18 `https://github.com/EnzymeAD/Reactant.jl.git#ss/split-invoke-call-rewrite`
    Updating `<project-path>/Manifest.toml`
  [3c362404] ~ Reactant v0.2.18 `https://github.com/EnzymeAD/Reactant.jl.git#main` ⇒ v0.2.18 `https://github.com/EnzymeAD/Reactant.jl.git#ss/split-invoke-call-rewrite`
Precompiling project...
  ✗ Reactant
  ✗ Reactant → ReactantStatisticsExt
  ✗ Reactant → ReactantPythonCallExt
  ✗ Reactant → ReactantSpecialFunctionsExt
  ✗ Reactant → ReactantAbstractFFTsExt
  ✗ Reactant → ReactantOffsetArraysExt
  ✗ Reactant → ReactantYaoBlocksExt
  ✗ Tenet → TenetReactantExt
  0 dependencies successfully precompiled in 32 seconds. 382 already precompiled.
  8 dependencies errored.
  For a report of the errors see `julia> err`. To retry use `pkg> precompile`

julia> err
PkgPrecompileError: The following 8 direct dependencies failed to precompile:

ReactantSpecialFunctionsExt 

Failed to precompile ReactantSpecialFunctionsExt [5981d66b-43ae-5ef9-a655-fb5bed72485b] to "~/.julia/compiled/v1.11/ReactantSpecialFunctionsExt/jl_29LleC".
ERROR: LoadError: InitError: could not load library "~/.julia/artifacts/bf5c832056d192639a75739fc1016591bc1c64c5/lib/libReactantExtra.so"
~/.julia/artifacts/bf5c832056d192639a75739fc1016591bc1c64c5/lib/libReactantExtra.so: undefined symbol: cuStreamBeginCaptureToGraph
Stacktrace:
[...]

PR #668, opened 2025-01-31
[docs] Add information about configuration on GPU and TPU systems

### LINKED ISSUE #584: Document solution of #526
If the linker find a `libcuda.so` in a machine that **doesn't** have NVIDIA GPUs, Reactant will falsely believe that a CUDA is available and it will download the wrong artifact. The solution to #526 is to add a preference for `Reactant_jll` to use the `gpu=none` artifact variant.

This should be documented in the docs, given that this situation can happen to many HPC cluster users.

cc @giordano @Todorbsc

giordano:
Fix #584


giordano:
You can see the rendered docs at https://enzymead.github.io/Reactant.jl/previews/PR668/configuration/, but I clearly did something wrong because this page doesn't show up anywhere in the navigation menus despite the fact I added it to `make.jl`.  CC @avik-pal.

avik-pal:
It also needs to be added to the mts file https://github.com/EnzymeAD/Reactant.jl/blob/main/docs/src/.vitepress/config.mts#L53

docs/src/configuration/index.md
```diff
@@ -0,0 +1,148 @@
...
+│  └ @ Reactant_jll ~/.julia/packages/Reactant_jll/daenT/.pkg/platform_augmentation.jl:108
+```
+
+Here you can see that on this system Reactant found the CUDA Driver at `/lib/x86_64-linux-gnu/libcuda.so.1` with version 12.2.0.
+
+#### Disabling CUDA support
+
+Reactant looks for the CUDA Driver library `libcuda` to determine whether the current system supports Nvidia GPUs.
+However in some cases this library may be actually present on the machine even though no GPU is actually attached to it, which would trick Reactant's installation process into believing a GPU is available.
+To force disabling GPU support, in the environment where you installed Reactant create a file called `LocalPreferences.toml` with the following content:
```
wsmoses, line 81: Can we add a sentence here that “normally this is not a problem as Reactant will detect that in spite of the cuda driver being present there are noGPUs and will default to the CPU backend”

avik-pal:
also we can remove https://github.com/EnzymeAD/Reactant.jl#installing-reactant-on-gpu-servers-without-internet and move that section here

wsmoses (commented):
Lgtm

docs/src/configuration/index.md
giordano: Is it not?  I believe people may still run into issues like #526, or not?

wsmoses: Yeah that issue is distinct where we failed to dlopen reactant because the Libcuda did not have the version we required/expected. This should be solved by the jll version of Cuda check

giordano:
@avik-pal I'm still missing something: in https://enzymead.github.io/Reactant.jl/previews/PR668/configuration/ I don't see the left sidebar, and there's no "previous page" at the bottom.

giordano:
Should I move the page to `introduction/configuration.md` instead of `configuration/index.md`?

avik-pal:
> Should I move the page to introduction/configuration.md instead of configuration/index.md?

yes

wsmoses (approved):

PR #673, opened 2025-02-01
Use `LLVMOpenMP_jll` to call OpenMP functions

giordano:


src/Compiler.jl
```diff
@@ -467,16 +464,16 @@ function compile_mlir!(
...
+        kern = "lower-kernel,canonicalize,lower-jit{debug=true cuResultHandlerPtr=$curesulthandler cuOptLevel=$(cuOptLevel[]) cubinFormat=$(cubinFormat[]) indexBitWidth=$(cuindexBitWidth[])  cubinChip=$(cubinChip[]) cubinFeatures=$(cubinFeatures()) run_init=true toolkitPath=$toolkit},symbol-dce"
     else
-        kern = "lower-kernel{cuOptLevel=$(cuOptLevel[]) indexBitWidth=$(cuindexBitWidth[]) cubinFormat=$(cubinFormat[]) cubinChip=$(cubinChip[]) cubinFeatures=$(cubinFeatures()) run_init=true toolkitPath=$toolkit cuLaunchKernelPtr=$(cuLaunch[]) cuModuleLoadDataPtr=$(cuModule[]) cuModuleGetFunctionPtr=$(cuFunc[])},symbol-dce"
+        kern = "lower-kernel,canonicalize,lower-jit{cuOptLevel=$(cuOptLevel[]) indexBitWidth=$(cuindexBitWidth[]) cubinFormat=$(cubinFormat[]) cubinChip=$(cubinChip[]) cubinFeatures=$(cubinFeatures()) run_init=true toolkitPath=$toolkit},symbol-dce"
```
giordano, line 476: In #684 I had to remove all the options `cuOptLevel` `indexBitWidth` `cubinFormat` `cubinChip` `cubinFeatures` `run_init` `toolkitPath` because I got errors about the fact they don't exist.

wsmoses: they got moved into options for the new lower-jit pass (when I forked lower-kernel into a kernel specific pass and a jit pass)

PR #674, opened 2025-02-01
[Test] Skip (again) CUDA integration tests when they are known to fail

giordano:
CUDA integration tests are still failing on non-CUDA systems in Julia v1.11 due to #614, but they were brought back in #667.  With this change we skip running those tests when we expect them to fail.

Note: as mentioned in https://github.com/EnzymeAD/Reactant.jl/issues/614#issuecomment-2629084527 these tests should finally be working again in Julia v1.11.4.

### DIFF
--- a/test/integration/cuda.jl
+++ b/test/integration/cuda.jl
@@ -18,9 +18,9 @@ function square!(x, y)
 end
 
 # https://github.com/EnzymeAD/Reactant.jl/issues/614
-const skip_non_cuda_tests = true
+const skip_non_cuda_tests = v"1.11.0" <= VERSION < v"1.11.4"
 
-@static if !Sys.isapple()
+@static if !Sys.isapple() || !skip_non_cuda_tests
     @testset "Square Kernel" begin
         oA = collect(1:1:64)
         A = Reactant.to_rarray(oA)
@@ -43,7 +43,7 @@ function sin!(x, y)
     return nothing
 end
 
-@static if !Sys.isapple()
+@static if !Sys.isapple() || !skip_non_cuda_tests
     @testset "Sin Kernel" begin
         oA = collect(Float64, 1:1:64)
         A = Reactant.to_rarray(oA)
@@ -67,7 +67,7 @@ function smul!(x)
     return nothing
 end
 
-@static if !Sys.isapple()
+@static if !Sys.isapple() || !skip_non_cuda_tests
     @testset "Constant Op Kernel" begin
         oA = collect(1:1:64)
         A = Reactant.to_rarray(oA)


wsmoses:
I think we should go for a different fix here and instead change the precompile reactantcudaext.jl to only run on 11.4+, which I believe should fix this while still running tests

wsmoses:
The Apple cuda tests are currently only disabled pending https://github.com/JuliaGPU/GPUCompiler.jl/pull/660

giordano:
> I think we should go for a different fix here and instead change the precompile reactantcudaext.jl to only run on 11.4+, which I believe should fix this while still running tests

OK, I can confirm

```diff
diff --git a/ext/ReactantCUDAExt.jl b/ext/ReactantCUDAExt.jl
index 391be0b..d8ccbec 100644
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -1,5 +1,7 @@
 module ReactantCUDAExt
 
+__precompile__(!(v"1.11.0" <= VERSION < v"1.11.4"))
+
 using CUDA
 using Reactant:
     Reactant, TracedRArray, AnyTracedRArray, AnyConcreteRArray, MLIR, TracedRNumber
```

would also make this work on Julia v1.11.3, the annoying thing is that it makes loading Reactant _a lot_ slower for everybody because the extension needs to be precompiled all times on load, and this seems to be quite slow:
```julia-repl
julia> using CUDA

julia> @time @eval using Reactant
AssertionError("Could not find registered platform with name: \"cuda\". Available platform names are: ")
Precompiling ReactantCUDAExt...
  ? Reactant → ReactantCUDAExt
[ Info: Precompiling ReactantCUDAExt [91356394-6239-50cb-9e6b-6f1f3f798cde] (cache misses: incompatible header (4), mismatched flags (2))
┌ Info: Skipping precompilation due to precompilable error. Importing ReactantCUDAExt [91356394-6239-50cb-9e6b-6f1f3f798cde].
└   exception = Error when precompiling module, potentially caused by a __precompile__(false) declaration in the module.
 32.698447 seconds (39.38 M allocations: 2.111 GiB, 4.15% gc time, 92.60% compilation time: 83% of which was recompilation)
```

I'm happy to make this change if you thing it's worth it, but I outlined the drawback!

wsmoses:
@giordano ah I didn’t mean that I meant this: https://github.com/EnzymeAD/Reactant.jl/blob/1e6037f35ed5a36d7486873dd6423a81fcc0e834/ext/ReactantCUDAExt.jl#L955

wsmoses:
Also incidentally we can now remove the aarch64 and is functional checks now that CPU codegen works (though obviously it will still require an unreleased Julia)

PR #675, opened 2025-02-01
[ReactantCUDAExt] Skip precompile load on Julia v1.11.3

giordano:
Implementing suggestion at https://github.com/EnzymeAD/Reactant.jl/pull/674#issuecomment-2629092644

Close #674.

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -952,12 +952,14 @@ function __init__()
     return nothing
 end
 
-@static if !Sys.isapple() && Sys.ARCH != :aarch64
+# In Julia v1.11.3 precompiling this module caches bad code:
+# <https://github.com/EnzymeAD/Reactant.jl/issues/614>.
+@static if !Sys.isapple()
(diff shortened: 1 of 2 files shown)


ext/ReactantCUDAExt.jl
```diff
@@ -952,7 +952,9 @@ function __init__()
     return nothing
 end
 
-@static if !Sys.isapple() && Sys.ARCH != :aarch64
+# In Julia v1.11.3 precompiling this module caches bad code:
+# <https://github.com/EnzymeAD/Reactant.jl/issues/614>.
+@static if !Sys.isapple() && Sys.ARCH != :aarch64 && VERSION != v"1.11.3"
```
wsmoses, line 957: Remove the asrch64 check and change the precompilation_supported line below to be precompilation supported and VERSION >= 11.4

giordano: Done!  Note that the issue in #614 is only in v1.11.3, it was due to the backport of https://github.com/JuliaLang/julia/pull/56749, which was only in that release.

wsmoses (approved):

Issue #677, opened 2025-02-01
Segmentation faults on aarch64-linux starting from introduction of extension of KernelAbstractions


giordano:
Starting from #667 we have seen lots of segmentation faults on aarch64:

[Julia 1.10 - integration - ubuntu-24.04-arm - aarch64 - packaged libReactant - assertions=false - push](https://github.com/EnzymeAD/Reactant.jl/actions/runs/13085357311/job/36532476725#step:9:423)
```
Failed to precompile ReactantStatisticsExt [963ed91e-491b-54ce-bb4b-249dcb1ed2bb] to "/home/runner/.julia/compiled/v1.10/ReactantStatisticsExt/jl_B0YRz3".
2025-02-01 04:56:31.905782: I external/xla/xla/service/llvm_ir/llvm_command_line_options.cc:51] XLA (re)initializing LLVM with options fingerprint: 11962807958986418783

[4252] signal (11.1): Segmentation fault
in expression starting at /home/runner/work/Reactant.jl/Reactant.jl/src/Precompile.jl:60
last_fde at /workspace/srcdir/gcc-13.2.0/libgcc/unwind-dw2-fde.h:174 [inlined]
classify_object_over_fdes at /workspace/srcdir/gcc-13.2.0/libgcc/unwind-dw2-fde.c:727
init_object at /workspace/srcdir/gcc-13.2.0/libgcc/unwind-dw2-fde.c:888 [inlined]
_Unwind_Find_registered_FDE at /workspace/srcdir/gcc-13.2.0/libgcc/unwind-dw2-fde.c:1210 [inlined]
_Unwind_Find_FDE at /workspace/srcdir/gcc-13.2.0/libgcc/unwind-dw2-fde-dip.c:541
uw_frame_state_for at /workspace/srcdir/gcc-13.2.0/libgcc/unwind-dw2.c:1005
_Unwind_Backtrace at /workspace/srcdir/gcc-13.2.0/libgcc/unwind.inc:303
__backtrace at /lib/aarch64-linux-gnu/libc.so.6 (unknown line)
tsl::CurrentStackTrace[abi:cxx11]() at /home/runner/.julia/artifacts/e08cc4d821f228b8f487acd163930546f0b6ff17/lib/libReactantExtra.so (unknown line)
[...]

PR #679, opened 2025-02-02
add tmate option to CI

jumerckx:


.github/workflows/CI.yml
```diff
@@ -96,6 +129,15 @@ jobs:
...
       - uses: actions/checkout@v4
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        if: |
+          github.event_name == 'workflow_dispatch' &&
+          inputs.debug_enabled == true &&
+          (inputs.debug_os == 'any' || inputs.debug_os == matrix.os) &&
+          (inputs.debug_arch == 'any' || inputs.debug_arch == matrix.arch)
+        with:
+          detached: ${{ inputs.debug_detached }}
```
mofeing, line 140: i guess you need to put this after the `setup-julia` step? if not, you won't have the julia binary (i believe you may have it but it's due to the github docker image and you don't have control over the version)

jumerckx: With `detached=true` I think the Julia setup (and all the rest) should still run after the tmate session is created?
If you're looking at this, do you have any idea to actually trigger the workflow manually 😅

mofeing: you need to merge it to see the manual option, but @giordano should know better

giordano: I had already set this up: https://github.com/EnzymeAD/Reactant.jl/issues/677#issuecomment-2629205537

Issue #663, opened 2025-01-30
We aren't actually using the ptxas and libdevice.bc shipped in the CUDA packages


giordano:
On a system with an Nvidia GPU supported by Reactant, by exporting the (undocumented) environment variable `TF_CPP_MAX_VLOG_LEVEL=3` before starting julia, one can see
```
julia> using Reactant

julia> @jit Reactant.to_rarray(randn(Float32, 1_000)) .+ 1;
[...]
2025-01-30 17:43:26.583495: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for ./cuda_sdk_lib/bin/ptxas
2025-01-30 17:43:26.583499: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /data/mose/repo/spack/bin/ptxas
2025-01-30 17:43:26.583503: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /home/mose/.juliaup/bin/ptxas
2025-01-30 17:43:26.583506: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /home/mose/.local/bin/ptxas
2025-01-30 17:43:26.583508: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /home/mose/bin/ptxas
2025-01-30 17:43:26.583511: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /usr/share/Modules/bin/ptxas
2025-01-30 17:43:26.583514: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /usr/condabin/ptxas
2025-01-30 17:43:26.583516: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /usr/local/bin/ptxas
2025-01-30 17:43:26.583519: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /usr/bin/ptxas
2025-01-30 17:43:26.583521: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /usr/local/sbin/ptxas
2025-01-30 17:43:26.583523: I external/xla/xla/stream_executor/cuda/subprocess_compilation.cc:177] Looking for /usr/sbin/ptxas
[...]

PR #683, opened 2025-02-03
[ReactantExtra] Add argument to `ClientCompile` to pass CUDA data dir

giordano:
Not tested yet, but based on https://github.com/EnzymeAD/Reactant.jl/issues/663#issuecomment-2629665961 it should work.  Will require a change in the compiler on the Julia side in a follow up PR.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -597,7 +597,8 @@ extern "C" MlirModule ConvertLLVMStrToMLIR(const char *lmod, MlirContext cctx) {
 extern "C" xla::PjRtLoadedExecutable *ClientCompile(PjRtClient *client,
                                                     MlirModule cmod,
                                                     int *global_ordinals,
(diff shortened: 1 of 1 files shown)


deps/ReactantExtra/API.cpp
```diff
@@ -597,7 +597,8 @@ extern "C" MlirModule ConvertLLVMStrToMLIR(const char *lmod, MlirContext cctx) {
 extern "C" xla::PjRtLoadedExecutable *ClientCompile(PjRtClient *client,
                                                     MlirModule cmod,
                                                     int *global_ordinals,
-                                                    int num_global_ordinals) {
+                                                    int num_global_ordinals,
+                                                    std::string xla_gpu_cuda_data_dir) {
```
wsmoses, line 601: can you make this a const char*?

giordano:
```suggestion
                                                    const char* xla_gpu_cuda_data_dir) {
```

PR #689, opened 2025-02-05
[deps] Some improvements to the `build_local.jl` script

giordano:
Remove unused packages from the environment, add more arguments to the script.


wsmoses (approved):

giordano:
I don't understand these lines: https://github.com/EnzymeAD/Reactant.jl/blob/21f76607862dae51a2cb8a5cf3c5f27c27a7761a/deps/build_local.jl#L55-L66 I don't see `LLVM_full_jll used anywhere, and the lines below "Build!" don't seem to do anything concrete.  It feels like this is a leftover of an old experiment?

giordano:
Tested successfully by building CUDA bindings on an x86_64-linux machine with GCC 11, the component that wouldn't work are now disabled automatically.

PR #692, opened 2025-02-05
feat: API changes for multi-device execution [ReactantExtra JLL changes]

avik-pal:


avik-pal:
still debugging 1 segfault from the finalizer

giordano (approved):
Looks OK for as much as I can tell (which is limited 😅).

> still debugging 1 segfault from the finalizer

Was that solved? I don't see new commits after that message.

avik-pal:
Actually wait no not solved. I thought it got fixed... But it does segfault on finalizer. Let me investigate under gdb and hope it reveals more info

deps/ReactantExtra/API.cpp
```diff
@@ -722,41 +770,40 @@ extern "C" void XLAExecute(xla::PjRtLoadedExecutable *exec, int num_args,
...
+    int64_t mesh_id = mesh_ids[device_idx];
     for (int result_idx = 0; result_idx < num_results; ++result_idx) {
-      int flat_index = device_idx * num_results + result_idx;
-      op_results[flat_index] = results[device_idx][result_idx].release();
+      int flat_index = mesh_id * num_results + result_idx;
+      op_results[flat_index] = results[mesh_id][result_idx].release();
+      if (returned_futures.has_value()) {
+        future_results[flat_index] = new FutureType((*returned_futures)[mesh_id]);
+      }
     }
```
avik-pal, line 807: Any obvious mistake here that I might be making that causes the segfault?

deps/ReactantExtra/API.cpp
```diff
@@ -160,6 +161,13 @@ extern "C" MlirAttribute mlirComplexAttrDoubleGetChecked(MlirLocation loc,
 // TODO mlirComplexAttrGetnValue
 // TODO extern "C" MlirTypeID mlirComplexAttrGetTypeID(void) { return
 // wrap(complex::NumberAttr::getTypeID()); }
+
+extern "C" void mlirFuncSetResultAttr(MlirOperation op, intptr_t pos,
+                                      MlirStringRef name, MlirAttribute attr) {
```
wsmoses, line 166: We should upstream this into LLVM proper (tag me and/or @ftynse as reviewiers, or basically any of the kernel team).

That said, we should be more general here and have this apply to FunctionOpInterface instead of FuncOp, if possible

deps/ReactantExtra/API.cpp
```diff
@@ -694,23 +732,33 @@ extern "C" void XLAExecuteSharded(xla::PjRtLoadedExecutable *exec, int num_args,
...
 
   // Distribute arguments across devices
-  for (int device_idx = 0; device_idx < num_devices; ++device_idx) {
-    argument_handles[device_idx].reserve(num_args);
+  for (int device_idx = 0; device_idx < num_mesh_ids; ++device_idx) {
+    int64_t mesh_id = mesh_ids[device_idx];
+
+    // Validate mesh_id
+    if (mesh_id < 0 || mesh_id >= num_mesh_ids) {
+        llvm::errs() << "Error: Invalid mesh_id " << mesh_id << " at device_idx " << device_idx << "\n";
```
wsmoses, line 753: We can make this a JuliaThrow instead like https://github.com/EnzymeAD/Reactant.jl/blob/d842b337edd3d6508bd48e41f9e787671e87ede5/deps/ReactantExtra/API.cpp#L138

wsmoses:
Not a segfault source but I do see a memory error.

For example: here we allocate memory https://github.com/EnzymeAD/Reactant.jl/blob/d842b337edd3d6508bd48e41f9e787671e87ede5/deps/ReactantExtra/API.cpp#L394
and when using it
https://github.com/EnzymeAD/Reactant.jl/blob/d842b337edd3d6508bd48e41f9e787671e87ede5/src/XLA.jl#L625
I don't think we ever free it (we need to do unsafe_string, and then free the memory).

deps/ReactantExtra/API.cpp
avik-pal: Let me force a no futures case for now to check if that fixes it

avik-pal: that's not it

deps/ReactantExtra/API.cpp
avik-pal: https://github.com/llvm/llvm-project/pull/125972

avik-pal:
We can go ahead with this JLL. Turns out I don't understand basic arithmetic :cry:

Issue #696, opened 2025-02-06
Integration with SpeedyWeather.jl


giordano:
```julia-repl
julia> using Reactant, SpeedyWeather
AssertionError("Could not find registered platform with name: \"cuda\". Available platform names are: ")

julia> spectral_grid = SpectralGrid(trunc=31, nlayers=8);

julia> model = PrimitiveWetModel(spectral_grid);

julia> simulation = initialize!(model);

julia> rsimulation = Reactant.to_rarray(simulation);
ERROR: NoFieldMatchError:
Cannot convert type PrognosticVariables{NF, ArrayType, NSTEPS, SpectralVariable2D, SpectralVariable3D, GridVariable2D, ParticleVector}, best attempt PrognosticVariables{NF, ArrayType, NSTEPS, SpectralVariable2D, SpectralVariable3D, GridVariable2D, ParticleVector} failed.
This could be because the type does not capture the fieldtypes that should be converted in its type parameters.
idx=1 Derived: Int64 Existing: Int64
idx=2 Derived: Int64 Existing: Int64
idx=3 Derived: Int64 Existing: Int64
idx=4 Derived: Int64 Existing: Int64
idx=5 Derived: NTuple{NSTEPS, SpectralVariable3D} Existing: NTuple{NSTEPS, SpectralVariable3D}
idx=6 Derived: NTuple{NSTEPS, SpectralVariable3D} Existing: NTuple{NSTEPS, SpectralVariable3D}
idx=7 Derived: NTuple{NSTEPS, SpectralVariable3D} Existing: NTuple{NSTEPS, SpectralVariable3D}
idx=8 Derived: NTuple{NSTEPS, SpectralVariable3D} Existing: NTuple{NSTEPS, SpectralVariable3D}
idx=9 Derived: NTuple{NSTEPS, SpectralVariable2D} Existing: NTuple{NSTEPS, SpectralVariable2D}
idx=10 Derived: SpectralVariable2D Existing: SpectralVariable2D
idx=11 Derived: PrognosticVariablesOcean{NF, ArrayType, GridVariable2D} Existing: PrognosticVariablesOcean{NF, ArrayType, GridVariable2D}
idx=12 Derived: PrognosticVariablesLand{NF, ArrayType, GridVariable2D} Existing: PrognosticVariablesLand{NF, ArrayType, GridVariable2D}
idx=13 Derived: ParticleVector Existing: ParticleVector
[...]

PR #698, opened 2025-02-06
Ref ptr fix

wsmoses:
Resolves https://github.com/EnzymeAD/Reactant.jl/issues/696

### DIFF
--- a/src/Tracing.jl
+++ b/src/Tracing.jl
@@ -380,7 +380,7 @@ for P in (Ptr, Base.RefValue)
         seen,
         mode::TraceMode,
         @nospecialize(track_numbers::Type)
-    ) where T
+    ) where {T}
         return $P{traced_type_inner(PT.parameters[1], seen, mode, track_numbers)}
     end
 end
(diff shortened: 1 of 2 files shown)


src/Tracing.jl
```diff
@@ -366,14 +366,31 @@ end
...
+        @nospecialize(PT::Type{$P{T}}),
+        seen,
+        mode::TraceMode,
+        @nospecialize(track_numbers::Type)
+    ) where T
+        return $P{traced_type_inner(PT.parameters[1], seen, mode, track_numbers)}
+    end
+end
+
+@eval Base.@nospecializeinfer function traced_type_inner(
```
giordano, line 386: No need for `@eval` here, right?

src/Tracing.jl
```diff
@@ -366,14 +366,31 @@ end
...
+    @nospecialize(PT::Type{Core.LLVMPtr{T, A}})
```
giordano, line 387: Also, I'm not sure this would work, I had tried doing something like this and just using `T` in the body of the function instead of `PT.parameters[1]`, but it doesn't work (I supposed because this is a `@nospecialize` argument)

wsmoses: no the reason T in the function doesn't work is due to https://github.com/JuliaLang/julia/issues/57259 sadly.

Essentially we have LLVMPtr{TypeVar, TypeVar} and julia doesn't properly match that (see the issue)

giordano (approved):

PR #700, opened 2025-02-06
Add GPUCompiler and LLVM as deps to CUDA extension and run CUDA tests on macOS

giordano:
With [`GPUCompiler@v1.1.1`](https://github.com/JuliaGPU/GPUCompiler.jl/releases/tag/v1.1.1) we can run CUDA integration tests also on macOS, I verified locally (on aarch64-darwin) that these tests now are successfully.


wsmoses (approved):

Project.toml
```diff
@@ -57,13 +59,15 @@ AbstractFFTs = "1.5"
...
 ArrayInterface = "7.17.1"
 CEnum = "0.5"
-CUDA = "5.5"
+CUDA = "5.6"
 Downloads = "1.6"
 Enzyme = "0.13.28"
 EnzymeCore = "0.8.8"
 Functors = "0.5"
-GPUArraysCore = "0.1.6, 0.2"
+GPUArraysCore = "0.2"
```
wsmoses, line 67: Sorry going back to this, I'm worried that the very late GPUCompiler will cause issues for Oceananigans which is presently stuck at a lower version.

Obviously this is needed for CPU execution, but we could also consider lowering

giordano: I think https://github.com/CliMA/Oceananigans.jl/pull/4081 should help?

PR #704, opened 2025-02-07
[ReactantExtra] Stop removing references to `hardware_interference_size`

giordano:
We used to do that because in older macOS SDKs this variable was not available, but with https://github.com/JuliaPackaging/Yggdrasil/pull/10473 we switched to a newer SDK (11.3) and this hack isn't necessary anymore.

I verified locally that I could build successfully for both x86_64-darwin and aarch64-darwin in BinaryBuilder.  Opening as draft waiting for https://github.com/JuliaPackaging/Yggdrasil/pull/10473 to be merged.


giordano:
This should be good to go now

PR #707, opened 2025-02-08
[test] Remove top-level testset

giordano:
This makes tests more verbose to show progress live, instead of waiting several minutes to get any output.


avik-pal:
Without it the first failing test would prevent others from running no?

avik-pal:
There's https://github.com/ssfrr/TestSetExtensions.jl if we want to see progress

giordano:
Fair

PR #713, opened 2025-02-08
fix: use different API

avik-pal:
MLIR bindings regen wiped out my function :cry:

PR #721, opened 2025-02-10
[ReactantExtra] feat: OpSharding bindings for Julia

avik-pal:
splitting JLL changes from https://github.com/EnzymeAD/Reactant.jl/pull/713


avik-pal:
We should get some of the passes in PRs into EnzymeJAX and build the JLL together

wsmoses (approved):

PR #722, opened 2025-02-10
[ReactantExtra] fix: build on mac

avik-pal:
https://github.com/JuliaPackaging/Yggdrasil/pull/10494#issuecomment-2647257079


giordano (approved):
Using fixed-size aliases is a good idea, assuming whatever XLA API you may need to use it happy with that and you verified this still works.

PR #735, opened 2025-02-12
[WIP] control flow if

glou-nes:
I want to test with CI.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -475,6 +475,11 @@ function rewrite_insts!(ir, interp, guaranteed_error)
     return ir, any_changed
 end
 
+include("ext.jl")
+
+global dico = Dict()
+global dico2 = Dict()
+
 # Generator function which ensures that all calls to the function are executed within the ReactantInterpreter
 # In particular this entails two pieces:
 #   1) We enforce the use of the ReactantInterpreter method table when generating the original methodinstance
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -562,7 +567,18 @@ function call_with_reactant_generator(
         ir = CC.run_passes(frame.src, CC.OptimizationState(frame, interp), result, nothing)
         rt = CC.widenconst(CC.ignorelimited(result.result))
     else
-        ir, rt = CC.typeinf_ircode(interp, mi, nothing)
+        result = CC.InferenceResult(mi, CC.typeinf_lattice(interp))
+        world = CC.get_inference_world(interp)
+        src = CC.retrieve_code_info(result.linfo, world)
+        #dico2[mi]=(CC.copy(src), goto_if_not_protection(src))
+        @error src
+        src = goto_if_not_protection(src)
```
jumerckx, line 575: Billy's concern was that since the transformation here happens before inlining, the code that will later get doesn't have the needed conversions inserted.
I'm wondering if you run into this issue?
If so, the idea was to do the transformation by specializing the abstract interpreter.
If that's necessary, the correct place to do it would be `InferenceState` like I did here:
https://github.com/JuliaLLVM/MLIR.jl/blob/cf22c54222033ed4f5f3d68a620448bcaf766ac4/src/Generate/absint.jl#L151C1-L160C4
based on an old pr from valentin:
https://github.com/JuliaGPU/GPUCompiler.jl/pull/311

glou-nes: I don't have an example of such a case in mind! I will try with this one firstly (seem easier to handle).

PR #740, opened 2025-02-13
feat: sharding via IFRT

avik-pal:


deps/ReactantExtra/API.cpp
```diff
@@ -691,20 +679,26 @@ ClientCompile(PjRtClient *client, MlirModule cmod, int64_t device_id,
     options.executable_build_options.set_device_assignment(device_assignment);
   }
 
-  auto addressable_devices = client->addressable_devices();
```
avik-pal, line 694: @wsmoses do you know if this is needed? Seems like something XLA should be doing on its own?

wsmoses: I do not know

wsmoses (approved):

avik-pal:
I will follow up with more. Let's build this JLL, needed to support some more pjrt cases for now

PR #749, opened 2025-02-14
feat: JLL changes to expose HloModule

avik-pal:
There are some more changes for using ReactantThrowError instead of crashing julia with an assert


deps/ReactantExtra/API.cpp
```diff
@@ -936,19 +941,18 @@ extern "C" void XLAExecuteSharded(xla::PjRtLoadedExecutable *exec, int num_args,
...
+    ReactantThrowError(
+        ("Error: results.size()=" + std::to_string(results.size()) +
+         " does not match num_results=" + std::to_string(num_results) + "\n")
+            .c_str());
   }
 
   // Handle futures if they are returned.
-  if (returned_future.has_value()) {
-    *futures = true;
+  *futures = returned_future.has_value();
```
wsmoses, line 951: we should store the result in a temporary rather than reloading futures
```
has_future = returned_future.has_value();
*futures = has_future
if (has_future) {

}
```

PR #750, opened 2025-02-15
[IFRT] add ifrt-proxy server and client bindings

mofeing:


deps/ReactantExtra/API.cpp
```diff
@@ -1419,3 +1423,62 @@ extern "C" void
...
+    return nullptr;
+  }
+
+  const PJRT_Api *pluginLoad =
+  LoadPjrtPlugin("tpu", tpu_library_path.c_str(), error);
+  if (pluginLoad == nullptr)
+    return nullptr;
+  auto tpu_status = InitializePjrtPlugin("tpu", error);
+  if (tpu_status)
+    return nullptr;
```
mofeing, line 1449: we need to test wether to load the TPU plugin here (on this function call), on call by the factory (inside the closure then) or even before

wsmoses (approved):

PR #751, opened 2025-02-15
[IFRT] add c-bindings for "Held" PjRt classes

mofeing:
@wsmoses this PR should solve the issues that arised in #738

@avik-pal i've added `ifrt_client_make_single_shard_array_from_host_buffer` for creating single-shard arrays and `ifrt_client_assemble_array_from_single_shards` for joining them into a sharded array. `ifrt_pjrt_ArrayFromHostBuffer` (now renamed to `ifrt_pjrt_array_create`) is "IFRT-PjRt" only and wouldn't work for IFRT-proxy.

the only missing thing to use `ifrt_client_make_single_shard_array_from_host_buffer` is that we need to pass a `ifrt::Device*`...


deps/ReactantExtra/API.cpp
```diff
@@ -1280,41 +1280,298 @@ template <typename T> HeldValue<T> *capture(T obj) {
...
+  // TODO we need a `xla::ifrt::CompileOptions` but this is `xla::CompileOptions`
+  auto options = std::make_unique<CompileOptions>(
+    GenerateCompileOptions(
+      device_id,
+      is_sharded,
+      mesh_ids,
+      num_mesh_ids,
+      xla_gpu_cuda_data_dir
+    )
+  );
```
mofeing, line 1574: this is crashing because `compiler->Compile` is expecting a `ifrt::Compiler::CompileOptions` and not `xla::CompileOptions`

@avik-pal would you taking care of this? i think we need to use this https://github.com/openxla/xla/blob/81335eabdc55f0a2ff02bde7f79da7adda7af2c9/xla/python/ifrt/ir/ifrt_ir_program.h#L102-L109

wsmoses (approved):

mofeing:
Do not merge. See https://github.com/EnzymeAD/Reactant.jl/pull/751#discussion_r1957171189

PR #752, opened 2025-02-15
Support MPI

mofeing:


wsmoses:
you won't, instead you'll emit something like 

```

function send_wrap(%arg : memref<axb>) {
    mpi.send %arg
}

function main() {
    ...
    enzymexla.jit_call @set_wrap(%x : tensor<...>)
}
```

And then lower-jit will convert into a custom call. however you will need to define a lowering of mpi.send into a corresponding MPI_Send call [which will use the symbol you just registered here]

Re CUDA though we also need to ensure we are sync'd wrt the current custream which you can get via enzymexla.get_stream

ext/ReactantMPIExt/Overrides.jl
```diff
@@ -0,0 +1,110 @@
+@reactant_overlay @noinline function MPI.Init(; kwargs...)
+    if !isempty(kwargs)
+        @warn "Ignoring MPI.Init kwargs when tracing over MPI..." kwargs...
+    end
+    return Ops.init()
+end
+
+@reactant_overlay @noinline function MPI.Init(; kwargs...)
```
Pangoraw, line 8: This should be Finalize?

```suggestion
@reactant_overlay @noinline function MPI.Finalize(; kwargs...)
```

mofeing: ah funk yeah

mofeing:
mmm from our last discussion on this a couple of weeks ago, i understood that we would emit this

```mlir
function main() {
    ...
    mpi.send(%arg0, ...)
    ...
}
```

and it would get lowered to

```mlir
function send_wrap(%arg : memref<axb>) {
    llvm.call <0xffff> (%arg)
}

function main() {
    ...
    enzymexla.jit_call @send_wrap(%x : tensor<...>)
    ...
}
```

which will finally lower to the following with the enzymexla.jit pass

```mlir
function main() {
    ...
    stablehlo.custom_call @mpi_send_wrap(%x : tensor<...>)
    ...
}
```

is this correct or do we need to emit the `enzymexla.jit_call` directly from Reactant?

ahh or do you mean that any wrapping we need to do around MPI should be done in this way right?

> Re CUDA though we also need to ensure we are sync'd wrt the current custream which you can get via enzymexla.get_stream

okay, this will probably be required for NCCL

ext/ReactantMPIExt/Overrides.jl
```diff
@@ -0,0 +1,110 @@
...
+
+# TODO use `make_tracer` to linearize arbitrary types? check out `MPI.Buffer`
+function MPI.Isend(buf::TracedRArray, dest::Number, tag::Number, comm::MPI.Comm)
+    @assert comm == MPI.COMM_WORLD "Only MPI.COMM_WORLD is supported currently"
+
+    tag = if !(tag isa TracedRNumber)
+        Ops.constant(tag)
+    end
+
+    return dest = if !(dest isa TracedRNumber)
```
avik-pal, line 55: accidental?

mofeing: no, most of the times the source/dest ranks will be hardcoded. For example, the user can have the following code:

```julia
MPI.Send(traced_buffer, MPI.COMM_RANK; source=0)
```

but we need them to be injected into MLIR ops, so I first convert them to `tensor` with `Ops.constant`

avik-pal: I was referring to the return.

mofeing: ahhh yes, that's a typo

mofeing, line 55:
```suggestion
    dest = if !(dest isa TracedRNumber)
```

mofeing:
The PR is ready for review. The missing MPI routines are waiting to other PRs or need some fix, but they can be added later.

@giordano MPI testset result gets printed multiple times due to them being run on all ranks but I guess is not a problem?

giordano (commented):
> MPI testset result gets printed multiple times due to them being run on all ranks but I guess is not a problem?

That's unfortunate.  Just don't use `@testset`?  We don't in https://github.com/JuliaParallel/MPI.jl/tree/5ef7fef6d6c3e2ab2ad380f346c77235f47213bf/test

mofeing:
@wsmoses @avik-pal i'm getting the following error during precompilation

```
ERROR: LoadError: failed to add pipeline:MLIR Textual PassPipeline Parser:1:1: error: 'mark-func-memory-effects' does not refer to a registered pass or pass pipeline
```

test/integration/mpi.jl
```diff
@@ -21,4 +21,64 @@ end
...
+        return
+    end
+
+    # test MPI.jl Send / Reactant Recv
+    @testset "MPI.jl Send / Reactant Recv!" begin
+        send_buf = fill(1)
+        tag = 43
+        if rank == 0
+            MPI.Send(send_buf, comm; dest=1, tag=tag)
+            @test true
```
mofeing: you don't really need this

test/integration/mpi.jl
```diff
@@ -0,0 +1,84 @@
...
+end
+
+@testset "Send, Recv!" begin
+    comm = MPI.COMM_WORLD
+    rank = MPI.Comm_rank(comm)
+    nranks = MPI.Comm_size(comm)
+
+    if nranks < 2
+        @warn "Need at least 2 MPI processes for send tests. Skipping."
+        return
```
mofeing: You cannot return from something that it's not a function. I would error instead of just emitting a warning.

Also, add this error to the beginning of the file, not inside a testset.

romanlee: I don't think it should go at the beginning of the file, the other tests can run fine with only 1 rank. Maybe we just remove it from inside the testset, but only run that testset if nranks>2? I've just removed it for now, see 32cd922

test/integration/mpi.jl
romanlee: Removed, 32cd922

romanlee:
@wsmoses are there any other features we want to add before merging? If not, this might be ready for review

ext/ReactantMPIExt/Ops.jl
```diff
@@ -0,0 +1,685 @@
...
+
+function convert_julia_type_to_mpi_datatype(T::Type)
+    if T === Bool
+        MPI.C_BOOL
+    elseif T === Int8
+        MPI.INT8_T
+    elseif T === Int16
+        MPI.INT16_T
+    elseif T === Int32
+        MPI.INT32_T
```
wsmoses: @vchuravy is there a version of this already in MPI.jl?

ext/ReactantMPIExt/ReactantMPIExt.jl
```diff
@@ -0,0 +1,335 @@
+module ReactantMPIExt
+
+using Reactant
+using Reactant: Reactant, Distributed, MLIR
+using MPI: MPI
+using Libdl
+
+# https://github.com/jax-ml/jax/blob/b0117366686ab084d38ad2657d9a2ae3a581ca7e/jax/_src/clusters/mpi4py_cluster.py
+Distributed.is_env_present(::Distributed.MPIEnvDetector) = MPI.Initialized()
```
wsmoses, line 9: @avik-pal can you review this?

test/integration/mpi.jl
```diff
@@ -0,0 +1,124 @@
+using Test, MPI, Reactant
+
+# # MPI only works on cpu currently --- is this the right way/place to enforce that?
+# Reactant.set_default_backend("cpu")
```
wsmoses: @avik-pal re integration bits

test/runtests.jl
```diff
@@ -74,6 +74,11 @@ const REACTANT_TEST_GROUP = lowercase(get(ENV, "REACTANT_TEST_GROUP", "all"))
         @safetestset "SpecialFunctions" include("integration/special_functions.jl")
         @safetestset "Random" include("integration/random.jl")
         @safetestset "Python" include("integration/python.jl")
+        @safetestset "MPI" begin
+            using MPI
+            nranks = 2
+            run(`$(mpiexec()) -n $nranks $(Base.julia_cmd()) integration/mpi.jl`)
```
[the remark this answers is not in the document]
wsmoses: seems relevant here

wsmoses:
did a quick review and it looks good, though @avik-pal and @giordano should comemnt.

ext/ReactantMPIExt/ReactantMPIExt.jl
```diff
@@ -0,0 +1,335 @@
...
+    return Int(MPI.Comm_rank(new_comm))
+end
+
+function __init__()
+    # # TODO improve this, temporary hack
+    # # when  you fix it, remember to possibly make TracedType const again
+    # Reactant.TracedType = Union{Reactant.TracedRArray,Reactant.TracedRNumber,Reactant.MissingTracedValue,TracedRequest}
+
+    # TODO maybe it's more efficient if we use `RTLD_NOW` instead of `RTLD_LAZY`?
+    libmpi_handle = Libdl.dlopen(MPI.API.libmpi, RTLD_LAZY | RTLD_GLOBAL)
```
giordano: Why dlopening libmpi again?  `MPI.API.libmpi_handle` has the handle of the library already open.

test/integration/mpi.jl
```diff
@@ -0,0 +1,124 @@
...
+
+MPI.Init()
```
giordano, line 6: Do I understand correctly this (and `Finalize`) can't be `@compile`d at the moment?

romanlee: Yeah that's right. Although it looks like there are overrides in Overrides.jl. Let me see if I can get these working easily, otherwise maybe we just remove them unless it's a priority

ext/ReactantMPIExt/ReactantMPIExt.jl
romanlee: I'm not sure, maybe Sergio had a good reason for that. I'll see if it works with the already open  `MPI.API.libmpi_handle` one, and if so maybe we just go with that

romanlee: Update: all tests seem to run fine with `libmpi_handle = MPI.API.libmpi_handle`. Do you think we should make this change?

giordano: Yes, dlopening a library multiple times is like opening a can of worms, don't recommend at all.

romanlee: Alright, fixed. Thanks for pointing this out.

ext/ReactantMPIExt/Ops.jl
romanlee: There is! I've removed this function and replaced it with calls to MPI.Datatype(). See c6def59

wsmoses (approved):
generally lgtm, but I want @avik-pal to look atthe Distributed.is_env_present and related to double check

romanlee:
Sounds good. There are a couple things I'm trying to clean up in the meantime

ext/ReactantMPIExt/Overrides.jl
```diff
@@ -103,12 +92,7 @@ end
 #     return Ops.isend(buf, tag, dest)
 # end
 
-function MPI.Recv!(
-    buf::TracedRArray,
-    source::Integer,
-    tag::Integer,
-    comm::MPI.Comm
-)
+function MPI.Recv!(buf::TracedRArray, source::Integer, tag::Integer, comm::MPI.Comm)
```
romanlee, line 74: @giordano I don't understand the idea behind the formatting changes in 4899d6d. E.g., here, `Recv!` has a different format (args all on one line), from `Isend` above (all args on different lines). Could you explain?

giordano: That's JuliaFormatter (I only accepted the suggestions), whether to wrap or not depends on what would be the length of all the arguments on a single line.

romanlee: Interesting alright fair enough

test/integration/mpi.jl
avik-pal: For safety I would do `client = Reactant.default_backend(); Reactant.set_default_backend("cpu")` and at the end of the script set the client back with `Reactant.set_default_backend(client)`.

Techinically we are running on a separate process so it shouldn't matter but in case during testing (local/ci) we include the file it will make debugging harder.

ext/ReactantMPIExt/ReactantMPIExt.jl
avik-pal: This looks good to me

test/integration/mpi.jl
romanlee: Thanks, done! 508ed6a

test/integration/mpi.jl
romanlee: Finalize should be pretty simple to implement. However Init might involve a little more thinking. It looks like MPI.jl's MPI.Init never actually calls MPI_Init, rather it calls MPI_Init_thread (since it accomplishes the same thing and is threadsafe, so might as well?). We may want to do something similar? In any case I say we defer this convo for a future PR since not having these shouldn't be a show stopper. For now I've commented out Init and Finalize in Overrides.jl, 9a312b7

romanlee:
~Alright, seems to me like we're ready to merge~
Oh wait let me merge with main

romanlee:
Alright, now we should be ready to merge

avik-pal:
https://github.com/EnzymeAD/Reactant.jl/actions/runs/17982471081/job/51151993555?pr=752#step:19:808

```julia
┌ Error: Error during loading of extension ReactantMPIExt of Reactant, use `Base.retry_load_extensions()` to retry.
│   exception =
│    1-element ExceptionStack:
│    InitError: UndefVarError: `MPI_ERR_RMA_RANGE` not defined
│    Stacktrace:
│      [1] getproperty
│        @ .\Base.jl:31 [inlined]
│      [2] __init__()
│        @ ReactantMPIExt D:\a\Reactant.jl\Reactant.jl\ext\ReactantMPIExt\ReactantMPIExt.jl:211
│      [3] run_module_init(mod::Module, i::Int64)
│        @ Base .\loading.jl:1193
│      [4] register_restored_modules(sv::Core.SimpleVector, pkg::Base.PkgId, path::String)
│        @ Base .\loading.jl:1181
│      [5] _include_from_serialized(pkg::Base.PkgId, path::String, ocachepath::Nothing, depmods::Vector{Any})
│        @ Base .\loading.jl:1126
│      [6] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt128)
│        @ Base .\loading.jl:1651
│      [7] _require(pkg::Base.PkgId, env::Nothing)
│        @ Base .\loading.jl:2008
│      [8] __require_prelocked(uuidkey::Base.PkgId, env::Nothing)
│        @ Base .\loading.jl:1882
│      [9] #invoke_in_world#3
│        @ .\essentials.jl:926 [inlined]
│     [10] invoke_in_world
│        @ .\essentials.jl:923 [inlined]
│     [11] _require_prelocked
│        @ .\loading.jl:1873 [inlined]
│     [12] _require_prelocked
│        @ .\loading.jl:1872 [inlined]
│     [13] run_extension_callbacks(extid::Base.ExtensionId)
│        @ Base .\loading.jl:1365
│     [14] run_extension_callbacks(pkgid::Base.PkgId)
│        @ Base .\loading.jl:1400
│     [15] run_package_callbacks(modkey::Base.PkgId)
│        @ Base .\loading.jl:1224
│     [16] __require_prelocked(uuidkey::Base.PkgId, env::String)
│        @ Base .\loading.jl:1889
│     [17] #invoke_in_world#3
│        @ .\essentials.jl:926 [inlined]
│     [18] invoke_in_world
│        @ .\essentials.jl:923 [inlined]
│     [19] _require_prelocked(uuidkey::Base.PkgId, env::String)
│        @ Base .\loading.jl:1873
│     [20] macro expansion
│        @ .\loading.jl:1860 [inlined]
│     [21] macro expansion
│        @ .\lock.jl:267 [inlined]
│     [22] __require(into::Module, mod::Symbol)
│        @ Base .\loading.jl:1823
│     [23] #invoke_in_world#3
│        @ .\essentials.jl:926 [inlined]
│     [24] invoke_in_world
│        @ .\essentials.jl:923 [inlined]
│     [25] require(into::Module, mod::Symbol)
│        @ Base .\loading.jl:1816
│     [26] include(mod::Module, _path::String)
│        @ Base .\Base.jl:495
│     [27] exec_options(opts::Base.JLOptions)
│        @ Base .\client.jl:323
│     [28] _start()
│        @ Base .\client.jl:557
│    during initialization of module ReactantMPIExt
└ @ Base loading.jl:1371
```

Seems like a windows only issue.

avik-pal:
https://github.com/search?q=repo%3AJuliaParallel%2FMPI.jl%20MPI_ERR_RMA_RANGE&type=code. That symbol is not defined for MicrosoftMPI I think

romanlee:
That makes sense. There's probably a more elegant way do this https://github.com/EnzymeAD/Reactant.jl/pull/752/files#diff-e0797260c0a01b9b89a003a1ff707eeec8142e34f0c48c06a443bb8f18977a27R61 that could get around that issue. Maybe for now I'll just comment out `MPI_ERR_RMA_RANGE`? I think there are higher priorities for the moment.

ext/ReactantMPIExt/ReactantMPIExt.jl
```diff
@@ -0,0 +1,325 @@
...
+        :MPI_ERR_RMA_FLAVOR,
+        :MPI_ERR_RMA_SHARED,
+        # :MPI_T_ERR_INVALID,
+        # :MPI_T_ERR_INVALID_NAME,
+        # :MPI_ERR_PROC_ABORTED,
+        # :MPI_ERR_PROC_FAILED,
+        # :MPI_ERR_PROC_FAILED_PENDING,
+        # :MPI_ERR_REVOKED,
+    ]
+        value = getproperty(MPI.API, name)
```
avik-pal, line 212:
```suggestion
        !hasproperty(MPI.API, name) && continue
        value = getproperty(MPI.API, name)
```

ext/ReactantMPIExt/ReactantMPIExt.jl
```diff
@@ -0,0 +1,326 @@
...
+        :MPI_ERR_RMA_FLAVOR,
+        :MPI_ERR_RMA_SHARED,
+        # :MPI_T_ERR_INVALID,
+        # :MPI_T_ERR_INVALID_NAME,
+        # :MPI_ERR_PROC_ABORTED,
+        # :MPI_ERR_PROC_FAILED,
+        # :MPI_ERR_PROC_FAILED_PENDING,
+        # :MPI_ERR_REVOKED,
+    ]
+        !hasproperty(MPI.API, name) && continue
```
avik-pal:
```suggestion
        !isdefined(MPI.API, name) && continue
```

wsmoses:
somehow the branch again got into "This branch cannot be rebased due to conflicts"

romanlee:
Really? On my side it looks like I could push the squash and merge button. And I'm merged with the latest main commit.

avik-pal:
> somehow the branch again got into "This branch cannot be rebased due to conflicts"

That's probably because of the intermediate merge commits we have in the PR. we could squash and merge

wsmoses:
Once it seems to pass anyone who can actually merge it should go ahead and merge it xD

PR #754, opened 2025-02-16
Support tracing of `rem` with only one operand being a `ConcreteRNumber`

giordano:
This should help with https://github.com/CliMA/Oceananigans.jl/pull/4096#issuecomment-2661293409.  ~Opening as a draft until I figure out where to test these new methods.~ ***Edit***: tests added.

### DIFF
--- a/src/TracedRNumber.jl
+++ b/src/TracedRNumber.jl
@@ -97,10 +97,20 @@ for (jlop, hloop) in (
     (:(Base.mod), :remainder),
     (:(Base.rem), :remainder),
 )
-    @eval function $(jlop)(
-        @nospecialize(lhs::TracedRNumber{T}), @nospecialize(rhs::TracedRNumber{T})
-    ) where {T}
-        return Ops.$(hloop)(lhs, rhs)
+    @eval begin
+        function $(jlop)(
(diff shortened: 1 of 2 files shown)


src/TracedRNumber.jl
```diff
@@ -97,10 +97,24 @@ for (jlop, hloop) in (
     (:(Base.mod), :remainder),
     (:(Base.rem), :remainder),
```
giordano, line 98: The fact that `Base.mod` and `Base.rem` both are lowered to the same HLO operator looks very fishy since they have different semantics in Julia.

giordano: https://openxla.org/xla/operation_semantics#element-wise_binary_arithmetic_operations

> When `Op` is `Rem`, the sign of the result is taken from the dividend, and the absolute value of the result is always less than the divisor's absolute value.

This is the semantic of Julia's [`mod`](https://docs.julialang.org/en/v1/base/math/#Base.mod), the opposite of [`rem`](https://docs.julialang.org/en/v1/base/math/#Base.rem)

giordano: I opened to #755 to further discuss this issue.

wsmoses (approved):

PR #756, opened 2025-02-16
Fix for ocean

wsmoses:


ext/ReactantCUDAExt.jl
```diff
@@ -339,6 +339,14 @@ end
 
 Adapt.adapt_storage(to::KA.ConstAdaptor, a::CuTracedArray) = Base.Experimental.Const(a)
 
+struct ReactantRefValue{T} <: Ref{T}
+    val::T
+end
+Base.getindex(r::ReactantRefValue{T}) where {T} = r.val
+function Adapt.adapt_structure(to::ReactantKernelAdaptor, ref::Base.RefValue)
+    return ReactantRefValue(adapt(to, ref[]))
+end
```
giordano, line 348: This overrides an existing method: https://github.com/EnzymeAD/Reactant.jl/blob/ae6e60c923cce2b3932b4a90574d75817058e235/ext/ReactantCUDAExt.jl#L259-L261 I presume we should remove the old one?

wsmoses: Yeah we should remove the old one

PR #764, opened 2025-02-17
feat: initial IFRT integration

avik-pal:
I will split up this PR into multiple smaller PRs once this is ready

- [x] ReactantExtra changes
- [x] Move certain functions into `XLA.PJRT`
- [x] Fixes to the `global_ordinals`
- [x] Distributed Runtime https://github.com/EnzymeAD/Reactant.jl/pull/780
- [x] Sharding C-API https://github.com/EnzymeAD/Reactant.jl/pull/788


mofeing:
changes are being made to the C-API, and specifically on the way we should make interact PjRt and IFRT (due to discussions in #738, where I was adding the Julia API). see #751

we are overlapping a bit: i was waiting on #751 to push some of the API you're adding here 😅. i think we should first fix this https://github.com/EnzymeAD/Reactant.jl/pull/751#discussion_r1957171189 and then add the Julia API. i have all the API you want locally but we need first to fix this comment i left, as otherwise sharding won't work for distributed TPU (which is why we want IFRT in the first place).

avik-pal:
I am not adding any PjRt-IfRt interaction stuff (Held* functions in that PR) here (except the ifrt::Client constructed from the PjRtClient).

We need to use the direct IFRT calls for constructing arrays from HloSharding which doesn't take PjRtBuffers (at least not without multiple roundtrips)

src/ConcreteRArray.jl
```diff
@@ -10,35 +10,50 @@ end
 function Base.collect(x::ConcreteRNumber{T}) where {T}
     return collect(ConcreteRArray{T,0}(copy(x).data, ()))
 end
+function Base.collect(x::ConcreteIFRTNumber{T}) where {T}
```
wsmoses, line 13: can we for ease here just define ConcreteRNumber = Union{ConcreteIFRTNumber, ConcretePJRTNumber} or something, to avoid rewriting all this?

avik-pal: I added a `AbstractConcreteNumber` for this. Some of the dispatches are duplicated either because

1. storage for IFRT is simpler since sharding is embedded with the IFRT.Array
2. needs to map to ConcreteIFRTArray

wsmoses: ah okay makes sense

avik-pal:
```
I0000 00:00:1740173243.405339 1677674 pjrt_client.cc:524] PjRt-IFRT device count: total=4, addressable=2
I0000 00:00:1740173243.405393 1677674 pjrt_client.cc:528] Addressable PjRt-IFRT device: CpuDevice(id=0)
I0000 00:00:1740173243.405399 1677674 pjrt_client.cc:528] Addressable PjRt-IFRT device: CpuDevice(id=1)
I0000 00:00:1740173243.405382 1677675 pjrt_client.cc:524] PjRt-IFRT device count: total=4, addressable=2
I0000 00:00:1740173243.405436 1677675 pjrt_client.cc:528] Addressable PjRt-IFRT device: CpuDevice(id=131072)
I0000 00:00:1740173243.405442 1677675 pjrt_client.cc:528] Addressable PjRt-IFRT device: CpuDevice(id=131073)
Reactant.XLA.process_index(client) = 1
Reactant.XLA.process_index(client) = 0
1 CPU:0 cpu
0 CPU:0 cpu
1 CPU:1 cpu
1 CPU:131072 cpu
1 CPU:131073 cpu
0 CPU:1 cpu
0 CPU:131072 cpu
0 CPU:131073 cpu
1 CPU:131072 cpu
1 CPU:131073 cpu
0 CPU:0 cpu
0 CPU:1 cpu
```

CPU setup for distributed (and presumably TPU setup) now works

wsmoses (approved):

PR #771, opened 2025-02-19
refactor: move PJRT into a specific module

avik-pal:
Mostly a non-functional change. Moves all PJRT stuff into `XLA.PJRT` and sets up abstract types and functions that we need to implement for IFRT


wsmoses (approved):

avik-pal:
needs to wait on https://github.com/JuliaPackaging/Yggdrasil/pull/10563

PR #774, opened 2025-02-19
feat: initial draft of IFRT.Array interface

avik-pal:


avik-pal:
@mofeing for the sharding we can't use the HeldPjRtClient -> ifrt::Client, can we?

```
ERROR: INVALID_ARGUMENT: Only SingleDeviceSharding or fully-replicated sharding is supported: sharding=HloSharding(memory_kind: unpinned_host, hlo_sharding: {devices=[2,4]<=[4,2]T(1,0)})

Stacktrace:
 [1] reactant_err
   @ ~/reactant/Reactant.jl/src/xla/Utils.jl:12
 [2] Reactant.XLA.IFRT.Array(client::Reactant.XLA.IFRT.Client, array::Matrix{Float32}, sharding::Reactant.XLA.IFRT.Sharding)
   @ Reactant.XLA.IFRT ~/reactant/Reactant.jl/src/xla/IFRT/Array.jl:33
 [3] top-level scope
   @ REPL[9]:1
```

Do we need to create a client via the proxy APIs?

avik-pal:
moving this into a single PR

mofeing:
ah damn, this is wrong. i detected sth might be wrong while reviewing your error. so there's a method that can create a `xla::ifrt::PjRtClient` from a `xla::PjRtClient` https://github.com/openxla/xla/blob/ab8fb43e97a9f1faaaba44d285b4c394bb2ab452/xla/python/pjrt_ifrt/pjrt_client.h#L144-L145 which is what i thought of using but it's deprecated

we should be using this method
https://github.com/openxla/xla/blob/ab8fb43e97a9f1faaaba44d285b4c394bb2ab452/xla/python/pjrt_ifrt/pjrt_client.h#L138-L139 but needs a lil bit more of engineering

our problem is in here: https://github.com/EnzymeAD/Reactant.jl/blob/83a2c1de6d33f60733cbeed82af906c764d08679/deps/ReactantExtra/API.cpp#L1340
we are trying to create a `xla::ifrt::PjRtClient::CreateOptions` with a `xla::PjRtClient` which is blatantly wrong and i don't know how it compiled then

> Do we need to create a client via the proxy APIs?

i believe it's the only way: client creation is the only backend-specific method we should be calling.

PR #775, opened 2025-02-20
refactor: rename R* to PJRT*

avik-pal:


README.md
```diff
@@ -10,16 +10,16 @@
 
 Reactant takes Julia function and compile it into MLIR and run fancy optimizations on top of it, including using EnzymeMLIR for automatic differentiation, and create relevant executables for CPU/GPU/TPU via XLA. It presently operates as a tracing system. Compiled functions will assume the same control flow pattern as was original taken by objects used at compile time, and control flow (e.g. if, for) as well as any type instabilities will be removed. The benefits of this approach is immediately making all such code available for advanced optimization with little developer effort. This system and corresponding semantics is subject to change to a (potentially partial) source rewriter in the future.
 
-Reactant provides two new array types at its core, a ConcreteRArray and a TracedRArray. A ConcreteRArray is an underlying buffer to whatever device data you wish to store and can be created by converting from a regular Julia Array.
+Reactant provides two new array types at its core, a ConcretePJRTArray and a TracedRArray. A ConcretePJRTArray is an underlying buffer to whatever device data you wish to store and can be created by converting from a regular Julia Array.
```
wsmoses, line 13: we definitely should still have as main docs and everything as concreterarray

wsmoses: it's fine for us to define ConcreteRArray = ConcretePJRTArray

or say make ConcreteRArray have an extra type parameter for the api or something

avik-pal: This was accidental from find and replace

benchmark/setup.jl
```diff
@@ -55,20 +55,20 @@ function setup_simple_benchmark!(suite::BenchmarkGroup, backend)
         suite["(Basics) 2D sum (2 x 10)"]["forward (compilation)"][backend][tag] = @benchmarkable begin
             @compile optimize = $(opt_pass) sum(x)
         end setup = begin
-            x = Reactant.ConcreteRArray(ones(2, 10))
+            x = Reactant.ConcretePJRTArray(ones(2, 10))
```
wsmoses, line 58: we should just define ConcreteRArray(x) = ConcretePJRT(x) imo

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -94,9 +94,9 @@ You need to ensure that all branches have the same type.
 Another example is the following for loop which changes the type of `x` between iterations.
 
 ```julia
-x = ... # ConcreteRArray{Int64, 1}
+x = ... # ConcretePJRTArray{Int64, 1}
```
wsmoses, line 97: leave these the same?

src/Ops.jl
```diff
@@ -1071,11 +1071,11 @@ end
         @assert 0 < dimension <= ndims(x) "$x invalid dimension"
     end
 
-    sample_inputs = Vector{Reactant.ConcreteRNumber}(undef, length(xs) * 2)
+    sample_inputs = Vector{Reactant.ConcretePJRTNumber}(undef, length(xs) * 2)
```
wsmoses, line 1074: now that I actually see this, we shouldn't need to create concrete values here, but we can fix that later

benchmark/setup.jl
```diff
@@ -55,20 +55,20 @@ function setup_simple_benchmark!(suite::BenchmarkGroup, backend)
...
         end
 
         suite["(Basics) sum(cos, x) (2 x 10)"]["forward (compilation)"][backend][tag] = @benchmarkable begin
             @compile optimize = $(opt_pass) sumcos(x)
         end setup = begin
-            x = Reactant.ConcreteRArray(ones(2, 10))
+            x = Reactant.ConcretePJRTArray(ones(2, 10))
```
wsmoses, line 64: these should remain as is imo

test/basic.jl
```diff
@@ -632,30 +632,35 @@ end
...
+    cond = ConcretePJRTNumber(true)
+    x = ConcretePJRTNumber(1.0)
+    @test @jit(ifelse(cond, x, 0.0)) == ConcretePJRTNumber(1.0)
+    @test @jit(ifelse(cond, 0.0, x)) == ConcretePJRTNumber(0.0)
+    @test @jit(ifelse(cond, 1.0, 0.0)) == ConcretePJRTNumber(1.0)
+    @test @jit(ifelse(cond, 0.0, 1.0)) == ConcretePJRTNumber(0.0)
 end
 
-@testset "fill! and zero on ConcreteRArray" begin
+@testset "fill! and zero on ConcretePJRTArray" begin
```
wsmoses, line 659: all the tests should just use ConcreteRNumber/Array imo

avik-pal: changed to `to_rarray` or `ConcreteRNumber`. Checks for types are using ConcretePJRTArray. We will need to switch those manually since IFRT wont store a tuple of buffers

benchmark/setup.jl
avik-pal: done

avik-pal:
@giordano I don't think this is breaking. Any code using `ConcreteRArray`/`ConcreteRNumber` should continue to work. If that doesn't, we should fix it

wsmoses (approved):

giordano:
This did break Oceananigans: https://github.com/CliMA/Oceananigans.jl/blob/2916ad139ff53140d83a2de7db0e0c1512187885/ext/OceananigansReactantExt/Architectures.jl#L14

PR #776, opened 2025-02-20
[build_local] Add argument to set `--color` option for Bazel

giordano:


giordano:
We have [automatic color output from Bazel](https://github.com/EnzymeAD/Reactant.jl/actions/runs/13434240277/job/37532814915?pr=776#step:7:43) now :slightly_smiling_face:

PR #787, opened 2025-02-22
Temporarily use fork of XLA to point to fork of LLVM

giordano:
This for including https://github.com/llvm/llvm-project/pull/128248

### DIFF
--- a/deps/ReactantExtra/WORKSPACE
+++ b/deps/ReactantExtra/WORKSPACE
@@ -130,7 +130,7 @@ http_archive(
 )
 
 # load("@jax//third_party/xla:workspace.bzl", "XLA_COMMIT", "XLA_SHA256")
-XLA_COMMIT = "b34ff92fce1c51bf895a80fb1a3672b1301eb7aa"
+XLA_COMMIT = "610bf0cda8acc144fc843a6580282c39fd537824"
 XLA_SHA256 = ""
 
 http_archive(


deps/ReactantExtra/WORKSPACE
```diff
@@ -86,15 +86,15 @@ LLVM_TARGETS = select({
...
-# LLVM_COMMIT = "b39c5cb6977f35ad727d86b2dd6232099734ffd3"
-# LLVM_SHA256 = ""
-# http_archive(
-#     name = "llvm-raw",
-#     build_file_content = "# empty",
-#     sha256 = LLVM_SHA256,
-#     strip_prefix = "llvm-project-" + LLVM_COMMIT,
-#     urls = ["https://github.com/llvm/llvm-project/archive/{commit}.tar.gz".format(commit = LLVM_COMMIT)],
-# )
+LLVM_COMMIT = "7d81eb0c711713252acc523722bc2bd647fbdc87"
```
wsmoses, line 89: you need the latter lines as well, thorugh zstd / etc. and I'm not sure if it plays nicely with ygg

giordano: As an alternative I can point XLA to https://github.com/giordano/xla/commit/610bf0cda8acc144fc843a6580282c39fd537824, would that be better?

wsmoses: yeah it would

giordano: Done.

PR #796, opened 2025-02-23
feat: JLL changes for IFRT integration

avik-pal:


deps/ReactantExtra/API.cpp
```diff
@@ -1465,43 +1388,65 @@ FreeHloModule(HeldValue<std::shared_ptr<xla::HloModule>> *hlo_module) {
...
-  std::string tpu_library_path;
-  if (auto path = llvm::sys::Process::GetEnv(kEnvTpuLibraryPath)) {
-    tpu_library_path = *path;
-  } else if (tpu_path) {
-    tpu_library_path = std::string(tpu_path);
-  } else {
-    *error = "Could not find TPU path";
-    return nullptr;
-  }
+// XXX: Bring back with the correct API
```
avik-pal, line 1391: The proxy APIs need to be setup with the correct kv_store from the DistributedRuntimeClient without which they should anyways fail. I will bring these ones back once the IFRT-PJRT clients land

avik-pal:
Builds are going though and none of the current tests are for IFRT merging

PR #797, opened 2025-02-23
[Compiler] Make `raise` a keyword argument

giordano:
@wsmoses is this what you had in mind?

Note that raising at the moment is broken because we need the new JLL (that's done in #788)


giordano:
Uhm, I don't know how to adapt https://github.com/EnzymeAD/Reactant.jl/blob/4aaae3c2be53d315eb88ab2ae8d6c964aa3f3870/ext/ReactantCUDAExt.jl#L297 That code wasn't calling explicitly any of the the Reactant macros.

src/Compiler.jl
```diff
@@ -588,6 +586,7 @@ function compile_mlir!(
     optimize::Union{Bool,Symbol}=true,
     no_nan::Bool=false,
     backend="gpu",
+    raise::Bool=false,
```
wsmoses, line 589: can we make this a union bool and string? That way we can change the raise pipeline for debugging without recompiling reactant

wsmoses:
> Uhm, I don't know how to adapt
> 
> https://github.com/EnzymeAD/Reactant.jl/blob/4aaae3c2be53d315eb88ab2ae8d6c964aa3f3870/ext/ReactantCUDAExt.jl#L297
> 
> That code wasn't calling explicitly any of the the Reactant macros.

yeah here I think we want to push the current "is it being raised" state like we do for an mlir context/module.

src/Compiler.jl
giordano: Yes, can do it.

giordano:
> like we do for an mlir context/module.

Can you point me to that?

wsmoses:
like https://github.com/EnzymeAD/Reactant.jl/blob/dbc0fe1359c863f507794d99ef73482a08cf8a54/src/mlir/IR/Module.jl#L63 and surrounding code for example

giordano:
Ok, CUDA integration tests are passing now.

Any suggestions for tests?  I know I can do something basic like `rsin = @compile raise="..." sin.(Reactant.to_rarray(rv)` and then call `rsin(rv)`, but that's only checking that the keyword argument doesn't trigger an error, I don't know how to test that the desired passes are indeed applied.

wsmoses:
Assuming raising works decently well, we should be able to do code_hlo of a manual sin broadcast kernel, and it should optimize it to a stablehlo.sin across the entire tensor (without a custom call)

giordano:
Ok, should that be postponed to later when raising is fixed?  In the meantime I added very basic tests that are simply making sure the keyword argument is accepted by `@jit` (as a prototype for all other macros)

wsmoses:
yeah / sg!

src/Compiler.jl
```diff
@@ -612,6 +611,12 @@ function compile_mlir!(
     MLIR.IR.activate!(MLIR.IR.body(mod))
     activate_callcache!(callcache)
 
+    # Save in the TLS whether we are raising.  We identify that condition by
+    # checking whether the user set an explicit list of passes, or chose
+    # `raise=true` to use the default passes.
+    is_raising = raise isa String || raise
+    task_local_storage(:reactant_is_raising, is_raising)
+
```
wsmoses, line 619: we should pop this out of TLS when compile returns

wsmoses: ideally we should make helper functions like https://github.com/EnzymeAD/Reactant.jl/blob/192bef35b8ee81279060bd6f2953c95b2593472e/src/mlir/IR/Block.jl#L172

to get and set.

Moreover, we should also make this something that is pushed and popped (like there)

wsmoses (approved):

PR #798, opened 2025-02-23
feat: add `Ops.sharding_constraint`

avik-pal:
I am hoping this allows us to generate a code to do the communication to all nodes for the IFRT case.

- [x] needs a test before merging


deps/ReactantExtra/API.cpp
```diff
@@ -2133,6 +2133,8 @@ hloShardingFromTensorShardingAttr(mlir::sdy::TensorShardingAttr attr,
       xla::sdy::convertToHloSharding(attr, get_mesh_attr, manual_axes));
 }
 
+// XXX: This is incorrect. We need to use the current mesh to generate this
+// instead of the global mesh
```
avik-pal, line 2137: This is bad. Incidentally on main we don't support multiple meshes so we weren't hitting this issue. But I am having some strange failures:

```mlir
module @reactant_fn attributes {mhlo.num_partitions = 8 : i64, mhlo.num_replicas = 1 : i64} {
  sdy.mesh @mesh = <["x"=2, "y"=2, "z"=2]>
  func.func @main(%arg0: tensor<2x2x4xf32> {sdy.sharding = #sdy.sharding<@mesh, [{"z"}, {}, {"x"}]>}) -> tensor<2x2x4xf32> {
    %0 = sdy.reshard %arg0 <@mesh, [{}, {}, {}]> : tensor<2x2x4xf32>
    return %0 : tensor<2x2x4xf32>
  }
}
```

```
Detailed error from MLIR: /home/avik-pal/reactant/Reactant.jl/src/Ops.jl:2260:0: error: 'sdy.reshard' op unknown mesh: @mesh
/home/avik-pal/reactant/Reactant.jl/src/Ops.jl:2260:0: note: see current operation: %0 = "sdy.reshard"(%arg0) <{sharding = #sdy.sharding<@mesh, [{}, {}, {}]>}> : (tensor<2x2x4xf32>) -> tensor<2x2x4xf32>
```

It could potentially originate from that

avik-pal: I fixed the mesh regeneration. But the error persists...

wsmoses: this feels like we didn't register the shardy dialect

avik-pal:
This also sets us up for other sdy ops that would be needed for the kernel sharding

avik-pal:
Note for future self: we are not supposed to insert `Ops.reshard` into the IR, those are meant to be generated from the sharding_constraints with an MLIR pass

test/sharding.jl
```diff
@@ -173,6 +173,51 @@ fn_test4(x, y) = x .+ sin.(y')
...
+        x_ra = Reactant.to_rarray(
+            x; sharding=Sharding.NamedSharding(mesh, ("data", "model"))
+        )
+
+        constraint = Sharding.NamedSharding(mesh, ("model", nothing))
+
+        function fn_with_constraint(x)
+            y = x .+ x
+            return Reactant.Ops.sharding_constraint(y, constraint)
+        end
```
avik-pal, line 190: @mofeing I think this is the case you were interested in. Where we don't specify how the inputs are sharded, but we want to do it for some intermediate tensor. Then we automatically infer how to partition the input tensors

wsmoses (approved):

PR #801, opened 2025-02-24
cf: extract condition to a name

### LINKED ISSUE #768: @trace seems to fail when `if` evaluates a function
We haven't been able to reproduce this simply yet but this code:

```julia
    for callback in values(sim.callbacks)
        @trace if callback.schedule(sim.model)
            callback(sim)
        end
```

errors with

```julia
  Got exception outside of a @test
  TypeError: non-boolean (Reactant.TracedRNumber{Bool}) used in boolean context
  Stacktrace:
    [1] macro expansion
      @ ~/.julia/packages/ReactantCore/t40zc/src/ReactantCore.jl:378 [inlined]
    [2] time_step!
      @ ~/Projects/dev/OC1.jl/ext/OceananigansReactantExt/Simulations/run.jl:107 [inlined]
```

and it's fixed by

```julia
    for callback in values(sim.callbacks)
        need_to_call = callback.schedule(sim.model)
        @trace if need_to_call
            callback(sim)
        end
```

found with @giordano

Pangoraw:
Closes #768.

### DIFF
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -225,7 +225,11 @@ function trace_if_with_returns(mod, expr)
     new_expr, _, all_check_vars = trace_if(
         mod, expr.args[2]; store_last_line=expr.args[1], depth=1
     )
+    cond_name = first(all_check_vars)
+    original_cond = expr.args[2].args[1]
+    expr.args[2].args[1] = cond_name
     return quote
+        $(cond_name) = $(original_cond)
         if $(within_compile)() && $(any)($(is_traced), ($(all_check_vars...),))
             $(new_expr)
         else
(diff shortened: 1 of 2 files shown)


Pangoraw:
this breaks one test, looking into this

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -225,7 +225,11 @@ function trace_if_with_returns(mod, expr)
     new_expr, _, all_check_vars = trace_if(
         mod, expr.args[2]; store_last_line=expr.args[1], depth=1
     )
+    cond_name = first(all_check_vars)
+    original_cond = expr.args[2].args[1]
+    expr.args[2].args[1] = cond_name
     return quote
+        $(cond_name) = $(original_cond)
```
jumerckx, line 232: This special handling might not be required if you wrap everything in a `begin-end` block?
```
result = begin
	cond = ...
	if cond
	...
	end
end
```
Feel free to ignore though, this works as well.

jumerckx (approved):
Cool! LGTM

lib/ReactantCore/src/ReactantCore.jl
Pangoraw: So the problem here was that the new cond_name is added to the list of variables in all_check_vars so it has to be defined before the check `if within_compile() && any(is_traced, (cond,...))`.

PR #805, opened 2025-02-24
Fix raising

wsmoses:


giordano:
For the record, XLA now integrated a pretty recent llvm, so we can stop using my fork introduced in  #787

wsmoses:
yeah, we should first update the jax commit in enzyme jax then remove the overload here

PR #808, opened 2025-02-25
adding transpose!() to linear algebra

tharittk:
Hello, I'm an applicant to GSoC this summer.  Avik told me that LinearAlgebra.jl is good for getting up to speed.

I started with what I assume is the most simple function. 

I hope I took the right approach by re-using the functions from Ops.jl

Please advise if I make any mistakes or anything I can do better.

### DIFF
--- a/src/stdlibs/LinearAlgebra.jl
+++ b/src/stdlibs/LinearAlgebra.jl
@@ -397,4 +397,21 @@ function LinearAlgebra._kron!(C::AnyTracedRMatrix, A::AnyTracedRMatrix, B::AnyTr
     return C
 end
 
+LinearAlgebra.transpose(a::AnyTracedRArray) = error("transpose not defined for $(typeof(a)).")
+
+function LinearAlgebra.transpose!(B::AnyTracedRVector, A::AnyTracedRMatrix)
+    LinearAlgebra.check_transpose_axes((size(B,1), size(B,2)), size(A))
+    set_mlir_data!(B, get_mlir_data(Ops.reshape(A, length(B))))
+end
+
+function LinearAlgebra.transpose!(B::AnyTracedRMatrix, A::AnyTracedRVector)
+    LinearAlgebra.check_transpose_axes(size(B), (size(A, 1), size(A, 2)))
+    set_mlir_data!(B, get_mlir_data(Ops.broadcast_in_dim(A, [2], [1, length(A)])))
+end
+
+function LinearAlgebra.transpose!(B::AnyTracedRMatrix, A::AnyTracedRMatrix)
+    LinearAlgebra.check_transpose_axes(size(B), size(A))
+    set_mlir_data!(B, get_mlir_data(Ops.transpose(A, [2,1])))
+end
+
 end
--- a/test/integration/linear_algebra.jl
+++ b/test/integration/linear_algebra.jl
@@ -183,3 +183,29 @@ end
         end
     end
 end
+
+@testset "transpose!" begin
+    v = zeros(5)
+    M = rand(1, 5)
+    v_ra = Reactant.to_rarray(v)
+    M_ra = Reactant.to_rarray(M)
+    
+    @jit transpose!(v_ra, M_ra)
+    @test v_ra ≈ transpose!(v, M)
+
+    v = rand(7)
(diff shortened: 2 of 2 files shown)


test/integration/linear_algebra.jl
```diff
@@ -183,3 +183,29 @@ end
         end
     end
 end
+
+@testset "transpose!" begin
+    v = zeros(5)
+    M = rand(1, 5)
```
mofeing, line 189: Would you mind not using `rand`? we had problems tracking numerical stability and bugs due to `rand`.

src/stdlibs/LinearAlgebra.jl
```diff
@@ -397,4 +397,21 @@ function LinearAlgebra._kron!(C::AnyTracedRMatrix, A::AnyTracedRMatrix, B::AnyTr
...
+end
+
+function LinearAlgebra.transpose!(B::AnyTracedRMatrix, A::AnyTracedRVector)
+    LinearAlgebra.check_transpose_axes(size(B), (size(A, 1), size(A, 2)))
+    set_mlir_data!(B, get_mlir_data(Ops.broadcast_in_dim(A, [2], [1, length(A)])))
+end
+
+function LinearAlgebra.transpose!(B::AnyTracedRMatrix, A::AnyTracedRMatrix)
+    LinearAlgebra.check_transpose_axes(size(B), size(A))
+    set_mlir_data!(B, get_mlir_data(Ops.transpose(A, [2,1])))
```
mofeing, line 414: @avik-pal @wsmoses Should we enforce same `eltype` for both args? if not, there can be undetected problems when we pass two arrays of same size but different `eltype`.

mofeing (commented):
lgtm

test/integration/linear_algebra.jl
tharittk: No problem. I will make a change to that. Also, the rest of the test set uses ```rand```.  Should I also change them?

src/stdlibs/LinearAlgebra.jl
```diff
@@ -397,4 +397,21 @@ function LinearAlgebra._kron!(C::AnyTracedRMatrix, A::AnyTracedRMatrix, B::AnyTr
     return C
 end
 
+LinearAlgebra.transpose(a::AnyTracedRArray) = error("transpose not defined for $(typeof(a)).")
```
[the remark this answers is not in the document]
tharittk: The stdlib LinearAlgebra.jl has this:
```transpose(a::AbstractArray) = error("transpose not defined for $(typeof(a)). Consider using `permutedims` for higher-dimensional arrays.")```

The transpose!() docstring also mentions about not supporting the in-place version.

"""
transpose!(dest,src)

Transpose array `src` and store the result in the preallocated array `dest`, which should
have a size corresponding to `(size(src,2),size(src,1))`. No in-place transposition is
supported and unexpected results will happen if `src` and `dest` have overlapping memory
regions.
"""

So yeah, I was just trying to mimic it exactly.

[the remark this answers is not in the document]
mofeing: i guess that then you can leave it unimplemented and it will call the default method which will error like is supposed to

src/stdlibs/LinearAlgebra.jl
avik-pal: We should do a promote and set here. Base implements it like `B[j,i] = f(A[i,j])` so its auto promoted

tharittk: noted. I'll go fix this.

mofeing: i know we do promote here but is it the correct thing? e.g. in `transpose!(::TracedRArray{Int}, ::TracedRArray{Float64})`, promotion of eltype will be `Float64` and the result we are storing will be wrong.

or should the promotion handle it and error in that case?

avik-pal: Ah I see what's going on. Unfortunately stablehlo doesn't seem to have conversion mode supported (https://github.com/openxla/stablehlo/issues/180). We can do the convert here and just defer to stablehlo to handle the truncation.

cc @wsmoses any thoughts here?

PR #810, opened 2025-02-25
Return T if mode=TracedSetPath

glwagner:


wsmoses:
@avik-pal I don't want to distract you from mesh stuff, but can you potentially lend a hand here?

avik-pal:
what is the issue here?

wsmoses:
CI is sad.

Essentially the reason for this PR is because a data structure passed non-traced TwicePrecision{Float64} and StepRange into a traced if. TracedIf asks for tracednumbers, which are ilelgal for steprange [expected the inner elem to be an int].

This PR doesn't change the inner value types to change on a tracedtrack and related [since it should just update in place]

PR #813, opened 2025-02-26
adding axpy! and axpby! to linear algebra

tharittk:
I follow the standard lib LinearAlgebra.jl that it always casts ```a``` and ```b``` to the ```eltype``` of ```X``` and ```Y```.  If ```a``` is ```float``` and ```X``` has ```eltype Int``` , it will throw an error.

### DIFF
--- a/src/stdlibs/LinearAlgebra.jl
+++ b/src/stdlibs/LinearAlgebra.jl
@@ -404,6 +404,7 @@ function LinearAlgebra.axpy!(α::Number, x::TracedRArray{T}, y::TracedRArray{T})
     ax = Ops.multiply(x, TracedUtils.broadcast_to_size(T(α), size(x)))
     
     set_mlir_data!(y, get_mlir_data(Ops.add(y, ax)))
+    return y
 end
 
 function LinearAlgebra.axpby!(α::Number, x::TracedRArray{T}, β::Number, y::TracedRArray{T}) where {T}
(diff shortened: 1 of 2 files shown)


src/stdlibs/LinearAlgebra.jl
```diff
@@ -397,4 +397,24 @@ function LinearAlgebra._kron!(C::AnyTracedRMatrix, A::AnyTracedRMatrix, B::AnyTr
...
 end
 
+function LinearAlgebra.axpy!(α::Number, x::TracedRArray{T}, y::TracedRArray{T}) where {T}
+    if length(x) != length(y)
+        throw(DimensionMismatch(lazy"x has length $(length(x)), but y has length $(length(y))"))
+    end
+    ax = Ops.multiply(x, TracedUtils.broadcast_to_size(T(α), size(x)))
+    
+    set_mlir_data!(y, get_mlir_data(Ops.add(y, ax)))
+end
```
mofeing, line 407: the base method also returns `y`

```suggestion
    return y
end
```

src/stdlibs/LinearAlgebra.jl
```diff
@@ -397,4 +397,24 @@ function LinearAlgebra._kron!(C::AnyTracedRMatrix, A::AnyTracedRMatrix, B::AnyTr
...
+
+function LinearAlgebra.axpby!(α::Number, x::TracedRArray{T}, β::Number, y::TracedRArray{T}) where {T}
+    if length(x) != length(y)
+        throw(DimensionMismatch(lazy"x has length $(length(x)), but y has length $(length(y))"))
+    end
+    ax = Ops.multiply(x, TracedUtils.broadcast_to_size(T(α), size(x)))
+    by = Ops.multiply(y, TracedUtils.broadcast_to_size(T(β), size(y)))
+ 
+    set_mlir_data!(y, get_mlir_data(Ops.add(ax, by)))
+end
```
mofeing, line 417:
```suggestion
    return y
end
```

mofeing (commented):
lgtm

avik-pal (approved):

PR #825, opened 2025-02-28
feat: sharding with non-divisible axes [alternatve approach]

avik-pal:


avik-pal:
One way to handle this would be to:

1. Detect potential padding
2. If no padding is needed, do what we were doing before
3. If padding is needed, construct a `DeferredSharding` object that constructs the buffers based on the final program

avik-pal:
A major caveat of this approach is that out `shard_type` won't be able to determine the correct types

avik-pal:
https://github.com/openxla/xla/blob/a83a2d3f1f977e4825cb210320597c5825c25ead/xla/client/executable_build_options.h#L327-L330 are false by default. so this is a bit unexpected.

```mlir
module @reactant_fn_test attributes {mhlo.frontend_attributes = {xla.sdy.meshes = "{mesh = #sdy.mesh<[\\\22data\\\22=2, \\\22model\\\22=4]>}"}, mhlo.num_partitions = 8 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<7x2xf32> {mhlo.frontend_attributes = {xla.sdy.sharding = "#sdy.sharding<@mesh, [{\\\22model\\\22}, {\\\22data\\\22}]>"}, mhlo.sharding = "{devices=[4,2]<=[2,4]T(1,0)}"}) -> tensor<7x1xf32> {
    %0 = mhlo.constant dense<0.000000e+00> : tensor<f32>
    %1 = mhlo.reduce(%arg0 init: %0) applies mhlo.add across dimensions = [1] : (tensor<7x2xf32>, tensor<f32>) -> tensor<7xf32>
    %2 = mhlo.reshape %1 : (tensor<7xf32>) -> tensor<7x1xf32>
    return %2 : tensor<7x1xf32>
  }
}
```

MHLO sharding: `{devices=[4,2]<=[2,4]T(1,0)}`
But after XLA passes: `sharding={devices=[1,2,4]<=[8] last_tile_dim_replicate}`

src/xla/Sharding.jl
```diff
@@ -251,7 +251,7 @@ function sharding_to_concrete_array_indices(
     sharding::CondensedOpSharding, shape::Dims{N}, device_ids
 ) where {N}
     if sharding.type == OpShardingType.Replicated || sharding.type == OpShardingType.Maximal
-        return ntuple(Returns(ntuple(i -> 1:shape[i], N)), length(device_ids)), false
+        return [ntuple(i -> 1:shape[i], N) for _ in 1:length(device_ids)], false
```
wsmoses, line 254: instead of the anonymous function could we do `Base.Fix1(Base.getindex, shape)`?

avik-pal:
@wsmoses is this good to go from your end?

wsmoses (approved):

PR #828, opened 2025-03-01
feat: High-Level IFRT dispatches

avik-pal:
- [x] add sharding
    - [x] sharding tests
- [x] get non-sharding tests to pass
- [x] add ifrt to buildkite runners
- [x] distributed workflow
- [x] update precompile workflow
- [x] reenable precompile workflow


.github/workflows/CI.yml
```diff
@@ -51,21 +51,27 @@ jobs:
           - core
           - neural_networks
           - integration
+        runtime:
+          - "PJRT"
+          - "IFRT"
```
giordano, line 56: Do I understand correctly this is duplicating the (multi-hour) build jobs, but the runtime is only relevant for the tests on the julia side?

avik-pal:
Locally even distributed works :tada:

avik-pal:
One thing that doesn't work is IFRT-PJRT doesn't have a proper interface to broadcast data to all processes. Currently we just fill in the ones that is addressable. With the Proxy backend we should be able to fix this

avik-pal:
not sure why the arm ones are failing more consistently than before

avik-pal:
I am disabling aarch64 precompilation for now

giordano:
> I am disabling aarch64 precompilation for now

That'd affect Alps.

wsmoses:
can we test that locally somewhere with a debug jll to see what's happening

avik-pal:
> can we test that locally somewhere with a debug jll to see what's happening

I have re-enabled the pre-compilation. But it seems like the failure is independent of the IFRT work (https://github.com/EnzymeAD/Reactant.jl/actions/runs/13731962902/job/38410549236#step:8:330). It was just hidden since the tests passed even when precompilation failed. I have updated how the preferences are set to match the previous behavior

wsmoses:
we should clearly still fix that, but can move past that for now

wsmoses (approved):
looks reasonable to me, but i presume you need to separate jll change

avik-pal:
I ended up not needing the jll change since some of the ifrt sharding types don't implement the whole api.

But I kept it since it still wraps useful functionality

PR #835, opened 2025-03-03
feat: overlay `error` with a custom_call

avik-pal:
Still need to register the FFI on the cpp end. Is that the best way to do this or do we have a way of registering function calls on julia end?

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -568,7 +568,7 @@ end
             Tuple(rsize)
         end
     else
-        error("Invalid FFT type: $type")
+        Base.error("Invalid FFT type: $type")
     end
 
     res = MLIR.IR.result(
@@ -841,7 +841,7 @@ end
     sizea = Dict(c => d for (c, d) in zip(ia, size(lhs)))
     sizeb = Dict(c => d for (c, d) in zip(ib, size(rhs)))
     sizes = mergewith(sizea, sizeb) do da, db
-        da == db ? da : error("Invalid dimensions in einsum equation")
+        da == db ? da : Base.error("Invalid dimensions in einsum equation")
     end
 
     rsize = Tuple(sizes[i] for i in ic)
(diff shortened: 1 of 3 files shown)


avik-pal:
though I am not super sure if we should do this by default. There are a lot of usecases for having the runtime error functionality separately but with this all errors will be embedded into mlir by default

avik-pal:
```julia
julia> fn(x) = begin
           Reactant.Ops.throw("dimension mismatch")
           return x .+ 1
       end
fn (generic function with 1 method)

julia> @jit fn(Reactant.to_rarray(rand(2)))
ERROR: INTERNAL: dimension mismatch

Stacktrace:
 [1] reactant_err(msg::Cstring)
   @ Reactant.XLA ~/reactant/Reactant.jl/src/xla/Utils.jl:12
 [2] macro expansion
   @ ~/reactant/Reactant.jl/src/xla/PJRT/LoadedExecutable.jl:175 [inlined]
 [3] execute_sharded
   @ ~/reactant/Reactant.jl/src/xla/PJRT/LoadedExecutable.jl:144 [inlined]
 [4] macro expansion
   @ ~/reactant/Reactant.jl/src/Compiler.jl:1435 [inlined]
 [5] (::Reactant.Compiler.Thunk{typeof(fn), Symbol("##fn_reactant#280"), Tuple{ConcretePJRTArray{…}}, false})(args::ConcretePJRTArray{Float64, 1, 1, Reactant.Sharding.ShardInfo{Reactant.Sharding.NoSharding, Nothing}})
   @ Reactant.Compiler ~/reactant/Reactant.jl/src/Compiler.jl:1690
 [6] top-level scope
   @ ~/reactant/Reactant.jl/src/Compiler.jl:1036
```

wsmoses (approved):

avik-pal:
need to fix cuda before merging this

PR #839, opened 2025-03-03
Fix `Colon(::TracedRNumber, ::TracedRNumber)`

mofeing:
cc @jumerckx


Pangoraw:
Note that the test passes without the fix because `@trace for` does just look for the syntactic variables

mofeing:
exactly, this fixes the construction of `start:end` when both vars are `TracedRNumber`s, which tests weren't checking

mofeing:
okay, so the solution i propose here doesn't really work because `UnitRange{T}` requires `T<:Real`, so `UnitRange{TracedRNumber{Int}}` is not valid because `!(TracedRNumber{Int}<:Real)` (too bad Julia is not covariant)

Issue #755, opened 2025-02-16
`mod` is JIT-ed to HLO operator with the semantic of Julia's `rem`


giordano:
> The fact that `Base.mod` and `Base.rem` both are lowered to the same HLO operator looks very fishy since they have different semantics in Julia.
> 
> https://openxla.org/xla/operation_semantics#element-wise_binary_arithmetic_operations
> 
> > When `Op` is `Rem`, the sign of the result is taken from the dividend, and the absolute value of the result is always less than the divisor's absolute value.
> 
> This is the semantic of Julia's [`rem`](https://docs.julialang.org/en/v1/base/math/#Base.rem), the opposite of [`mod`](https://docs.julialang.org/en/v1/base/math/#Base.mod).  See also https://openxla.org/stablehlo/spec#remainder for more details about the operator.
> 

_Originally posted by @giordano in https://github.com/EnzymeAD/Reactant.jl/pull/754#discussion_r1957311768_

giordano:
A possible implementation of `mod` could be
```julia
# Based on https://github.com/JuliaLang/julia/blob/39255d47db7657950ff1c82137ecec5a70bae622/base/float.jl#L608-L617
function Base.mod(@nospecialize(x::Reactant.TracedRNumber{T}), @nospecialize(y::Reactant.TracedRNumber{T})) where {T}
    r = rem(x, y)
    return ifelse(r == 0, copysign(r,y), ifelse((r > 0) ⊻ (y > 0), r + y, r))
end
```
but we don't have a method for `copysign`.  How do I find the HLO function for this operation?  Google is failing me.

mofeing:
https://openxla.org/stablehlo/spec#sign

giordano:
That's not exactly the same thing.

giordano:
You can implement `copysign(a, b) = sign(b) * abs(a)`, but this is likely going to be much less efficient than a proper builtin implementation.  LLVM has a [specific intrinsic](https://llvm.org/docs/LangRef.html#llvm-copysign-intrinsic) for example.

mofeing:
yep, but stablehlo has a limited amount of ops, and the only op close to it is that one. not even CHLO has one like this https://github.com/openxla/stablehlo/blob/main/stablehlo/dialect/ChloOps.td

remember that this are very high-level ops, so probably XLA can optimize it if we express well enough.

giordano:
> You can implement `copysign(a, b) = sign(b) * abs(a)`

[...]

PR #840, opened 2025-03-03
adding reduce to Ops.jl

tharittk:
I deliberately did not test on non-commutative operators—it will fail when tested against Julia's reduce. I read the discussion on the internet that it is supposed to be so, i.e., reduce should take a commutative operator; otherwise, the result will be non-deterministic, especially in a parallel system.

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -1483,7 +1483,7 @@ julia> Reactant.@jit(
                     MLIR.IR.Attribute("private"),
                 )
 
-                # Change function name
+                # Change function nane
                 MLIR.IR.attr!(op, symbol_attr_name, MLIR.IR.Attribute(new_name))
             end
         end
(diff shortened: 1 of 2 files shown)


giordano (commented):
> I deliberately did not test on non-commutative operators—it will fail when tested against Julia's reduce.

Does that mean that the `reduce` method introduced here has different semantics than Julia's `reduce`?  That'd be unideal, asi it can lead to hard-to-track bugs when people write generic code and Reactant suddenly changes their meaning (similar for example to #755)

src/Ops.jl
```diff
@@ -1483,7 +1483,7 @@ julia> Reactant.@jit(
                     MLIR.IR.Attribute("private"),
                 )
 
-                # Change function name
+                # Change function nane
```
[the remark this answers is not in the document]
glwagner: image

mofeing:
one thing to note is that methods in `Ops` should replicate the semantics of StableHLO (and possibly other MLIR dialects), not Julia.

the adaptation from StableHLO to Julia semantics is done in the method specializations outside the `Ops` module.

giordano:
> one thing to note is that methods in `Ops` should replicate the semantics of StableHLO (and possibly other MLIR dialects), not Julia.

Ah, yes, that's a good point.  This is not extending `Base.reduce`, so that's fine (although the same name is slightly confusing :sweat_smile:).

src/Ops.jl
avik-pal, line 1486:
```suggestion
                # Change function name
```

src/Ops.jl
```diff
@@ -2313,4 +2313,50 @@ Produces a [`Reactant.MLIR.Dialects.sdy.sharding_constraint`](@ref) operation wi
     end
 end
 
+@noinline function reduce(
```
avik-pal, line 2316: Can you follow up with a PR to update the mapreduce impl to call this function

tharittk: Sure I'll do that

avik-pal:
the cuda reduce tests are failing

tharittk:
Based on my initial investigation, It seems to be about how the init_values is handled. In case of CPU - via Reactant.set_default_backend("cpu"), there is no issue.

But when using GPU, if the init_values is set to be 1 for * and 0 for +, for example,  it will be OK. Other than that, it looks like the GPU applies init_val multiple times.

tharittk:
Ok I was not reading the specification carefully. Here what it says
```
Semantics
Applies a reduction function body to inputs and init_values along the dimensions and produces results tensors.
The order of reductions is implementation-defined, which means that body and init_values must form a monoid to guarantee that the operation produces the same results for all inputs on all implementations. However, this condition doesn't hold for many popular reductions. E.g. floating-point addition for body and zero for init_values don't actually form a monoid because floating-point addition is not associative.
```

So it is the word ```monoid```. I honestly google that word and to my understanding it means the operation must be commutative and the ```init_values``` has to be the identity of that operation to guarantee the same result across all implementation.

giordano:
>  Here what it says

For the benefit of readers, this is the source: https://openxla.org/stablehlo/spec#semantics_71

avik-pal:
Run JuliaFormatter on the code to fix the formatter ci

tharittk:
hmm.. the latest update from upstream branch seems to break this, specifically for macOS. Anyone has any suggestion?

src/Ops.jl
```diff
@@ -2313,4 +2313,98 @@ Produces a [`Reactant.MLIR.Dialects.sdy.sharding_constraint`](@ref) operation wi
...
+
+    func =
+        Reactant.TracedUtils.make_mlir_fn(
+            fn,
+            (sample_inputs),
+            (),
+            "reduce_fn";
+            args_in_result=:none,
+            return_dialect=:stablehlo,
+        ).f
```
avik-pal, line 2379:
```suggestion
    sample_inputs = [
        Reactant.TracedUtils.promote_to(TracedRNumber{T}, 0),
        Reactant.TracedUtils.promote_to(TracedRNumber{T}, 0)
    ]

    func =
        Reactant.TracedUtils.make_mlir_fn(
            fn,
            (sample_inputs),
            (),
            "reduce_fn",
            false;
            args_in_result=:none,
            return_dialect=:stablehlo,
        ).f
```

avik-pal (approved):

PR #844, opened 2025-03-04
ReactantExtra changes to fix MPI PR

mofeing:
This PR fixes the registration of the LLVM dialect, and adds a lil extra utility for parsing single MLIR ops. The only limitation is that since we don't know the SSA names in advance, it fails when we need pass a argument, but it's still useful for writing some simple `func.func`s that need to be injected in the module.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -33,6 +33,7 @@
 #include "mlir/InitAllPasses.h"
 #include "mlir/Pass/PassRegistry.h"
 #include "mlir/Transforms/Passes.h"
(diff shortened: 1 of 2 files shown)


giordano (commented):
I had a look because my review was requested and it seems ok as far as I can tell, but I'm not very familiar with the xla API, so I'm really sure my opinion matters much here.

mofeing:
yeah, i just mentioned you for the build process

wsmoses (approved):

wsmoses:
For fun can you also bump enzymexla for the next jll too

mofeing:
done

wsmoses (approved):

PR #853, opened 2025-03-07
Simplify `Thunk` generated function

glou-nes:
@wsmoses I don't know if it's alright. Is that a good starting point for serialization?


wsmoses:
so this PR kind of does the opposite of what we want to do. We want to move expressions into the place you moved it from (though we do want to add the closure arg as you have)

PR #856, opened 2025-03-08
Fix execute ir pointer

wsmoses:


wsmoses:
@avik-pal after this bugfix goes in we should probably fix the calling conv of the other ones to be a similar llvmcall

src/xla/XLA.jl
```diff
@@ -114,6 +115,23 @@ function __init__()
...
+        for name in ("XLAExecute", "XLAExecuteSharded", "ifrt_loaded_executable_execute")
+            ptr = Libdl.dlsym(Reactant_jll.libReactantExtra_handle, name)
+            Enzyme.LLVM.define(
+                jd_main,
+                Enzyme.Compiler.JIT.absolute_symbol_materialization(
+                    Enzyme.LLVM.mangle(lljit, name), ptr
+                ),
+            )
+        end
+    end
```
vchuravy, line 133: This a better solution than doing RTLD_GLOBAL.

vchuravy: Or use https://github.com/JuliaLLVM/LLVM.jl/blob/8757e8a438bb5a67fb27536923954abb6cae6d55/src/orc.jl#L178C10-L178C55 to make all symbols findable.

giordano: You mean that with this we don't actually need https://github.com/JuliaPackaging/Yggdrasil/pull/10706?  I'd be happy to revert that change 🙂

wsmoses: No this is still required even with that (from local testing).

Specifically loading a precompiled .so fails with just this and is fixed by the global

wsmoses: I’ll also note here that we don’t use the JIT from llvm.jl. The exclusive purpose of this is such that these symbols are findable from Julia’s JIT

vchuravy: What is the concrete error when loading a plgimage?

wsmoses: Will get when by computer, but cannot find symbol XLAExecuteSharded when loading GordonBell2025/…/.so when doing https://github.com/PRONTOLab/GB-25/issues/31

wsmoses: Note that this code without the global works normally without the precompiled code done in the linked issue

wsmoses:
```
/home/wmoses/git/Enzyme.jl/julia11/julia: symbol lookup error: /home/wmoses/.julia/compiled/v1.11/GordonBell25/qFJ42_9Shjc.so: undefined symbol: XLAExecuteSharded
```

giordano:
Besides x86_64 macOS failures due to #867, all integration tests are failing on Linux (which is affected by making all symbols global): https://github.com/EnzymeAD/Reactant.jl/actions/runs/13740137086/job/38428385024?pr=856.  Sounds like we really need a solution to avoid that.

vchuravy:
> Sounds like we really need a solution to avoid that.

I have an idea.

Project.toml
```diff
@@ -84,7 +84,7 @@ PythonCall = "0.9"
 Random = "1.10"
 Random123 = "1.7"
 ReactantCore = "0.1.5"
-Reactant_jll = "0.0.83"
+Reactant_jll = "0.0.84"
```
giordano, line 87: We should probably test with the old version (without rtld_global)
```suggestion
Reactant_jll = "0.0.83"
```

giordano:
On the merge commit of this PR: https://github.com/EnzymeAD/Reactant.jl/actions/runs/13764654846/job/38488238961#step:9:539
```
findfirst / findlast: Error During Test at /home/runner/work/Reactant.jl/Reactant.jl/test/sorting.jl:182
  Test threw exception
  Expression: ffirstlinindices(x) == #= /home/runner/work/Reactant.jl/Reactant.jl/test/sorting.jl:182 =# @jit(findfirst(x_ra))
  ArgumentError: invalid index: nothing of type Nothing
  Stacktrace:
    [1] to_index(i::Nothing)
      @ Base ./indices.jl:315
    [2] to_index(A::LinearIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}, i::Nothing)
      @ Base ./indices.jl:292
    [3] to_indices
      @ ./indices.jl:368 [inlined]
    [4] to_indices
      @ ./indices.jl:360 [inlined]
    [5] getindex
      @ ./abstractarray.jl:1312 [inlined]
    [6] (::Main.var"##Sorting#243".var"#ffirstlinindices#36")(x::Matrix{Bool})
      @ Main.var"##Sorting#243" ~/work/Reactant.jl/Reactant.jl/test/sorting.jl:177
    [7] macro expansion
      @ /opt/hostedtoolcache/julia/1.11.3/x64/share/julia/stdlib/v1.11/Test/src/Test.jl:676 [inlined]
    [8] macro expansion
      @ ~/work/Reactant.jl/Reactant.jl/test/sorting.jl:182 [inlined]
    [9] macro expansion
      @ /opt/hostedtoolcache/julia/1.11.3/x64/share/julia/stdlib/v1.11/Test/src/Test.jl:1704 [inlined]
   [10] top-level scope
      @ ~/work/Reactant.jl/Reactant.jl/test/sorting.jl:174
```
I've never seen it before, is that new with this change?

wsmoses:
odd....in any case this is resolved by the other pr (which was just merged), so closing this

giordano:
Sorry, I posted in the wrong PR, it was meant to be in https://github.com/EnzymeAD/Reactant.jl/pull/869#issuecomment-2711025847

PR #858, opened 2025-03-08
Shorten mapreduce by using Ops.reduce

tharittk:
From PR #840, the mapreduce() can be changed to call the Ops.reduce(). 
I found that I have to change the function signature by taking the `@nonspecialize(f)` out and substitute with (somewhat ugly?) `f::F` to force the Julia's compiler to recompile this function every time the new anonymous function for map is introduced to the function call i.e., what is in the test cases. Otherwise, there will be the world age issue. 

I'm not so sure where to put the test in, so put it in `ops.jl`

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -462,103 +462,36 @@ for (jlop, hloop, hlocomp, merge) in
 end
 
 function Base.mapreduce(
-    @nospecialize(f),
+    f::F, # type parameter so that it recompiles for a different (anonymous) function
     @nospecialize(op),
     @nospecialize(A::AnyTracedRArray{T,N});
     dims=:,
     init=nothing,
-) where {T,N}
+) where {T,N,F}
     A = materialize_traced_array(A)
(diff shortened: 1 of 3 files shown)


tharittk:
Looks like CI fails badly when integrating to mapreduce. I have to work more on it.

src/TracedRArray.jl
```diff
@@ -462,103 +462,36 @@ for (jlop, hloop, hlocomp, merge) in
...
-    if dims != (:)
-        red = Ops.reshape(TracedRArray(red), toonedims...)
-    else
-        if length(outdims) == 0
-            red = TracedRNumber{redT}((), red)
-        else
-            red = TracedRArray{redT,length(outdims)}((), red, (outdims...,))
-        end
-    end
-    return red
```
avik-pal, line 561: these checks need to be preserved to ensure that the final result matches julia semantics

avik-pal:
Handled this in a different PR

PR #859, opened 2025-03-08
missed device in expr

glou-nes:
Following #855, We should add this one too. What are the next steps for serialization?

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -1136,6 +1136,7 @@ function codegen_flatten!(
 )
     flatten_names = Symbol[]
     flatten_code = Expr[]
+    devices = []
 
     if is_sharded
         inv_seen_args = Reactant.OrderedIdDict()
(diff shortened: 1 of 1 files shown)


src/Compiler.jl
```diff
@@ -1220,7 +1224,7 @@ function codegen_flatten!(
     is_sharded &&
         (flatten_names = vcat(eachrow(reshape(flatten_names, length(mesh), :))...))
 
-    return flatten_names, flatten_code
+    return flatten_names, flatten_code, devices
```
wsmoses, line 1227: Make this into a tuple not an array

src/Compiler.jl
```diff
@@ -1665,6 +1666,7 @@ struct Thunk{FTy,tag,IsClosure,ArgTypes,ExecTy,DeviceTy}
     f::FTy
     exec::ExecTy
     device::DeviceTy
+    devices
```
wsmoses: This needs to be specialized and not an any

PR #862, opened 2025-03-08
Throw an error when generating a large constant

giordano:


src/Ops.jl
```diff
@@ -71,6 +71,10 @@ end
 @noinline function constant(
     x::DenseArray{T,N}; location=mlir_stacktrace("constant", @__FILE__, @__LINE__)
 ) where {T,N}
+    sizeof(x) > 100 << 20 || error("""
```
wsmoses, line 74: If you can make the threshold a user tunable const Reference, I’m fine merging this even

giordano: Done, but throwing an error here seems to make everything very angry 😄

src/Ops.jl
```diff
@@ -71,6 +72,10 @@ end
 @noinline function constant(
     x::DenseArray{T,N}; location=mlir_stacktrace("constant", @__FILE__, @__LINE__)
 ) where {T,N}
+    sizeof(x) > LARGE_CONSTANT_THRESHOLD[] || error("""
```
wsmoses, line 75: Probably you want lebngth not sizeof

giordano: `sizeof` gives the size in bytes, `length` would be type-dependent, I thought the former would be more interesting.

wsmoses: Yeah fair enough, from other projects I just have too many eltypes that may not have a defined size, but here we need to have that

wsmoses (approved):

PR #864, opened 2025-03-09
Revert "Update jll"

giordano:
Reverts EnzymeAD/Reactant.jl#852

I want to see if #852 is the cause of the increased rate of failure of macOS jobs, although the funny thing is that CI was green in the PR itself


giordano:
Failing also here.  Sigh.

PR #865, opened 2025-03-09
Constant cse

wsmoses:


src/Ops.jl
```diff
@@ -67,14 +67,56 @@ struct Token
     mlir_data::MLIR.IR.Value
 end
 
+function activate_constant_context!(blk::MLIR.IR.Block)
+    stack = get!(task_local_storage(), :entry_block) do
+        return Tuple{MLIR.IR.Block,Dict{MLIR.IR.Attribute,TracedRArray}}[]
```
giordano, line 72: Doesn't this dictionary risk to blow up in a case like https://github.com/PRONTOLab/GB-25/pull/28 where we have 150k such constants?

wsmoses: it's based off of the number of unique constants within the current function. We already need to create that constant in the current mlir module, so it is strictly less than the current list of operations in the block

wsmoses: doesn't mean that it's not a large number, but it doesnt change any asymptotics

Issue #867, opened 2025-03-09
x86_64 macOS jobs failing


giordano:
[Since](https://github.com/EnzymeAD/Reactant.jl/actions/workflows/CI.yml?query=branch%3Amain) the merge commit of #852, x86_64 macOS jobs are suddenly [failing with](https://github.com/EnzymeAD/Reactant.jl/actions/runs/13726746535/job/38394887372#step:8:485)
```
libc++abi: terminating due to uncaught exception of type std::__1::system_error: recursive_mutex lock failed: Invalid argument

[6132] signal 6: Abort trap: 6
in expression starting at /Users/runner/work/Reactant.jl/Reactant.jl/test/ifrt/low_level.jl:10
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
pthread_kill at /usr/lib/system/libsystem_pthread.dylib (unknown line)
abort at /usr/lib/system/libsystem_c.dylib (unknown line)
abort_message at /usr/lib/libc++abi.dylib (unknown line)
demangling_terminate_handler() at /usr/lib/libc++abi.dylib (unknown line)
_objc_terminate() at /usr/lib/libobjc.A.dylib (unknown line)
std::__terminate(void (*)()) at /usr/lib/libc++abi.dylib (unknown line)
std::terminate() at /usr/lib/libc++abi.dylib (unknown line)
__clang_call_terminate at /Users/runner/.julia/artifacts/863461cadb4a5a3845fab7ad02bcca44bc89faf8/lib/libReactantExtra.dylib (unknown line)
llvm::orc::MaterializationResponsibility::~MaterializationResponsibility() at /Users/runner/.julia/artifacts/863461cadb4a5a3845fab7ad02bcca44bc89faf8/lib/libReactantExtra.dylib (unknown line)
std::__1::__shared_ptr_pointer<llvm::orc::MaterializationResponsibility*, std::__1::default_delete<llvm::orc::MaterializationResponsibility>, std::__1::allocator<llvm::orc::MaterializationResponsibility>>::__on_zero_shared() at /Users/runner/.julia/artifacts/863461cadb4a5a3845fab7ad02bcca44bc89faf8/lib/libReactantExtra.dylib (unknown line)
[...]

PR #868, opened 2025-03-09
[CI] Use debug build for x86_64 macOS jobs

giordano:
Try to debug #867.


giordano:
Now I need to go, but if someone wanted to have a look at this I set up debug with tmate.  Once you log in, you can run
```
julia --color=yes --code-coverage=user --depwarn=yes --check-bounds=yes --warn-overwrite=yes --inline=yes --startup-file=no --project=test -e 'using Pkg; Pkg.instantiate()'
lldb -- julia --color=yes --code-coverage=user --depwarn=yes --check-bounds=yes --warn-overwrite=yes --inline=yes --startup-file=no --project=test test/integration/special_functions.jl
```
and you should be good to go.  It doesn't always crash at the same point, occasionally you may have to run the test script a few times to trigger the crash.

PR #870, opened 2025-03-09
[CI] Disable failing macOS-13 jobs

giordano:
Ref: https://github.com/EnzymeAD/Reactant.jl/issues/867.

I'm not particularly happy about this, but short of having someone to fix the issue above in this way we can have green CI back.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -56,6 +56,13 @@ jobs:
           - "IFRT"
         assertions:
           - false
+        exclude:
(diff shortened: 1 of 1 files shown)


wsmoses (approved):
I'm okay with this for now, and presume you can make an issue to remind ourselves to come back to this post GB

giordano:
There's #867 already (also referenced in the comment in the code) 🙂

wsmoses:
ah got it!

PR #876, opened 2025-03-11
fix: auto-compile fill of concretearray

avik-pal:


avik-pal:
should resolve https://github.com/PRONTOLab/GB-25/issues/32#issuecomment-2711630979

jumerckx (approved):
LGTM as well.
This fix reveals the next error for the Oceananigans code, namely `permutedims!`, and after that `map!`.
I'll clean up my code for those and make a follow up pr.
My fixes on top of these still don't make the code work though, I'm starting to think it might be easier to force building of the simulation grid on the cpu.

PR #880, opened 2025-03-11
`mapreducedim!` for concreterarrays

jumerckx:
ref: https://github.com/EnzymeAD/Reactant.jl/pull/876#pullrequestreview-2674489437
ref: https://github.com/PRONTOLab/GB-25/issues/32

### DIFF
--- a/src/ConcreteRArray.jl
+++ b/src/ConcreteRArray.jl
@@ -448,3 +448,9 @@ function Base.fill!(x::Union{AnyConcreteIFRTArray,AnyConcretePJRTArray}, val)
     fn(x, val, idxs...)
     return x
 end
+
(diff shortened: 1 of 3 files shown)


src/ConcreteRArray.jl
```diff
@@ -448,3 +448,19 @@ function Base.fill!(x::Union{AnyConcreteIFRTArray,AnyConcretePJRTArray}, val)
     fn(x, val, idxs...)
     return x
 end
+
+function mymapreducedim!(f, op, R, A)
```
jumerckx, line 452: I'm not sure why this weird little dance with this indirection is needed, but it is needed as it doesn't work by just `compile`ing `Base.mapreducedim!` directly...

avik-pal (approved):

PR #885, opened 2025-03-12
Add backend tls

wsmoses:
@giordano should fix https://github.com/EnzymeAD/Reactant.jl/pull/883

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -295,7 +295,7 @@ function ka_with_reactant(ndrange, workgroupsize, obj, args...)
 
     # figure out the optimal workgroupsize automatically
     if KA.workgroupsize(obj) <: KA.DynamicSize && workgroupsize === nothing
-        if !Reactant.Compiler.PartitionKA[] || raising()
+        if !Reactant.Compiler.PartitionKA[] || raising() || backend() in ("cpu", "tpu")
             threads = prod(ndrange)
         else
             config = CUDA.launch_configuration(kernel.fun; max_threads=prod(ndrange))
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -672,6 +672,38 @@ function raising!(f, is_raising::Bool)
     end
 end
 
+function activate_backend!(backend::String)
+    stack = get!(task_local_storage(), :reactant_backend) do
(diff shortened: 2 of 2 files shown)


ext/ReactantCUDAExt.jl
```diff
@@ -295,7 +295,7 @@ function ka_with_reactant(ndrange, workgroupsize, obj, args...)
 
     # figure out the optimal workgroupsize automatically
     if KA.workgroupsize(obj) <: KA.DynamicSize && workgroupsize === nothing
-        if !Reactant.Compiler.PartitionKA[] || raising()
+        if !Reactant.Compiler.PartitionKA[] || raising() || backend() in ("cpu", "tpu")
```
giordano, line 298: So, this `backend` ehre is `ReactantKernelAbstractionsExt.ReactantBackend`, so both calling it and comparing it with strings like `"cpu"` and `"tpu"` is wrong.  Need to figure out what should be the right backend to check

giordano: Should it be this backend: https://github.com/EnzymeAD/Reactant.jl/blob/ff58e26f484e6817ace53db0e1488e8e3ebe8955/src/Compiler.jl#L558 ?

src/Compiler.jl
```diff
@@ -672,6 +672,38 @@ function raising!(f, is_raising::Bool)
...
+        error("Deactivating wrong Reactant backend context")
+    return pop!(task_local_storage(key))
+end
+
+function backend(; throw_error::Bool=true)
+    key = :reactant_backend
+    if !(haskey(task_local_storage(), key) && !Base.isempty(task_local_storage(key)))
+        throw_error && error("No Reactant backend context")
+    end
+    return last(task_local_storage(key)::Vector{Bool})
```
giordano, line 695:
```suggestion
    return last(task_local_storage(key)::Vector{String})
```
?

ext/ReactantCUDAExt.jl
giordano:
```suggestion
        if !Reactant.Compiler.PartitionKA[] || raising() || Reactant.Compiler.backend() in ("cpu", "tpu")
```

giordano (approved):

PR #887, opened 2025-03-12
Revert "fix: apply init values after reduction"

giordano:
Reverts EnzymeAD/Reactant.jl#881.  I believe this broke the GB-25 code.


wsmoses (approved):
Can you open an issue of relevance to track

giordano:
https://github.com/PRONTOLab/GB-25/pull/46 is running fine, which confirms #881 was indeed buggy. I also bumped the version so that we can tag it and have back a working state.

PR #890, opened 2025-03-12
Emit debug info on big `Ops.constant` values

mofeing:
@giordano with this PR, you just need to configure the `LARGE_CONSTANT_THRESHOLD` and it will emit the full backtrace as the MLIR location in the debug info of the `stablehlo.constant` op.

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -89,8 +89,12 @@ end
 @noinline function constant(
     x::DenseArray{T,N}; location=mlir_stacktrace("constant", @__FILE__, @__LINE__)
 ) where {T,N}
-    sizeof(x) > LARGE_CONSTANT_THRESHOLD[] &&
-        error("Generating a constant larger than $(LARGE_CONSTANT_THRESHOLD[]) bytes.")
+    if sizeof(x) > LARGE_CONSTANT_THRESHOLD[]
+        location = with_debug() do
+            mlir_stacktrace("constant", @__FILE__, @__LINE__)
+        end
+    end
+
     value = MLIR.IR.DenseElementsAttribute(x)
     constants = constant_context()[2]
     if haskey(constants, value)


src/Ops.jl
```diff
@@ -89,8 +89,12 @@ end
 @noinline function constant(
     x::DenseArray{T,N}; location=mlir_stacktrace("constant", @__FILE__, @__LINE__)
 ) where {T,N}
-    sizeof(x) > LARGE_CONSTANT_THRESHOLD[] &&
-        error("Generating a constant larger than $(LARGE_CONSTANT_THRESHOLD[]) bytes.")
+    if sizeof(x) > LARGE_CONSTANT_THRESHOLD[]
```
wsmoses, line 92: we might want to have two constants. early erroring is also useful from julia side, in addition to the location one

mofeing: the second constant should be a second threshold above which to error? or should be a on/off flag to either error or emit backtrace?

wsmoses: either is fine with me, cc @giordano

mofeing: just what i added

src/Ops.jl
```diff
@@ -36,6 +36,7 @@ end
 
 const DEBUG_MODE::Ref{Bool} = Ref(false)
 const LARGE_CONSTANT_THRESHOLD = Ref(100 << 20) # 100 MiB
+const LARGE_CONSTANT_RAISE_ERROR = Ref(false)
```
mofeing, line 39:
```suggestion
const LARGE_CONSTANT_RAISE_ERROR = Ref(true)
```

Pangoraw:
It looks like you would need to update that test:


https://github.com/EnzymeAD/Reactant.jl/blob/31a375c023ba1fff121787bda0d392336c893255/test/ops.jl#L1124-L1144

Pangoraw:
Is this ready?

giordano:
Failing tests are broken by this PR, so probably not yet.

src/Ops.jl
```diff
@@ -89,8 +90,16 @@ end
 @noinline function constant(
     x::DenseArray{T,N}; location=mlir_stacktrace("constant", @__FILE__, @__LINE__)
 ) where {T,N}
-    sizeof(x) > LARGE_CONSTANT_THRESHOLD[] &&
-        error("Generating a constant larger than $(LARGE_CONSTANT_THRESHOLD[]) bytes.")
+    if sizeof(x) > LARGE_CONSTANT_THRESHOLD[]
+        if LARGE_CONSTANT_RAISE_ERROR[]
+            error("Large constant array of size $(sizeof(x))")
```
giordano, line 95: This message should really have a unit (bytes)

wsmoses: I mean honestly we could probably take most of the old error message (and perhaps also add the sizeof to it

wsmoses:
bumping this

test/ops.jl
```diff
@@ -1129,14 +1129,18 @@ end
...
     default_threshold = Ops.LARGE_CONSTANT_THRESHOLD[]
+    default_raise_error = Ops.LARGE_CONSTANT_RAISE_ERROR[]
     try
         Ops.LARGE_CONSTANT_THRESHOLD[] = N
+        Ops.LARGE_CONSTANT_RAISE_ERROR[] = true
         @compile f!(vr)
     catch err
-        @test err.msg == "Generating a constant larger than $(N) bytes."
+        @test err.msg ==
+            "Generating a constant of 16 bytes, which larger than the 5 bytes threshold"
```
giordano, line 1139:
```suggestion
            "Generating a constant of 40 bytes, which larger than the 5 bytes threshold"
```

test/ops.jl
```diff
@@ -1129,14 +1129,18 @@ end
...
     default_threshold = Ops.LARGE_CONSTANT_THRESHOLD[]
+    default_raise_error = Ops.LARGE_CONSTANT_RAISE_ERROR[]
     try
         Ops.LARGE_CONSTANT_THRESHOLD[] = N
+        Ops.LARGE_CONSTANT_RAISE_ERROR[] = true
         @compile f!(vr)
     catch err
-        @test err.msg == "Generating a constant larger than $(N) bytes."
+        @test err.msg ==
+            "Generating a constant of 40 bytes, which larger than the 5 bytes threshold"
```
giordano, line 1139:
```suggestion
            "Generating a constant of 40 bytes, which larger than the $(N) bytes threshold"
```

Issue #889, opened 2025-03-12
Reduction PR #881 breaks Oceananigans


giordano:
#881 broke reductions inside Oceananigans.

https://github.com/PRONTOLab/GB-25/pull/45#issuecomment-2718514600
```
ERROR: LoadError: InexactError: Int64(Inf)
Stacktrace:
  [1] Int64
    @ ./float.jl:994 [inlined]
  [2] convert
    @ ./number.jl:7 [inlined]
  [3] _round_convert
    @ ./rounding.jl:480 [inlined]
  [4] round
    @ ./rounding.jl:479 [inlined]
  [5] ceil(::Type{Int64}, x::Float64)
    @ Base ./rounding.jl:476
  [6] split_explicit_substepping
```

Code in Oceananigans:
https://github.com/CliMA/Oceananigans.jl/blob/875a06a0964115c7f6ea1ca14c12ede5272d5ab1/src/Fields/field.jl#L671
https://github.com/CliMA/Oceananigans.jl/blob/875a06a0964115c7f6ea1ca14c12ede5272d5ab1/src/Fields/field.jl#L620
https://github.com/CliMA/Oceananigans.jl/blob/875a06a0964115c7f6ea1ca14c12ede5272d5ab1/src/Fields/field.jl#L715

simone-silvestri:
This is a MWE that shows the difference in behavior between a `ReactantState()` architecture and a `CPU()` architecture
```julia
using Reactant, Oceananigans

grid = RectilinearGrid(CPU(), size=(2, 4, 8), extent=(1, 1, 1))
c = CenterField(grid)
set!(c, 1)
r = Field{Nothing, Nothing, Nothing}(grid)
c = Oceananigans.Fields.condition_operand(c, nothing, Inf)

Base.mapfirst!(identity, interior(r), interior(c))
@show interior(r)

grid = RectilinearGrid(Oceananigans.Architectures.ReactantState(), size=(2, 4, 8), extent=(1, 1, 1))
rc = CenterField(grid)
set!(rc, 1)
rr = Field{Nothing, Nothing, Nothing}(grid)
rc = Oceananigans.Fields.condition_operand(c, nothing, Inf)

Base.mapfirst!(identity, interior(rr), interior(rc))
@show interior(rr)
```
In particular `interior(r)` has, correctly, a value of 1, while `interior(rr)` is not being initialized correctly.

wsmoses:
[...]

PR #892, opened 2025-03-12
fix: `map!` and `map`

avik-pal:
https://github.com/EnzymeAD/Reactant.jl/pull/882/ but without https://github.com/EnzymeAD/Reactant.jl/pull/881

### DIFF
--- a/src/ConcreteRArray.jl
+++ b/src/ConcreteRArray.jl
@@ -457,8 +457,13 @@ function Base.mapreducedim!(
     return R
 end
 
+function mymap!(f, R, A)
+    map!(f, R, A)
+    return nothing
+end
+
 function Base.map!(f, R::Union{AnyConcreteIFRTArray,AnyConcretePJRTArray}, A::AbstractArray)
-    fn = compile(Base.map!, (f, R, A))
+    fn = compile(mymap!, (f, R, A))
     fn(f, R, A)
     return R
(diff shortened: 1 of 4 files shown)


wsmoses (approved):

avik-pal:
I need to add some tests here

giordano:
This is still crashing https://github.com/PRONTOLab/GB-25/pull/48

giordano:
Simone posted a reproducer: https://github.com/EnzymeAD/Reactant.jl/issues/889#issuecomment-2718729409

avik-pal:
atleast now we know the real source of the issue

avik-pal:
fixes #889

src/ConcreteRArray.jl
```diff
@@ -464,3 +456,14 @@ function Base.mapreducedim!(
     fn(f, op, R, A)
     return R
 end
+
+function mymap!(f, R, A)
```
avik-pal, line 460: same as the `mymapreduce!`, we are definitely doing some incorrect codegen for in place updates, else this change shouldn't matter

avik-pal:
https://github.com/PRONTOLab/GB-25/pull/48 is passing with the latest commit

wsmoses:
mose is off for the night but if it works for you, go for it

PR #897, opened 2025-03-13
or backend == tpu

wsmoses:


src/xla/XLA.jl
```diff
@@ -184,8 +184,9 @@ for runtime in (:PJRT, :IFRT)
         state.default_client = cpu
 
         # Try TPU if possible, then try GPU (CUDA)
+	if !Reactant.precompiling()
```
avik-pal, line 187: This function shouldn't be triggered during precompilation.

wsmoses: sadly it was =/

avik-pal: do you have a repro? We would have to patch that. This function being called during precompilation might lead to some distributed settings being set incorrectly

PR #903, opened 2025-03-13
[docs] Simplify multihost example, no need for extra script

giordano:


docs/src/tutorials/multihost.md
```diff
@@ -43,17 +43,10 @@ will run on multiple devices across multiple nodes.
...
 
 export JULIA_DEBUG="Reactant,Reactant_jll"
-
-srun --preserve-env bash ./matmul.sh
-```
-
-```bash [matmul.sh]
-#!/bin/bash -l
-
 # Important else XLA might hang indefinitely
```
avik-pal, line 46: For some reason doing this in a single script wasn't unsetting the env vars in the compute nodes

PR #911, opened 2025-03-14
Don't jit execute kernel during precompilation

wsmoses:


ext/ReactantKernelAbstractionsExt.jl
```diff
@@ -92,7 +92,11 @@ function tokw(ndrange, workgroupsize, obj, args...)
 end
 
 function (obj::KA.Kernel{ReactantBackend})(args...; ndrange=nothing, workgroupsize=nothing)
-    @jit tokw(ndrange, workgroupsize, obj, args...)
+    if Reactant.precompiling()
+        @code_hlo optimize = false tokw(ndrange, workgroupsize, obj, args...)
+    else
+        @jit tokw(ndrange, workgroupsize, obj, args...)
```
giordano, line 98: Uhm, I agree with the idea in principle of reducing compilation cost during precompilation, but I think we're giving a completely different object in output with this (`Reactant.MLIR.IR.Module` instead of whatever is the output of the `@jit` call)?

PR #896, opened 2025-03-13
fix: apply init values after reduce

avik-pal:
the breakage in GB was likely from https://github.com/EnzymeAD/Reactant.jl/pull/892. Testing in https://github.com/PRONTOLab/GB-25/pull/50

PR #915, opened 2025-03-14
Revert "fix: apply init values after reduce"

giordano:
Reverts EnzymeAD/Reactant.jl#896.  Ref: https://github.com/EnzymeAD/Reactant.jl/pull/896#issuecomment-2725429143

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -391,7 +391,7 @@ end
 end
 
 # shape ops
-function reshape(x::TracedRArray, dims::Integer...; kwargs...)
+function reshape(x::TracedRArray, dims...; kwargs...)
     return reshape(x, collect(dims); kwargs...)
 end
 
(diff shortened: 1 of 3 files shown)


giordano:
@avik-pal looks like this fixed the issue https://github.com/EnzymeAD/Reactant.jl/pull/896#issuecomment-2725352852.  I don't know if that was an issue in Julia v1.10 (this doesn't happen with v1.11), but the problem is that now we have CI systematically red again.

avik-pal:
let's revert in that case, we can revisit this at some later point

giordano:
Uh, #896 was needed for GB-25 though 😞

avik-pal:
Was it? I separated the `map!` stuff that GB needed into a diff PR

giordano:
Uh, I thought that was what https://github.com/EnzymeAD/Reactant.jl/pull/896#issue-2915500519 meant, but maybe I misunderstood.

giordano:
> (this doesn't happen with v1.11)

Oh, that's simply because we don't run the assertions tests with v1.11 🫠  ok, if this wasn't necessary for GB-25 maybe it's good to revert it.

PR #924, opened 2025-03-15
Fix irrational promotion

wsmoses:
cc @alexandrebouchard

I think the missing piece was just still needing to promote the eltype


alexandrebouchard:
Thank you! Only minor question is about why the both promote_rule(::X, ::Y) and promote_rule(::Y, ::X) is added in this PR? My understanding from the base doc was that the [symmetrization was done for you](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Defining-Promotion-Rules).

wsmoses:
Oh didn’t know that. We do that the same elsewhere, so in case it is needed for something we can do a follow up that removes it for all at once.

PR #940, opened 2025-03-18
Raise without unrolling

wsmoses:


wsmoses:
this is blocked waiting for new jll

giordano:
This is looking good, based on https://github.com/PRONTOLab/GB-25/pull/79#issuecomment-2736741610

wsmoses:
@giordano if you can confirm runtime perf is reasonable by comparison lets merge

wsmoses:
we can still do lazy @btime on outside

giordano:
Uhm, this seems to degrade performance _a lot_: according to profiling information, on A100 3000 iterations take 58s with Reactant v0.2.45 (Reactant_jll v0.0.92), on this branch (and Reactant_jll v0.0.93) they take 4m 15s

Pangoraw:
Yeah without the parallelize pass the raised code will be pretty inefficient.

giordano:
Uhm, maybe the profiler was messing up timing a lot (I got lots of warnings about large overhead), with `@time` I get a smaller difference (but still disfavouring this change): v0.2.95
```julia-repl
julia> @time rloop!(model, ConcreteRNumber(3000));
 57.401839 seconds (926.27 k allocations: 49.752 MiB, 1.75% compilation time)
```
this PR:
```julia-repl
julia> @time rloop!(model, ConcreteRNumber(3000));
 72.907649 seconds (591.45 k allocations: 33.813 MiB, 0.04% gc time, 1.34% compilation time)
```

Pangoraw:
Updating the branch so we can test with parallelization (https://github.com/EnzymeAD/Enzyme-JAX/pull/429).

giordano:
One of the recent enzyme-jax updates actually caused a slowdown analogous to [the one I reported above](https://github.com/EnzymeAD/Reactant.jl/pull/940#issuecomment-2737443450), so that enabling this may not be as a bad performance penalty as it was before (and disabling unrolling at this stage appeared to have significant compile time benefit, which is becoming increasingly important)

PR #942, opened 2025-03-18
Add specialization of copyto! from julia array to concreterarray

jumerckx:


src/ConcreteRArray.jl
```diff
@@ -394,6 +394,10 @@ function Base.copyto!(
     return dest
 end
 
+function Base.copyto!(dest::Union{AnyConcreteIFRTArray,AnyConcretePJRTArray}, src::AbstractArray)
+    copyto!(dest, to_rarray(src))
```
avik-pal, line 398: can you check if `dest .= src` does the job? we might need to `vec(dest)`, `vec(src)` (I don;t remember the exact semantics for copyto!)

wsmoses, line 398: This is a bit dangerous to use for any abstractarray, can we limit ourselves to just a regular array?

Also we can specialize this further where then to_rarray will return a new ConcreteRArray, and instead of calling an underlying method via compile we should be able to just set the inner buffer pointer

glwagner: We need to capture sharding too right?

wsmoses: which I see this latter part id done in 379, so yeah perhaps just edit this to apply to Array?

glwagner, line 398:
```suggestion
    copyto!(dest, to_rarray(src, sharding=to_reactant_sharding(dest.sharding)))
```

Not sure if that is exactly right or somethign like that needed

avik-pal: yeah doing the broadcasting will push it down `Reactant.compile` which will mark the output sharding

glwagner: we may call this outside broadcasting though -- will it work in that case?

wsmoses: cc @avik-pal

avik-pal: [see the comment above](https://github.com/EnzymeAD/Reactant.jl/pull/942#discussion_r2001389327)

wsmoses: I mean here it's fine since we are recursing into the inner function which must have the correct semantics, subject to sharding questions.

avik-pal: this is possible from the beauty of tracking mutations which lets us automatically attach the sharding annotations if we detect a particular output is being stored into a specific sharded array.

avik-pal: > I think? this should be fine if we just get the sharding of the dest (since these must match anyways)

Yes, the only caveat being we will now create 2 IFRT buffers and then rely on GC to free one

wsmoses: yeah but we already need to do this already, right (to_rarray creates a buffer, dest has the existing buffer, and incidentally we'll make a constant buffer within stablehlo.constant)?

avik-pal: The argument is marked donated so it should be reused? (though my understanding of donation is quite flaky and I could be wrong)

src/ConcreteRArray.jl
```diff
@@ -394,6 +394,10 @@ function Base.copyto!(
     return dest
 end
 
+function Base.copyto!(dest::Union{AnyConcreteIFRTArray,AnyConcretePJRTArray}, src::AbstractArray)
```
glwagner, line 397:
```suggestion
function Base.copyto!(dest::Union{AnyConcreteIFRTArray,AnyConcretePJRTArray}, src::Array)
```

src/ConcreteRArray.jl
wsmoses: the output buffer is yes, but to_rarray() creates a buffer. So what I see happening is as follows:

```
some time in the past one allocates dest (array #1)

someone calls copyto!(dest, julia_src)

inside of that we do 
 copyto!(dest, to_rarray(src))
 
That to_rarray allocates a new concreterarray (alloc #2). Let's call this src

eventually we can @compile something. If that takes dst and src as inputs (as it should), we can make dst as donated.

That way we don't need to allocate a third buffer, but we can update it in place. However, we still have two buffers here, and a compile function to do the copy (instead of pointer swap).
```

jumerckx: If I try this, the broadcasting ends up calling `copyto` recursively.
```jl
function mycopyto!(dest, src)
    vec(dest)[axes(vec(src))...] .= vec(src) # use broadcasting instead of copyto!
    return nothing
end
```

```
ERROR: StackOverflowError:
Stacktrace:
     [1] mycopyto!
       @ ~/.julia/dev/Reactant/src/ConcreteRArray.jl:385 [inlined]
     [2] copyto!
       @ ~/.julia/dev/Reactant/src/ConcreteRArray.jl:399 [inlined]
     [3] copyto!
       @ ./broadcast.jl:966 [inlined]
     [4] copyto!
       @ ./broadcast.jl:925 [inlined]
     [5] materialize!
       @ ./broadcast.jl:883 [inlined]
     [6] materialize!(dest::SubArray{…}, bc::Base.Broadcast.Broadcasted{…})
       @ Base.Broadcast ./broadcast.jl:880
--- the above 6 lines are repeated 12989 more times ---
 [77941] mycopyto!
       @ ~/.julia/dev/Reactant/src/ConcreteRArray.jl:385 [inlined]
 [77942] copyto!
       @ ~/.julia/dev/Reactant/src/ConcreteRArray.jl:399 [inlined]
 [77943] copyto!
       @ ./broadcast.jl:966 [inlined]
 [77944] copyto!
       @ ./broadcast.jl:925 [inlined]
 [77945] materialize!
       @ ./broadcast.jl:883 [inlined]
 [77946] materialize!
       @ ./broadcast.jl:880 [inlined]
 [77947] mycopyto!(dest::ConcreteIFRTArray{Float64, 2, Reactant.Sharding.ShardInfo{…}}, src::Matrix{Float64})
       @ Reactant ~/.julia/dev/Reactant/src/ConcreteRArray.jl:385
 [77948] copyto!(dest::ConcreteIFRTArray{Float64, 2, Reactant.Sharding.ShardInfo{…}}, src::Matrix{Float64})
       @ Reactant ~/.julia/dev/Reactant/src/ConcreteRArray.jl:399
```

wsmoses (approved):

Issue #952, opened 2025-03-18
Scalar indexing error in traced subarray of oceanaigans field


wsmoses:
https://pastebin.com/xpUtp9TV

x/ref @avik-pal @glwagner

glwagner:
This might reproduce

```julia
using Reactant
a = Reactant.to_rarray(rand(4, 4))
copyto!(view(a, 2:3, 2:3), rand(2, 2))
```


I get

```julia
julia> using Reactant
┌ Warning: The project dependencies or compat requirements have changed since the manifest was last resolved.
│ It is recommended to `Pkg.resolve()` or consider `Pkg.update()` if necessary.
└ @ Pkg.API ~/.julia/juliaup/julia-1.10.8+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Pkg/src/API.jl:1142
Info Given Reactant was explicitly requested, output will be shown live
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1742320904.700423 13666499 pjrt_client.cc:524] PjRt-IFRT device count: total=1, addressable=1
I0000 00:00:1742320904.700443 13666499 pjrt_client.cc:528] Addressable PjRt-IFRT device: CpuDevice(id=0)
Precompiling Reactant finished.
  1 dependency successfully precompiled in 15 seconds. 52 already precompiled.
  1 dependency had output during precompilation:
┌ Reactant
│  [Output was shown above]
└

julia> a = Reactant.to_rarray(rand(4, 4))
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1742320955.974763 13666295 pjrt_client.cc:524] PjRt-IFRT device count: total=1, addressable=1
I0000 00:00:1742320955.974783 13666295 pjrt_client.cc:528] Addressable PjRt-IFRT device: CpuDevice(id=0)
4×4 ConcreteIFRTArray{Float64,2}:
 0.280439  0.335184  0.616016  0.0920095
 0.414621  0.481342  0.747264  0.303107
 0.152902  0.130549  0.512659  0.483569
 0.895095  0.470901  0.248339  0.80677

julia> copyto!(a, rand(4, 4))
4×4 ConcreteIFRTArray{Float64,2}:
 0.349598  0.443604   0.302546  0.280744
 0.338337  0.80198    0.772586  0.546135
 0.873156  0.935873   0.533483  0.153791
[...]

PR #953, opened 2025-03-18
fix: noop check

avik-pal:
@wsmoses can you check if this fixes https://github.com/EnzymeAD/Reactant.jl/issues/952


giordano:
Can you try the reproducer in https://github.com/EnzymeAD/Reactant.jl/issues/952#issuecomment-2734265215?

wsmoses:
@giordano thats a separate issue, but in any case we’re kind of in a version hell to try. Let’s merge this release, merge the dependent Oceananigans PR etc then we can do some proper testing

wsmoses (approved):

PR #960, opened 2025-03-19
TracedStepRangeLen

jumerckx:
I imagine some code should maybe be moved around a bit?


giordano:
Uhm, the history of this PR contains unrelated commits

jumerckx:
> Uhm, the history of this PR contains unrelated commits

Sorry, should be fixed

jumerckx:
failing CI is all:
```
...
  15864.0 ms  ✓ Reactant → ReactantCUDAExt
   4266.6 ms  ✓ LuxLib → LuxLibReactantExt
   5220.7 ms  ✓ Lux → LuxReactantExt

[2255] signal (2): Interrupt: 2
in expression starting at /Users/runner/work/_temp/8434ebd3-cb2c-4917-ad80-59675f07439b:3
kevent at /usr/lib/system/libsystem_kernel.dylib (unknown line)
unknown function (ip: 0x0)
Allocations: 6396017 (Pool: 6381939; Big: 14078); GC: 10
Error: The action 'Run Tests' has timed out after 60 minutes.
```
Is this due to flaky Github runners or should I attempt debugging this? (@giordano)

giordano:
yes, we've had those hangs in the macOS-latest runners after precompilation in a while.

jumerckx:
> Note: instead of pushing a new commit which restart some 50 new jobs (11 of which on macOS, for which we have 5 runners across the entire organisation), you can simply go to the [summary page of CI run](https://github.com/EnzymeAD/Reactant.jl/actions/runs/13957752795?pr=960) and restart only failed jobs:

I'll keep this in mind in the future! In this case it was a merge commit from main, though, shouldn't CI fully rerun in that case?

src/TracedRNumber.jl
```diff
@@ -460,7 +709,7 @@ function Base.searchsortedfirst(
         !Base.Order.lt(o, f, x),
         1,
         ifelse(
-            h == 0 || Base.Order.lt(o, l, x),
+            (h == 0) | Base.Order.lt(o, l, x),
```
wsmoses, line 712: is this needed

wsmoses (approved):
looks good, maybe add a quick test?

src/TracedRNumber.jl
jumerckx: Yes || does not work for tracedtnumber{bool}

wsmoses: ah right

jumerckx:
Added a test—and fixed a bug in getindex that was unearthed by it.

wsmoses (approved):

jumerckx:
I think this is ready to be merged

PR #961, opened 2025-03-19
[deps] Update to new Enzyme-JAX (which updates to new JAX and Co.)

giordano:
Starting the dance here.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -1715,8 +1715,10 @@ extern "C" bool ifrt_DeviceIsAddressable(ifrt::Device *device) {
   return device->IsAddressable();
 }
 
-tsl::RCReference<ifrt::DeviceList> ifrt_CreateDeviceListFromDevices(
-    ifrt::Client *client, ifrt::Device **device_list, int32_t num_devices) {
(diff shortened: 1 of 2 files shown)


deps/ReactantExtra/API.cpp
```diff
@@ -1715,8 +1715,10 @@ extern "C" bool ifrt_DeviceIsAddressable(ifrt::Device *device) {
   return device->IsAddressable();
 }
 
-tsl::RCReference<ifrt::DeviceList> ifrt_CreateDeviceListFromDevices(
-    ifrt::Client *client, ifrt::Device **device_list, int32_t num_devices) {
+static xla::ifrt::RCReferenceWrapper<ifrt::DeviceList>
```
wsmoses, line 1718: why make this static (that means not exported anymore)

Pangoraw: That's only called from within API.cpp from grepping around.

avik-pal: we don't need devicelist in julia, we just use a `Vector{<:IFRT.Device}` and convert it before giving it to XLA

wsmoses: ah fair nuff

PR #959, opened 2025-03-19
feat: distributed + fix codegen

avik-pal:

PR #964, opened 2025-03-19
feat: if passes run fail, dump to a file

giordano:
Ref: https://github.com/EnzymeAD/Reactant.jl/pull/959#discussion_r2003969079

### DIFF
--- a/src/mlir/IR/Pass.jl
+++ b/src/mlir/IR/Pass.jl
@@ -79,6 +79,10 @@ function run!(pm::PassManager, mod::Module)
         dir = mktempdir()
         path = joinpath(dir, "module.mlir")
         open(path, "w") do io
+            println(io, "// Pass pipeline:")
+            print(io, "// ")
+            print_pass_pipeline(io, OpPassManager(pm))
+            println(io)
             show(IOContext(io, :debug => true), mod)
         end
         @error "Dumped module to " * path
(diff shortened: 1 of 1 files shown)


src/mlir/IR/Pass.jl
```diff
@@ -76,6 +76,12 @@ function run!(pm::PassManager, mod::Module)
         API.mlirPassManagerRun(pm, mod)
     end)
     if isfailure(status)
+        dir = mktempdir()
+        path = joinpath(dir, "module.mlir")
+        open(path, "w") do io
+            show(io, mod)
+        end
+        @error "Dumped module to " * path
         throw("failed to run pass manager on module")
```
giordano, line 85: Continuing from https://github.com/EnzymeAD/Reactant.jl/pull/959/files#r2003996116, how do I print the name of the pass here?  Can I just interpolate `$(pm)` here?

wsmoses: oh right, no you can't (the code we have in compiler.jl run_pass_pipeline or whatnot does have that info)

giordano: Uhm, `PassManager` doesn't seem to have a dedicate `show` method.  So what you want to see (and how to print it)?

avik-pal (approved):

src/mlir/IR/Pass.jl
Pangoraw: You can `show(IR.OpPassManager(passManager))` which should have the pipeline

Pangoraw: https://github.com/EnzymeAD/Reactant.jl/pull/966

wsmoses (approved):

src/mlir/IR/Pass.jl
```diff
@@ -76,6 +76,16 @@ function run!(pm::PassManager, mod::Module)
         API.mlirPassManagerRun(pm, mod)
     end)
     if isfailure(status)
+        dir = mktempdir()
```
giordano, line 79: This directory (with `cleanup=true` by default) isn't very useful: it'll disappear at julia exit, making the MLIR module unusable: https://github.com/PRONTOLab/GB-25/pull/95#issuecomment-2741829963.  I'll address this in a follow up PR.

PR #965, opened 2025-03-19
fix: add Array{<:TracedNumber} to constant

avik-pal:


src/Ops.jl
```diff
@@ -155,6 +155,12 @@ end
     return constant(Base.convert(T, x); kwargs...)
 end
 
+@noinline function constant(
+    x::Array{TracedRNumber{T},N}; location=mlir_stacktrace("constant", @__FILE__, @__LINE__)
```
wsmoses, line 159: why did this come up? I feel like this is more of an issue for where calls constant like this. I'm not opposed, but making constant only apply to abstractarray{<:ReactantPrimitive} is also reasonable

avik-pal: This is OffsetArrays storing that type and the CUDAExt calling adapt_storage which in-turn calls constant.

>  I feel like this is more of an issue for where calls constant like this

Let me stare at the code for sometime and see if I can figure out why OffsetArrays does all the weird stuff it does

avik-pal:
It was bad codegen. Let's not add it. It is more useful to get this as an error.

Even if at somepoint we decide to add this we should print a warning here.

PR #968, opened 2025-03-19
fix: bad codegen

avik-pal:
turns out if you do weird shit, julia crashes


wsmoses (approved):

wsmoses:
12/10 description. No notes

PR #974, opened 2025-03-20
[ReactantExtra] Add new function to dump an Operation to file

giordano:
Absolutely not tested (I'm not even sure `Operation` is the right class here?), but saving the progress before going to sleep.

PS: made in collaboration with @glwagner's best friend.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -1,3 +1,5 @@
+#include <iostream>
+
 #include "mlir-c/IR.h"
 #include "mlir-c/Support.h"
 
@@ -143,6 +145,8 @@
 #include "triton/Dialect/Triton/IR/Dialect.h"
 
 #include "llvm/Support/ExtensibleRTTI.h"
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/raw_ostream.h>
 
 using namespace mlir;
 using namespace llvm;
@@ -2381,6 +2385,18 @@ extern "C" void ifrt_hlo_module_cost_analysis_properties(
   ReactantThrowError(("Cost analysis not supported for this client: " +
                       std::string(client->runtime_type()))
                          .c_str());
+}
+
+extern "C" void dump_operation(Operation *op, const char *filename) {
+  std::error_code EC;
+  llvm::raw_fd_ostream file(filename, EC, llvm::sys::fs::OF_Text);
+
+  if (EC) {
+    std::cerr << "Error opening file: " << EC.message() << std::endl;
+    return;
+  }
+
+  file << *op << "\n";
 }
 
 #pragma endregion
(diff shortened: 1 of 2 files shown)


wsmoses (approved):
cc @ivanradanov

deps/ReactantExtra/API.cpp
```diff
@@ -2383,4 +2387,26 @@ extern "C" void ifrt_hlo_module_cost_analysis_properties(
                          .c_str());
 }
 
+extern "C" void dump_operation(Operation *op, const char *filename) {
+  // Save original stderr
+  FILE *original_stderr = stderr;
+
+  // Redirect stderr to a file
+  FILE *file = freopen(filename, "w", stderr);
```
wsmoses, line 2395: we don't need to redirect to stderr, we can open a file with llvm, and then run out <<  *op

deps/ReactantExtra/API.cpp
```diff
@@ -2383,4 +2387,26 @@ extern "C" void ifrt_hlo_module_cost_analysis_properties(
...
+  if (!file) {
+    std::cerr << "Failed to redirect stderr to file.\n";
+    return;
+  }
+
+  // Call dump()
+  op->dump();
```
Pangoraw, line 2402:
```suggestion
  unwrap(op)->dump();
```

deps/ReactantExtra/API.cpp
```diff
@@ -2383,4 +2387,26 @@ extern "C" void ifrt_hlo_module_cost_analysis_properties(
                          .c_str());
 }
 
+extern "C" void dump_operation(Operation *op, const char *filename) {
```
Pangoraw, line 2390:
```suggestion
extern "C" void dump_operation(MlirOperation op, const char *filename) {
```

Pangoraw (commented):
FYI, this should be doable from Julia using `MLIR.API.mlirOperationDump(operation)`

https://github.com/llvm/llvm-project/blob/2df51fd9c491ef1ad90fbad5ce18ea8ff8bd3117/mlir/include/mlir-c/IR.h#L796-L797

deps/ReactantExtra/API.cpp
giordano: Do you have an example of this at hand?

giordano:
> FYI, this should be doable from Julia using MLIR.API.mlirOperationDump(operation)

This was for use inside GDB, not Reactant.  But nevertheless, @wsmoses do we still need this if there's that function already available?

wsmoses:
no we want this to be an Operation*. not MlirOperation. And yes we still need this because we need to dump to a file, not stderr

giordano:
> And yes we still need this because we need to dump to a file, not stderr

I already forgot :facepalm:

wsmoses (approved):
lgtm

PR #975, opened 2025-03-20
Update to new Enzyme-JAX with new LLVM

giordano:
Let's see how far we get.


giordano:
@avik-pal only test failures are in sharding, can you please take a look?

avik-pal:
the non-jll tests can be ignored for now

wsmoses:
per latency, concurrently starting the jll in the bg [on main, so we'll discard this PR after checking tests pass]

wsmoses:
so @avik-pal feel free to do the julia changes now

giordano:
Superseded by #984.

PR #977, opened 2025-03-20
[Compiler] Add `print{debug=true}` pass

giordano:
@wsmoses is this the right place where to add the pass for https://github.com/PRONTOLab/GB-25/pull/75#issuecomment-2740045031?


wsmoses (commented):
yeah this is fine

PR #978, opened 2025-03-20
feat: add is_addressable for pjrt devices

avik-pal:


avik-pal:
@wsmoses if you are building the jll can we get this in as well

wsmoses:
I thought i started a build but i guess not, so yes!

PR #979, opened 2025-03-20
feat: more readable error printing

avik-pal:
### Before

image

### After

image


giordano:
Honestly, in my case the only solution would have been to redirect the output to file and read it carefully, in which case the current version is probably better (colours would most likely be lost when redirecting to a file). My problem is that I got a huge error message (almost 2k lines in tmux) in an interactive session, which was completely unmanageable in any case.

PR #963, opened 2025-03-19
WIP: misc GB fixes for constant

wsmoses:

PR #982, opened 2025-03-20
feat: dispatch on eltype for more wrapped arrays

avik-pal:


giordano (approved):
Looks good as far as I can tell

avik-pal:
Still needs some fixew the GB repo is broken

giordano:
@avik-pal I merged `main` to fix conflicts with #963 (in `src/TracedUtils.jl` and `src/stdlibs/LinearAlgebra.jl` only), please double check I didn't mess up!

PR #996, opened 2025-03-21
fix: misc GB fixes

avik-pal:


wsmoses (approved):

ext/ReactantCUDAExt.jl
```diff
@@ -40,6 +40,8 @@ struct CuTracedRNumber{T,A} <: Number
     end
 end
 
+CuTracedRNumber{T,A}(val::Number) where {T,A} = convert(CuTracedRNumber{T,A}, val)
```
wsmoses, line 43: Doesn't this need to convert to a tracedrnumber and then use that as an argument for the cutracedrnumber constructor?

wsmoses (approved):

PR #1001, opened 2025-03-22
fix: more missing functions

avik-pal:
This gets the kernels traced correctly.

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -63,13 +63,20 @@ for jlop in (
     :(Base.:/),
     :(Base.:^),
     :(Base.rem),
+    :(Base.isless),
+    :(Base.:(==)),
+    :(Base.:(!=)),
 )
     @eval begin
         @inline $jlop(a::CuTracedRNumber, b::CuTracedRNumber) = $jlop(a[], b[])
+        @inline $jlop(a::CuTracedRNumber{T,A}, b::Number) where {T,A} =
+            $jlop(a, convert(CuTracedRNumber{T, A}, b))
+        @inline $jlop(a::Number, b::CuTracedRNumber{T,A}) where {T,A} =
+            $jlop(convert(CuTracedRNumber{T, A}, a), b)
     end
 end
 
-for jlop in (:(Base.:+), :(Base.:-))
+for jlop in (:(Base.:+), :(Base.:-), :(Base.isnan), :(Base.isfinite))
     @eval begin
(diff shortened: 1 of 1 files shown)


ext/ReactantCUDAExt.jl
```diff
@@ -63,13 +63,20 @@ for jlop in (
     :(Base.:/),
     :(Base.:^),
     :(Base.rem),
+    :(Base.isless),
+    :(Base.:(==)),
+    :(Base.:(!=)),
 )
     @eval begin
         @inline $jlop(a::CuTracedRNumber, b::CuTracedRNumber) = $jlop(a[], b[])
+        @inline $jlop(a::CuTracedRNumber{T,A}, b::Number) where {T,A} =
```
wsmoses, line 72: the single side version specifically causes issues here (and shouldn't be necessary per the existance of promote_type.

Also note that isless is defined below (and does require both sides), so perhaps adding another for loop for isless, equal, and != that defines all 3 would be the right move

avik-pal: > shouldn't be necessary per the existance of promote_type

This was leading to a dynamic dispatch

ext/ReactantCUDAExt.jl
```diff
@@ -66,15 +66,25 @@ for jlop in (
...
+for jlop in (:(Base.:+), :(Base.:-), :(Base.isnan), :(Base.isfinite))
     @eval begin
         @inline $jlop(a::CuTracedRNumber) = $jlop(a[])
     end
 end
 
+for jlop in (:(Base.isless), :(Base.:(==)), :(Base.:(!=)))
+    @eval begin
+        @inline $jlop(a::CuTracedRNumber, b::CuTracedRNumber) = $jlop(a[], b[])
+        @inline $jlop(a::CuTracedRNumber{T,A}, b) where {T,A} = $jlop(a[], b)
```
wsmoses, line 83: can you change this to only apply if the other side is a Number

ext/ReactantCUDAExt.jl
wsmoses: huh, weirdly enough i was getting a dynamic dispatch when that was added itself (per CI in the mul test)

wsmoses: that is to say when i added it earlier in https://github.com/EnzymeAD/Reactant.jl/pull/996 then realized i needed to remove it

avik-pal: tests are passing now

wsmoses (approved):

PR #1005, opened 2025-03-22
Don't cleanup MLIR module written to file when pass fails

giordano:
At the moment the MLIR module file disappears after the Julia process exits, making this feature not much useful. Ref: https://github.com/EnzymeAD/Reactant.jl/pull/964#discussion_r2008841380.

### DIFF
--- a/src/mlir/IR/Pass.jl
+++ b/src/mlir/IR/Pass.jl
@@ -64,6 +64,8 @@ function enable_verifier!(pm, enable=true)
     return pm
 end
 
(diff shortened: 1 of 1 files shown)


wsmoses (approved):

giordano:
Can confirm this is finally working as expected: in https://github.com/PRONTOLab/GB-25/actions/runs/14013420361?pr=84 files are still there after the julia process exits, so that they can be uploaded automatically as artifacts.

PR #1006, opened 2025-03-22
feat: add safety checks to ensure no values outside function

avik-pal:


src/Ops.jl
```diff
@@ -48,6 +48,68 @@ function with_debug(f)
...
+    # generate a "unchecked" version that returns the error code instead
+    unchecked_sig = Expr(:call, Symbol("unchecked_", sig.args[1]), sig.args[2:end]...)
+    unchecked_def = Expr(:function, unchecked_sig, body)
+
+    return esc(:($safe_def, $unchecked_def))
+end
+
+function check(f::Function, args...)
+    result = f()
+    check_all_in_same_region([args, result])
```
wsmoses, line 109: we can simplify this (and also this isn't quite right).

We can ask for the current insertion context before creating the new operation. We also want to ensure that all operands are in the same function, which may not be the same region. In c++ we want to do the equlvalent of getParentOfType<FunctionOpInterface>()

wsmoses: specifically, as is, this won't work as is for an if/for

src/Ops.jl
```diff
@@ -48,6 +51,66 @@ function with_debug(f)
...
+    end
+
+    safe_fn = combinedef(fdef)
+
+    return esc(:($safe_fn, $unsafe_fn))
+end
+
+function check(f::Function, args...)
+    result = f()
+    check_all_in_same_region([args, result])
```
wsmoses, line 110: we can use https://github.com/EnzymeAD/Reactant.jl/blob/4c0e77f89c7094b5deb0ee80d84500604dc1882d/src/mlir/IR/Block.jl#L203 to get the current block to insert into

deps/ReactantExtra/API.cpp
```diff
@@ -2404,3 +2404,7 @@ extern "C" void dump_operation(Operation *op, const char *filename) {
 extern "C" bool pjrt_device_is_addressable(PjRtDevice *device) {
   return device->IsAddressable();
 }
+
+extern "C" mlir::Operation *mlirGetParentOfTypeFunctionOp(mlir::Operation *op) {
+  return op->getParentOfType<mlir::FuncOp>();
```
wsmoses, line 2409: functionopinterface ideally, not funcop

avik-pal:
will open a new PR, there's an easier solution

PR #1008, opened 2025-03-23
feat: early fail if not correct region

avik-pal:


src/mlir/IR/Operation.jl
```diff
@@ -331,16 +331,38 @@ function create_operation_common(
...
+    #         # TODO: add the checks
+    #     end
+    # end
+    return op
+end
+
 function create_operation(args...; kwargs...)
-    res = create_operation_common(args...; kwargs...)
+    res = create_operation_common_with_checks(args...; kwargs...)
     if _has_block()
```
wsmoses, line 348: technically speaking the actual issue here is actually on like 349, where we're accidentally creating a stablehlo.constant operation, and pushing it into a module. It's probably better to figure out we're creating it at the wrong place, rather than figure out during the use stage that we're using an operation which is itself in the wrong place

avik-pal: We had an older issue where a concreteRNG state wasn't being reset, so the region was completely different. We can't capture that without adding checks here

avik-pal:
currently low priority, will deal with this later

PR #1012, opened 2025-03-23
fix: move away from sdy internal API

avik-pal:


avik-pal:
nearly there for GB. need to handle subaxis for that problem

wsmoses (approved):

PR #1015, opened 2025-03-23
`copyto!(dest, dstart, src, sstart)`

jumerckx:


src/TracedRArray.jl
```diff
@@ -628,6 +628,10 @@ function Base.copyto!(dest::TracedRArray{T,N}, src::TracedRArray{T,N}) where {T,
     dest.mlir_data = src.mlir_data
     return dest
 end
+function Base.copyto!(dest::Reactant.TracedRArray{T}, dstart::Integer, src::Reactant.TracedRArray{T}, sstart::Integer, n::Integer) where T
+    dest[dstart:(dstart + n - 1)] .= src[sstart:(sstart + n - 1)]
```
wsmoses, line 632: For compilation speed can we change this to a copyto of the view rather than broadcast?

jumerckx: copyto! between TracedRArray views is not implemented and I don't know how better to doing a broadcasting copy? That's also how it's done for ConcreteRArray.

wsmoses: It should exist for the dest? (Though we don’t need the view)

wsmoses: But maybe that is not for copyto and only setindex

jumerckx: Alright! With setindex! it works as well, (without views then.)

wsmoses (approved):

PR #1017, opened 2025-03-24
Defer materialize until last moment

wsmoses:


avik-pal (approved):

wsmoses:
@avik-pal just pushed one more thing, could I get a re-review?

src/Ops.jl
```diff
@@ -488,6 +488,7 @@ end
     permutation;
     location=mlir_stacktrace("transpose", @__FILE__, @__LINE__),
 ) where {T,N}
+    @assert length(permutation) == length(size(x))
```
avik-pal, line 491:
```suggestion
    @assert length(permutation) == ndims(x)
```

PR #1031, opened 2025-03-25
fix: codegen for resharded input arrays

### LINKED ISSUE #1027: Bad codegen for resharding
```julia
x_ra = Reactant.to_rarray(
    randn(Float32, 32, 32);
    sharding=Sharding.NamedSharding(Sharding.Mesh(reshape(0:11, 3, 4), (:x, :y)), (:x, :y))
);

z_ra = Reactant.to_rarray(ones(Float32, 32, 32))

function test1!(x, z)
    y = x .+ x'
    x .+= y
    z .= x
    return z
end

@jit test1!(x_ra, z_ra)

@show Reactant.XLA.sharding(z_ra.data.buffer)
```

```
XLA.IFRT.Sharding("ConcreteEvenSharding(devices: BasicDeviceList([CpuDevice(id=0),CpuDevice(id=1),CpuDevice(id=2),CpuDevice(id=3),CpuDevice(id=4),CpuDevice(id=5),CpuDevice(id=6),CpuDevice(id=7),CpuDevice(id=8),CpuDevice(id=9),CpuDevice(id=10),CpuDevice(id=11)]), shape: [32,32], shard_shape: [32,32], memory_kind: unpinned_host, is_fully_replicated: false)")
```

This is incorrect, it should be XLA.IFRT.Sharding("SingleDeviceSharding(TFRT_CPU_0, memory_kind: unpinned_host)")

while not the exact error @giordano got, this will also lead to similar failures

avik-pal:
- [x] fixes #1027
  - [x] do it for all codepaths
- [x] make unresharding more efficient -- we need to ensure that we only unreshard once
- [x] add test
- [x] faster replication --- no tracing (we directly construct the MLIR needed) + more checks to avoid having to replicate in the first place

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -94,12 +94,23 @@ end
     return Base.setfield!(obj, field, (val[idx],))
 end
 
-function traced_setfield_buffer!(::Val, cache_dict, val, concrete_res)
-    return traced_setfield!(val, :data, concrete_res)
+function traced_setfield_buffer!(runtime::Val, cache_dict, concrete_res, obj, field)
+    return traced_setfield_buffer!(
+        runtime, cache_dict, traced_getfield(obj, field), concrete_res, obj, field
+    )
+end
+
+function traced_setfield_buffer!(::Val, cache_dict, val, concrete_res, obj, field)
+    return traced_setfield!(obj, field, concrete_res)
 end
 
 function traced_setfield_buffer!(
-    ::Val{:PJRT}, cache_dict, val::Union{TracedRArray,TracedRNumber}, concrete_res
+    ::Val{:PJRT},
+    cache_dict,
+    val::Union{TracedRArray,TracedRNumber},
+    concrete_res,
+    obj,
+    field,
 )
     if haskey(cache_dict, val)
         cval = cache_dict[val]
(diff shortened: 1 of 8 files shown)


src/Compiler.jl
```diff
@@ -588,24 +634,15 @@ const context_gc_vector = Dict{MLIR.IR.Context,Vector{Union{TracedRArray,TracedR
...
         MLIR.IR.verifyall(MLIR.IR.Operation(mod); debug=true)
         Text(repr(mod))
     end
-    Base.delete!(context_gc_vector, ctx)
-    return result
 end
 
 function compile_mlir(f, args; client=nothing, kwargs...)
-    ctx = MLIR.IR.Context(Reactant.registry[], false)
-    context_gc_vector[ctx] = Vector{Union{TracedRArray,TracedRNumber}}(undef, 0)
```
wsmoses, line 606: we shouldn't delete the constant cache?

wsmoses: ah jk I see the new funcgion does this

wsmoses (approved):

src/mlir/IR/Context.jl
```diff
@@ -79,6 +79,33 @@ function context!(f, ctx::Context)
...
+    if allow_use_existing && _has_context()
+        ctx = context()
+    else
+        delete_context = true
+        ctx = Context(Reactant.registry[], false)
+        Reactant.Compiler.context_gc_vector[ctx] = Vector{
+            Union{Reactant.TracedRArray,Reactant.TracedRNumber}
+        }(
+            undef, 0
+        )
```
giordano, line 93:
```suggestion
        empty!(Reactant.Compiler.context_gc_vector[ctx])
```

avik-pal: This will error no? We are trying to set an empty vector with the ctx to ensure it is not GCed (Reactant.Compiler.context_gc_vector doesn't contain ctx at this point).

giordano: What errors?
```julia-repl
julia> v = Vector{Union{Float64,Int64}}(undef, 10)
10-element Vector{Union{Float64, Int64}}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> empty!(v)
Union{Float64, Int64}[]
```

PR #1035, opened 2025-03-25
nicer error for type mutation in traced loop body

jumerckx:
I should still bump versions but I'm afk for a bit.
```jl
using Reactant

function mysum(x)
    s = 0.
    a = 10
    @trace track_numbers=false for _ in 1:3
        a += x
        s += x
    end
    s
end
@code_hlo mysum(ConcreteRNumber(4.))

```

```
ERROR: Types do not match between function arguments and results.
The following arguments should be traced:
a, s
```

### DIFF
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -231,7 +231,7 @@ function trace_for(mod, expr; track_numbers)
                 end
 
             $(ReactantCore).traced_while(
-                cond_fn, body_fn, args; track_numbers=$(track_numbers)
+                cond_fn, body_fn, args; track_numbers=$(track_numbers), verify_arg_names=$(QuoteNode(args_init))
             )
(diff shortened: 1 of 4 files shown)


wsmoses:
can we also print the full path if its not a top level variable?

(for example within the GB code its almost certainly deep within the single model struct)

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -231,7 +231,7 @@ function trace_for(mod, expr; track_numbers)
                 end
 
             $(ReactantCore).traced_while(
-                cond_fn, body_fn, args; track_numbers=$(track_numbers)
+                cond_fn, body_fn, args; track_numbers=$(track_numbers), verify_arg_names=$(QuoteNode(args_init))
```
wsmoses, line 234: can you also bump reactant core version in project.toml?

PR #1037, opened 2025-03-25
fix project tomls

wsmoses:


giordano:
@wsmoses can we merge this to tag a new version?

PR #1038, opened 2025-03-25
traced loop conflicts: Better error message

jumerckx:


jumerckx:
TODO: test

```jl
using Reactant

mutable struct Pair
    a
    b
end

function f(x, p)
    @trace track_numbers=false for _ in 1:3
        # newp = Pair(p.a + x, p.b + x)
        p.a += x
        p.b += x
    end
    return p
end

p = Pair(ConcreteRNumber(3), 4)
x = ConcreteRNumber(2)

@jit f(x, p)
```

PR #1039, opened 2025-03-25
Fix arg prefixing

wsmoses:


src/Ops.jl
```diff
@@ -1754,10 +1754,10 @@ end
...
     seen_args = Reactant.OrderedIdDict()
     traced_args = Vector{Any}(undef, N)
-    for i in 1:N
-        @inbounds traced_args[i] = Reactant.make_tracer(
-            seen_args, args[i], (), Reactant.NoStopTracedTrack; track_numbers
-        )
+
+    loopargsym = gensym("looparg")
+    for (i, prev) in enumerate(args)
+        Reactant.make_tracer(seen_args, prev, (), Reactant.NoStopTracedTrack; track_numbers)
```
jumerckx, line 1760: Still somewhat figuring out what's going on, but `loopargsym` is not used here?

PR #1042, opened 2025-03-26
feat: allow configuring precision & algorithm for dot_general

avik-pal:
Atleast locally if I force Highest Precision this fixes the issue with the autotuner. I still need to figure out how to use ScopedValues here


giordano:
ScopedValues would be useful in a lot of places in the compiler where we roll our own using task-local storage, repeating the same pattern all over the place. Once this dependency is introduced I can look into replacing that, if time.

wsmoses (approved):

PR #1044, opened 2025-03-26
Add canonicalize before libdevice-func-raise in pass pipeline

glwagner:


src/Compiler.jl
```diff
@@ -625,6 +625,8 @@ function optimization_passes(; no_nan::Bool=false, sroa::Bool=false, inline::Boo
             passes,
             "sroa-wrappers{instcombine=false instsimplify=true $(SROA_ATTRIBUTOR[] ? "" : "attributor=false")}",
         )
+
+        push!(passes, "canonicalize")
```
Pangoraw, line 629: Can you try to run lib device raise twice like this ?

```suggestion
        push!(passes, "libdevice-funcs-raise")
        push!(passes, "canonicalize")
```

wsmoses: actually im curious if as is fixes it

wsmoses:
hopefully resolved via https://github.com/EnzymeAD/Reactant.jl/pull/1047

PR #1045, opened 2025-03-26
Run passes on a single PM

Pangoraw:
We don't need to disable the verifier between enzyme and arith raising passes because arith raise is passed in enzyme post passes option. This should help with getting the IR on PM failure.

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -931,14 +931,14 @@ function compile_mlir!(
     end
 
     if optimize === :all
-        run_pass_pipeline!(mod, join([opt_passes, "enzyme-batch", opt_passes2], ","))
-        run_pass_pipeline!(
-            mod, "$enzyme_pass,arith-raise{stablehlo=true}"; enable_verifier=false
-        )
         run_pass_pipeline!(
             mod,
             join(
                 [
+                    opt_passes,
+                    "enzyme-batch",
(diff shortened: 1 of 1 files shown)


wsmoses (changes requested):
No we do need to do this, as Enzyme will generate codee which is invalid until raised by arith raise (and thus that part of the pipeline has enable_verifier=false).

We can use it now for debugging though for GB since we don't call autodiff

Pangoraw:
Since `arith-raise` is part of the `postpasses` option of `enzyme` it will be run before the enzyme pass is finished and the verifier is invoked. From the standpoint of the verifier, there will be no invalid IR.

Postpasses invokation in Enzyme: https://github.com/EnzymeAD/Enzyme/blob/f1bad2eccfebbc211f530b5c28aa9f7075298921/enzyme/Enzyme/MLIR/Interfaces/EnzymeLogicReverse.cpp#L269

wsmoses:
oh right, I forgot we added that (this code predates postpasses). Yeah for sure then

wsmoses (approved):
do we even need the arith raise after enzyme then?

Pangoraw:
> do we even need the arith raise after enzyme then?

No, I removed them. We could also remove canonicalize which I did not remove.

PR #1054, opened 2025-03-27
Fix tpu

wsmoses:


src/Compiler.jl
```diff
@@ -1189,13 +1189,22 @@ function compile_mlir!(
     # Add a `donated` attr to the function arguments. This doesn't affect XLA, but lets us
     # check which arguments were donated.
     preserved_args_idx = last.(preserved_args)
+    if backend != "tpu"
```
wsmoses, line 1192: @avik-pal this is required for TPU (unknown attributes create a bug where it fails to serialize)

PR #1056, opened 2025-03-27
feat: expose more ifrt API + abi changes for compile

avik-pal:


avik-pal:
ABI changed so tests will fail

PR #1060, opened 2025-03-27
fix: logical vs physical device handling

avik-pal:


wsmoses (approved):

avik-pal:
I am disabling the support for non-iota meshes on PJRT. It gives too much headache to handle that case, and its not even worth it given we should be using IFRT for sharding in most cases

PR #1063, opened 2025-03-27
fix: print mlir

avik-pal:


src/Compiler.jl
```diff
@@ -1132,12 +1132,12 @@ function compile_mlir!(
...
             end
 
-            run_pass_pipeline!(mod, join(["xla-sdy-stablehlo-export-pipeline"], ','))
+            run_pass_pipeline!(mod, join(["print{debug=true}", "xla-sdy-stablehlo-export-pipeline", "print{debug=true}"], ','))
 
             # Run our optimization passes here -- we need to be careful to not apply folding
             # here since that violates the semantics of `sdy.constant` which was converted to
             # `stablehlo.constant` by the previous pass.
-            run_pass_pipeline!(mod, join(["canonicalize", "cse"], ','))
+            run_pass_pipeline!(mod, join(["canonicalize", "case", "print{debug=true}"], ','))
```
wsmoses, line 1140: @avik-pal i presume you meant cse not case?

avik-pal, line 1140:
```suggestion
            run_pass_pipeline!(mod, join(["canonicalize", "cse", "print{debug=true}"], ','))
```

Issue #1020, opened 2025-03-24
Shardy StableHLO Import failed Check failed: product == localAxisSize (4 vs. 2)


avik-pal:
```
[3431209] signal 6 (-6): Aborted                                                                                                                                                                                                                                                                                                                                           
in expression starting at /home/avik-pal/reactant/GB-25/sharding/simple_sharding_problem.jl:60                                                                                                                                                                                                                                                                             
pthread_kill at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)                                                                                                                                                                                                                                                                                                             
gsignal at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)                                                                                                                                                                                                                                                                                                                  
abort at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)                                                                                                                                                                                                                                                                                                                    
_ZN3tsl8internal15LogMessageFatalD1Ev at /home/avik-pal/.julia/artifacts/ad1b5fcb6f65919d73818a6cb908a4221cf29afb/lib/libReactantExtra.so (unknown line)                                                                                                                                                                                                                   
_ZN3xla3sdy20convertToSdyShardingERKNS_11HloShardingEN4mlir3sdy8MeshAttrERKN4llvm13SmallDenseMapIlNS7_9StringRefELj4ENS7_12DenseMapInfoIlvEENS7_6detail12DenseMapPairIlS9_EEEElb.cold at /home/avik-pal/.julia/artifacts/ad1b5fcb6f65919d73818a6cb908a4221cf29afb/lib/libReactantExtra.so (unknown line)                                                                   
hloShardingToTensorShardingAttr at /home/avik-pal/.julia/artifacts/ad1b5fcb6f65919d73818a6cb908a4221cf29afb/lib/libReactantExtra.so (unknown line)                                                                                                                                                                                                                         
#get_tensor_sharding_attribute#38 at /home/avik-pal/reactant/Reactant.jl/src/Sharding.jl:542                                                                                                                                                                                                                                                                               
get_tensor_sharding_attribute at /home/avik-pal/reactant/Reactant.jl/src/Sharding.jl:530 [inlined]                                                                                                                                                                                                                                                                         
#get_tensor_sharding_attribute#39 at /home/avik-pal/reactant/Reactant.jl/src/Sharding.jl:574 [inlined]                                                                                                                                                                                                                                                                     
[...]

PR #1073, opened 2025-03-28
fix: shardy passes

avik-pal:
As per discussion with @GleasonK on slack, we can't safely run `canonicalize` just after propagation.

* For `xla_default` we are still running propagation outside XLA since we need to extract the sdy shardings (xref #1020). (changed the name because :default wasn't our "default" :P)
* `to_mhlo_shardings` goes all the way and does the full export and drops the sdy attributes (and runs cse, being added upstream to sdy pipeline as well)
* `:none` doesn't run any passes and exists to allow a nicer printing of the generated mlir for end users

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -790,7 +790,7 @@ function compile_mlir!(
     sdycache=default_sdycache();
     optimize::Union{Bool,Symbol}=true,
     # default refers to letting XLA handle the shardy inport/propagation/export
-    shardy_passes::Symbol=:to_mhlo_shardings, # [:default, :to_mhlo_shardings]
+    shardy_passes::Symbol=:to_mhlo_shardings, # [:none, :xla_default, :to_mhlo_shardings]
     no_nan::Bool=false,
     assert_nonallocating::Bool=false,
     backend="gpu",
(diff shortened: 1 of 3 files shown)


GleasonK:
CSE may be dangerous too on further inspection, as it can un-sink constants:

```mlir
$ cat /tmp/t.mlir
module {
  func.func @while_with_sinked_constants(%arg0: tensor<32x96xf32>) -> tensor<32x96xf32> {
    %c = stablehlo.constant dense<1> : tensor<i32>
    %0:2 = stablehlo.while(%iterArg = %arg0, %iterArg_0 = %c) : tensor<32x96xf32>, tensor<i32>
     cond {
      %c_1 = stablehlo.constant dense<32> : tensor<i32>
      %1 = stablehlo.compare  LT, %iterArg_0, %c_1 : (tensor<i32>, tensor<i32>) -> tensor<i1>
      stablehlo.return %1 : tensor<i1>
    } do {
      %c_1 = stablehlo.constant dense<1> : tensor<i32>
      %1 = stablehlo.add %iterArg_0, %c_1 : tensor<i32>
      %2 = stablehlo.add %iterArg, %iterArg : tensor<32x96xf32>
      stablehlo.return %2, %1 : tensor<32x96xf32>, tensor<i32>
    }
    return %0#0 : tensor<32x96xf32>
  }
}

$ stablehlo-opt /tmp/t.mlir --cse
module {
  func.func @while_with_sinked_constants(%arg0: tensor<32x96xf32>) -> tensor<32x96xf32> {
    %c = stablehlo.constant dense<1> : tensor<i32>
    %0:2 = stablehlo.while(%iterArg = %arg0, %iterArg_0 = %c) : tensor<32x96xf32>, tensor<i32>
     cond {
      %c_1 = stablehlo.constant dense<32> : tensor<i32>
      %1 = stablehlo.compare  LT, %iterArg_0, %c_1 : (tensor<i32>, tensor<i32>) -> tensor<i1>
      stablehlo.return %1 : tensor<i1>
    } do {
      %1 = stablehlo.add %iterArg_0, %c : tensor<i32>
      %2 = stablehlo.add %iterArg, %iterArg : tensor<32x96xf32>
      stablehlo.return %2, %1 : tensor<32x96xf32>, tensor<i32>
    }
    return %0#0 : tensor<32x96xf32>
  }
}
```

I'm going to work on an explicit constant de-duper in SDY so we can safely do this. Won't be as powerful as CSE, but if we're reducing constants by 25% then we may not need something as powerful as CSE

PR #1076, opened 2025-03-28
fix: run finalizer on array

avik-pal:


wsmoses (approved):
lgtm and we should do, but also I'm separately surprised if we are allocating constantly within the loop

avik-pal:
this seems to solve the issue @giordano was facing. now the memory allocations are constant

avik-pal:
arm is segfaulting ....

avik-pal:
have a fix locally... will need a new jll unfortunately

avik-pal:
xref https://github.com/JuliaPackaging/Yggdrasil/pull/10848

PR #1077, opened 2025-03-28
fix: remove ifrt::HloSharding handling

avik-pal:
JLL changes for https://github.com/EnzymeAD/Reactant.jl/pull/1076


avik-pal:
having ifrt::HloSharding leads to unnecessary problems of double free.

wsmoses (approved):

PR #1090, opened 2025-03-29
fix: allow not passing a memory kind

avik-pal:


deps/ReactantExtra/API.cpp
```diff
@@ -1941,6 +1941,14 @@ extern "C" ifrt::MemoryKind *ifrt_memory_kind_from_string(const char *c_str) {
   return new ifrt::MemoryKind(std::string(c_str));
 }
 
+extern "C" ifrt::MemoryKind *ifrt_memory_kind_with_optional_memory_space() {
+  return new ifrt::MemoryKind(std::nullopt);
```
wsmoses, line 1945: tangentially a memorykind is apparently just a string wrapper, we could potentially store this as a string if we want to

avik-pal: agreed, though at this point I just want to get the multi-node stuff running. so opening an issue to handle this later https://github.com/EnzymeAD/Reactant.jl/issues/1093

wsmoses: Yeah for sure, super non urgent

avik-pal:
waiting for https://github.com/JuliaPackaging/Yggdrasil/pull/10859

giordano:
Usual reminder that `git push --force-with-lease` is a thing 🙃

PR #1094, opened 2025-03-29
fix: kernel abstractions

avik-pal:


avik-pal:
lets see if this fixes the kernel abstractions error we were seeing

wsmoses:
I’m ever so slightly ahead of you https://github.com/EnzymeAD/Reactant.jl/pull/1092

avik-pal:
Actually we will still need this one independently from your changes

wsmoses (approved):

PR #1095, opened 2025-03-29
Revert "feat: codegen without embedding runtime pointers into body (#…

avik-pal:
…1089)"

This reverts commit 7a4a9d0e2912506252016db5d51fa928434fc108.


avik-pal:
false alarm this is not the source of our troubles

PR #1108, opened 2025-04-01
fix: let shardy run everything

avik-pal:


avik-pal:
@glwagner we finally have a `Sharding.Replicated` type. No need to do the whole `Returns(nothing)` trick anymore

avik-pal:
this moves the entire sharding to inside XLA to avoid any screwups on our end. Though it has no effect on the all gathers

wsmoses (approved):

PR #1120, opened 2025-04-02
feat: padded sharding

avik-pal:


src/Types.jl
```diff
@@ -283,14 +293,21 @@ mutable struct ConcreteIFRTArray{T,N,S<:Sharding.ShardInfo} <: AbstractConcreteA
...
     sharding::S
     donated::Bool
+    padding::NTuple{N,Int}
 
     function ConcreteIFRTArray{T,N,S}(
-        data::XLA.IFRT.AsyncArray, shape::NTuple{N,Int}, sharding::S
+        data::XLA.IFRT.AsyncArray,
+        shape::NTuple{N,Int},
+        sharding::S,
+        padding::NTuple{N,Int}=ntuple(Returns(0), N),
```
wsmoses, line 302: Can we have this be a type parameter where nothing has a special case

e.g.
```
padding <: union{Nothing, NTuple{N, Int}}
padding::Padding
```

wsmoses (approved):

PR #1123, opened 2025-04-03
feat: init directly on device

avik-pal:
- [x] add a test

### DIFF
--- a/src/ConcreteRArray.jl
+++ b/src/ConcreteRArray.jl
@@ -600,12 +600,13 @@ end
 
 # Directly initialize a Device Array
 function Base.fill(
-    ::Type{<:ConcreteIFRTArray},
+    ::Type{<:Union{ConcreteIFRTArray,ConcretePJRTArray}},
     val,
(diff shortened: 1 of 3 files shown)


wsmoses (approved):

src/ConcreteRArray.jl
```diff
@@ -597,3 +597,17 @@ function Base.map!(f, R::Union{AnyConcreteIFRTArray,AnyConcretePJRTArray}, A::Ab
     fn(f, R, A)
     return R
 end
+
+# Directly initialize a Device Array
+function Base.fill(
+    ::Type{<:Union{ConcreteIFRTArray,ConcretePJRTArray}},
```
glwagner, line 603: Is there a way to get this from the runtime setting? Because right now, the user may not have explicit access to that but I believe it dictates `to_rarray` under the hood for example.

glwagner: Ah I see in your example that you just use `ConcreteRArray`.

PR #1125, opened 2025-04-04
[Compiler] Add new passes

giordano:
Requires https://github.com/JuliaPackaging/Yggdrasil/pull/10923.


Pangoraw:
Any reasons to omit `dus_to_concat` from https://github.com/EnzymeAD/Enzyme-JAX/pull/608 ?

PR #1131, opened 2025-04-04
feat: shardy progagation options [JLL]

avik-pal:


deps/ReactantExtra/API.cpp
```diff
@@ -2559,3 +2563,20 @@ extern "C" HeldIfrtArray *ifrt_make_array_from_host_buffer_shards(
       static_cast<ifrt::Client::HostBufferSemantics>(c_host_buffer_semantics)));
   return reactant::capture(arrays[0]);
 }
+
+extern "C" void addSdyPropagationPipeline(
+    mlir::OpPassManager &pm, bool keepShardingRules /*false*/,
+    bool conservativePropagation /*false*/, bool debugShardingOrigins /*false*/,
```
wsmoses, line 2569: can you make these uint8_t, and convert to bool by doing val != 0?

PR #1135, opened 2025-04-05
updates vitepress config

lazarusA:
this PR only deals with updating the docs pages.

- [x] latest `npm` package.json list packages
- [x] adds favicon
- [x] fixes to sidebar syntax (using documented way of adding multiple sidebars)
- [x] GitHub icon with GitHub stars   
- [x] Enhanced Readabilities menu (introduced in the latest DVPress update)
- [x] use of default css files
- [x] use default versionPicker (removes the old one) 
- [x] simplifies `make.jl` since `pages` is actually being done at the `config.mts`  level.
- [x]  Dead links (some) NOTE that upstream links from Dialects will be overwritten. Ignore for now, until a better solution or strategy is envisioned.


lazarusA:
WIP:

Screenshot 2025-04-05 at 11 43 35

docs/src/.vitepress/config.mts
```diff
@@ -15,9 +16,35 @@ export default defineConfig({
   description: "Documentation for Reactant.jl",
   cleanUrls: true,
   outDir: 'REPLACE_ME_DOCUMENTER_VITEPRESS', // This is required for MarkdownVitepress to work correctly...
-
-  ignoreDeadLinks: true, // tested in Documenter.jl directly
-
+  ignoreDeadLinks: false, // tested in Documenter.jl directly
```
avik-pal, line 19: the dead links are coming from the auto-generated MLIR Dialects files, lets ignore them for the time being

lazarusA: too late 😢, I almost finish update them.

lazarusA: done! from my side, things should work now. 

Please, if comments for small details, please  commit/suggest here directly 😄 .

avik-pal: Unfortunately, any changes to the Dialects directory will be overwritten by the automatic CI :cry:. The links come from upstream packages which we have limited control over

lazarusA: that doesn't sound ideal. But, surely(?) that automatic step doesn't happened every time, we should be able to fix this, specially the links, otherwise this will be broken always.

avik-pal: it does, every 24 hr the CI runs, fetches code from upstream (we live on the master branch of pretty much all upstream repositories so updates are very frequent), and if there is a diff overwrites the version on main.

We could try to write an automatic mapping from auto-generated links to the correct link, but that sounds like a lot of effort and needs to be done separately for each dialect

lazarusA: uff, ok, got it! setting ignore deadinks to `true`  for when things get overwritten and dead links are back. Otherwise, things should work.

avik-pal (approved):

docs/src/.vitepress/config.mts
wsmoses: yeah its about to be auto overwritten here: https://github.com/EnzymeAD/Reactant.jl/pull/1139

coming from https://github.com/EnzymeAD/Reactant.jl/blob/c8e42bbe16c4049fec45e1cd25fbf23516aead36/.github/workflows/regenerate-mlir-bindings.yml#L39

from https://github.com/EnzymeAD/Reactant.jl/blob/2ba8da2991f5ddaa7aa03514b1cd47134b2b3b1d/deps/ReactantExtra/make-bindings.jl#L44

autogenerated via https://github.com/EnzymeAD/Reactant.jl/blob/2ba8da2991f5ddaa7aa03514b1cd47134b2b3b1d/deps/ReactantExtra/tblgen/jl-generators.cc

if you want to help add support for the actual url that would be amazing!

PR #1136, opened 2025-04-05
feat: disassemble IFRT array into shards for serialization

avik-pal:


src/Types.jl
```diff
@@ -434,6 +434,43 @@ function ConcreteIFRTArray{T,N}(x::AnyConcreteIFRTArray; kwargs...) where {T,N}
...
+
+    all_devices = XLA.get_device.((XLA.client(x),), x.sharding.mesh.device_ids)
+    array_slices, _ = XLA.sharding_to_concrete_array_indices(
+        convert(XLA.CondensedOpSharding, hlo_sharding),
+        padded_size,
+        x.sharding.mesh.logical_device_ids,
+    )
+    return [
+        slice => ConcreteIFRTArray{T,N}(
+            XLA.IFRT.AsyncArray(shard, nothing), length.(slice), Sharding.NoShardInfo()
```
glwagner, line 467:
```suggestion
            XLA.IFRT.AsyncArray(shard, nothing), map(length, slice), Sharding.NoShardInfo()
```

src/Types.jl
```diff
@@ -434,6 +434,71 @@ function ConcreteIFRTArray{T,N}(x::AnyConcreteIFRTArray; kwargs...) where {T,N}
...
+                mapping_unpadded[i] =
+                    Tuple(new_slice) => shard[map(Base.OneTo ∘ length, new_slice)...]
+            end
+        end
+    end
+
+    return [
+        mapping_unpadded[i] for
+        i in 1:length(mapping_unpadded) if isassigned(mapping_unpadded, i)
+    ]
```
glwagner, line 499: in the typical case that we have one array per process, we'll use something like

```julia
adev = Sharding.disassemble_into_single_device_arrays(a)
acpu = Array(adev[1])
```

is that right?

avik-pal: I dont think you will have 1 array per process. If you have 4 gpus per node based on our current run configurations, you will have 4 arrays per process.

to get the arrays you can do `[s => Array(x) for (s, x) in ret_val]`. `s` is basically telling you the slice of this array as part of the global array

glwagner: ah right ok thanks. per device, not per process

PR #1145, opened 2025-04-06
fix: use window_reversal for conv

avik-pal:
ideally we write a pass to do this automatically

### DIFF
--- a/ext/ReactantNNlibExt/Implementations.jl
+++ b/ext/ReactantNNlibExt/Implementations.jl
@@ -95,15 +95,15 @@ function overloaded_conv!(
         collect(Int, size(y)), Reactant.MLIR.IR.Type(T)
     )
 
-    weight = W
-    if !flipkernel
-        weight = Reactant.Ops.reverse(weight; dimensions=kernel_spatial_dims)
-    end
-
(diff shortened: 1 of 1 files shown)


ext/ReactantNNlibExt/Implementations.jl
```diff
@@ -95,15 +95,15 @@ function overloaded_conv!(
...
-
     conv = Reactant.MLIR.Dialects.stablehlo.convolution(
         get_mlir_data(x),
-        get_mlir_data(weight);
+        get_mlir_data(W);
         result_0=result_type,
+        window_reversal=if flipkernel
+            nothing
+        else
+            [true for _ in 1:length(kernel_spatial_dims)]
```
giordano, line 105:
```suggestion
            fill(true, length(kernel_spatial_dims))
```
? I find it easier to read.

Pangoraw:
We can do this, but FYI: https://github.com/openxla/stablehlo/issues/1181

ext/ReactantNNlibExt/Implementations.jl
Pangoraw: or just `trues(length(kernel_spatial_dims))`

avik-pal:
Oh if it's deprecated then it doesn't make sense to use it

PR #1148, opened 2025-04-06
Print the passes in MLIR dump also before running them

giordano:


src/mlir/IR/Pass.jl
```diff
@@ -147,7 +147,7 @@ Run the provided `passManager` on the given `module`.
 function run!(pm::PassManager, mod::Module, key::String="")
     # Dump MLIR before running the pass manager.  We set `pm` to nothing because
     # the pass manager isn't called yet here.
```
giordano, line 149:
```suggestion
    # Dump MLIR before running the pass manager, but also print the list of passes that will be called later.
```

avik-pal (approved):

giordano:
Actually knowing from the filename that the file is pre or post pass manager reduces the confusion of always printing the passes, so thanks for that!

PR #1153, opened 2025-04-07
make optimize string option more effectful

Pangoraw:


src/Compiler.jl
```diff
@@ -1274,7 +1274,8 @@ function compile_mlir!(
         )
     end
 
-    if optimize ∉ (:none, :just_batch, :canonicalize) &&
+    if !(optimize isa String) &&
```
wsmoses, line 1277: we also need this form the transpose_propagate above

src/Compiler.jl
```diff
@@ -1090,7 +1090,7 @@ function compile_mlir!(
         # TODO remove redundant libdevice raise after fixing phase ordering
         result =
             "canonicalize,llvm-to-memref-access,canonicalize,convert-llvm-to-cf,canonicalize,enzyme-lift-cf-to-scf,canonicalize,func.func(canonicalize-loops),canonicalize-scf-for,canonicalize,libdevice-funcs-raise,canonicalize,affine-cfg,canonicalize,func.func(canonicalize-loops),canonicalize,llvm-to-affine-access,canonicalize,delinearize-indexing,canonicalize,simplify-affine-exprs,affine-cfg,canonicalize,func.func(affine-loop-invariant-code-motion),canonicalize,sort-memory,raise-affine-to-stablehlo{prefer_while_raising=false dump_failed_lockstep=$(DUMP_FAILED_LOCKSTEP[])},canonicalize,arith-raise{stablehlo=true}," *
-            opt_passes2
+            ifelse(optimize isa String, optimize, opt_passes2)
```
wsmoses, line 1093: this shuoldn't get called ever if using optimize as a string so isn't needed as a change

wsmoses:
partially incorporated into https://github.com/EnzymeAD/Reactant.jl/pull/1160

PR #1157, opened 2025-04-07
Add dus_slice_simplify

wsmoses:


Pangoraw:
The fix for the crash caused by dus_slice_simplify: https://github.com/EnzymeAD/Enzyme-JAX/pull/689

PR #1162, opened 2025-04-08
[deps] Define `using_clang` option when building with Clang

giordano:


deps/build_local.jl
```diff
@@ -160,7 +160,15 @@ if cc_is_gcc
         if gcc_version < v"12"
             push!(build_cmd_list, "--define=xnn_enable_avx512fp16=false")
         end
+        if gcc_version < v"11"
+            # TODO: this is not sufficient to complete a build with GCC 10.
```
giordano, line 164: I didn't manage to make GCC 10 work.  There are some compiler flags forced by XNN which I didn't figure out how to disable.  Maybe we should have a warning that versions <= 10 aren't well supported (by XLA, that is)?

wsmoses (approved):

PR #1174, opened 2025-04-12
fix: don't delete the compiled function

avik-pal:
the compiled function might not get inlined


avik-pal:
dont merge adding a test

PR #1192, opened 2025-04-18
fix: aliasing of results with buffers set later

### LINKED ISSUE #1289: Duplicating Scalars for Optimisers prevents CSE
The learning rates are duplicated for each parameter which means we will have a bazzillion `broadcast_in_dim`s in the IR and those cant be CSEd

avik-pal:
fixes https://github.com/LuxDL/Lux.jl/issues/1289

### DIFF
--- a/test/integration/optimisers.jl
+++ b/test/integration/optimisers.jl
@@ -14,10 +14,15 @@ is_empty_buffer(x::ConcretePJRTNumber) = any(x === C_NULL for x in x.data)
 end
 
 @testset "Correct Aliasing" begin
-    ps = (a = rand(4), b = rand(2)) |> Reactant.to_rarray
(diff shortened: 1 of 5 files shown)


test/integration/optimisers.jl
```diff
@@ -12,3 +12,12 @@ is_empty_buffer(x::ConcretePJRTNumber) = any(x === C_NULL for x in x.data)
     @test !is_empty_buffer(opt_state[2].rule.eta)
     @test !is_empty_buffer(opt_state[3].rule.eta)
 end
+
+@testset "Correct Aliasing" begin
+    ps = (a = rand(4), b = rand(2)) |> Reactant.to_rarray
```
wsmoses, line 17: lgtm but for ease can we come up with a smaller test that would error (just to make future debugging of the result cache code easier)

wsmoses (approved):

PR #1198, opened 2025-04-21
fix: codegen for partialsortperm

avik-pal:


avik-pal:
I am adding a lowering for `approx_top_k` and  we can configure that with `with_config`

PR #1210, opened 2025-04-24
feat: lower mapslices to batch

avik-pal:


src/Ops.jl
```diff
@@ -177,7 +177,11 @@ function fill(v, ::Tuple{}; location=mlir_stacktrace("fill", @__FILE__, @__LINE_
     return fill(v, Int[]; location)
 end
 
-function fill(number::TracedRNumber{T}, shape::Vector{Int}; location) where {T}
+function fill(
+    number::TracedRNumber{T},
+    shape::Vector{Int};
+    location=mlir_stacktrace("fill", @__FILE__, @__LINE__),
+) where {T}
     return Base.fill(number, Tuple(shape))
```
wsmoses, line 185: this feels problematic, why are we even calling base? (and we're obviously not forwarding the location)

wsmoses: yeah I think so

avik-pal: done

avik-pal:
@wsmoses 
 https://github.com/EnzymeAD/Reactant.jl/commit/6868dbe9f7d817c5f4a17c813db97285096ce435 breaks the complex test

wsmoses:
go ahead and mark that as expected broken for now, opened an issue (essentially I didn't realize when we fused the pass managers into a single one, that it wasn't an NFC and we changed the order of passes, which causes other issues that I'm now working through)

PR #1218, opened 2025-04-28
feat: SparseTensor dialect

avik-pal:


wsmoses (approved):
We may also need to add to registration in enzyme jax

PR #1225, opened 2025-04-30
Add verifyPostPasses = false in enzyme pass

Pangoraw:


giordano:
Uhm, I didn't close this PR, but I did delete the target branch.  If this PR is still relevant it should probably be reopened against `main`.

wsmoses:
so it incidentally does turn out that we resolved this since (I hope) by making enzyme-math into an interface and registering stablehlo add/sub into

giordano:
If so, the test marked as broken should be successful now (hence failing), right?

wsmoses:
The broken was removed in the jll bump importing that

Issue #1229, opened 2025-04-30
Error from `mapfirst!` into a SubArray of TracedArray


glwagner:
Originates from `ClimaOcean.Bathymetry.regrid_bathymetry` ultimately

cc @NoraLoose from your global simulation example

```
julia> include("global_ocean_simulation.jl")
[ Info: Oceananigans will use 16 threads
ERROR: LoadError: Scalar indexing is disallowed.
Invocation of getindex(::TracedRArray, ::Vararg{Int, N}) resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] errorscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:151
  [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:124
  [4] assertscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:112
  [5] getindex(::Reactant.TracedRArray{Float64, 2}, ::Int64, ::Int64)
    @ Reactant.TracedRArrayOverrides ~/.julia/packages/Reactant/yO34a/src/TracedRArray.jl:62
  [6] getindex
    @ ~/.julia/packages/Reactant/yO34a/ext/ReactantOffsetArraysExt.jl:52 [inlined]
  [7] Δxᶜᶜᵃ
    @ ~/.julia/packages/Oceananigans/lJ5jT/src/Operators/spacings_and_areas_and_volumes.jl:230 [inlined]
  [8] Δxᶜᶜᶜ
    @ ~/.julia/packages/Oceananigans/lJ5jT/src/Operators/spacings_and_areas_and_volumes.jl:93 [inlined]
  [9] Δx
    @ ~/.julia/packages/Oceananigans/lJ5jT/src/Operators/spacings_and_areas_and_volumes.jl:370 [inlined]
 [10] getindex
[...]

PR #1230, opened 2025-04-30
Fix broadcastosize of general abstract array

wsmoses:


src/TracedUtils.jl
```diff
@@ -952,7 +952,7 @@ function elem_apply(f, args::Vararg{Any,Nargs}) where {Nargs}
     return traced2_result
 end
 
-function broadcast_to_size(arg::AnyTracedRArray, rsize)
+function broadcast_to_size(arg::AbstractArray{<:TracedRNumber}, rsize)
```
avik-pal, line 955: AnyTracedRArray is an alias for this

wsmoses: Oh darn, then clearly this is not a quick fix for https://github.com/EnzymeAD/Reactant.jl/issues/1229

PR #1240, opened 2025-05-03
feat: ReactantExtra expose register profile

avik-pal:


wsmoses (approved):

wsmoses (commented):
Also make sure to add to the exported fn list

PR #1246, opened 2025-05-04
feat: add a macro to directly visualize the generated mlir

avik-pal:
```julia
julia> @mlir_visualize fn(x_ra)
Loading extensions...
2025-05-04 18:33:43.039914: I external/org_tensorflow/tensorflow/core/util/port.cc:153] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
Loaded 8 extensions:
 - TFLite adapter (Flatbuffer)
 - TFLite adapter (MLIR)
 - TF adapter (MLIR)
 - TF adapter (direct)
 - GraphDef adapter
 - Pytorch adapter (exported program)
 - MLIR adapter
 - JSON adapter

Starting Model Explorer server at:
http://localhost:8080/?data=%7B%22models%22%3A%20%5B%7B%22url%22%3A%20%22/tmp/jl_jSFMXkryRI.mlir%22%7D%5D%7D

Press Ctrl+C to stop.
Opening in existing browser session.
```

Same as code_hlo but opens a browser session


src/Compiler.jl
```diff
@@ -2068,6 +2068,51 @@ macro jit(args...)
     #! format: on
 end
 
+"""
+    @mlir_visualize [optimize = ...] [no_nan = <true/false>] f(args...)
+
+Runs `@code_hlo` and visualizes the MLIR module using `model-explorer`. This expects the
+`model-explorer` executable to be in your `PATH`. Installation instructions can be found
```
wsmoses, line 2075: this feels worthwhile just making into a (separate) jll of our own imo, rather than relying on path

src/Compiler.jl
```diff
@@ -2068,6 +2068,51 @@ macro jit(args...)
...
+[here](https://github.com/google-ai-edge/model-explorer).
+"""
+macro mlir_visualize(args...)
+    default_options = Dict{Symbol,Any}(
+        :optimize => true,
+        :no_nan => false,
+        :client => nothing,
```
wsmoses, line 2082: tangentially from this PR, we should consider having these options defaults in one place (rather than copied many places)

src/Compiler.jl
avik-pal: I agree, though it seems to be a JS framework, not sure how to ship the binary via yggy. @giordano might know?

giordano: Something like https://github.com/JuliaPackaging/Yggdrasil/blob/179e92aaf902a50c7876b0d088b3de88091be228/D/D3/build_tarballs.jl? What's the executable? A script which requires an interpreter?

avik-pal: Ok I was mistaken, only the UI components are present in the npm package. https://github.com/google-ai-edge/model-explorer/blob/main/src/builtin-adapter/python/pip_package/build_pip_package.sh has the bazel scripts for the python wheel

PR #1259, opened 2025-05-06
Undefined box

wsmoses:
idk if there's a better way to check for an undefined box

```
julia> x = Core.Box()
Core.Box(#undef)

julia> x.contents
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getproperty(x::Core.Box, f::Symbol)
   @ Base ./Base.jl:37
 [2] top-level scope
   @ REPL[3]:1
   ```

### DIFF
--- a/src/Tracing.jl
+++ b/src/Tracing.jl
@@ -1754,8 +1754,7 @@ Base.@nospecializeinfer function make_tracer(
     ))
 end
 
-struct UndefinedBox
-end
+struct UndefinedBox end
 
 Base.@nospecializeinfer function make_tracer(
     seen,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Tracing.jl [hunk 1]
```diff
@@ -1762,14 +1764,23 @@ Base.@nospecializeinfer function make_tracer(
     @nospecialize(sharding = Sharding.NoSharding()),
     kwargs...,
 )
+    prev2 = try
+        prev.contents
```
avik-pal, line 1768: `isassigned(prev, :contents)`

giordano: Probably `isdefined` rather than `isassigned`

PR #1260, opened 2025-05-06
feat: intel xpu register

avik-pal:
we should really be doing a jll, but for the time being...


giordano:
On my laptop I have both Nvidia and Intel GPUs, I'm curious to see how this would work :sweat_smile:

avik-pal:
https://buildkite.com/julialang/reactant-dot-jl/builds/8739/steps?jid=0196a394-3ddb-4972-9e75-a57a92eae6c1#0196a394-3ddb-4972-9e75-a57a92eae6c1/273-692 this is probably why we should do a jll :sweat:. Shipping https://github.com/JuliaGPU/oneAPI.jl/blob/28d7eedb553831e1d8bc82919808d1106220843b/Project.toml#L27C1-L29 should fix this, I think?

> On my laptop I have both Nvidia and Intel GPUs, I'm curious to see how this would work 😅

That is same as my setup. This should ideally just register both the intel and nvidia gpus with reactant but use nvidia GPU by default unless we do `Reactant.set_default_backend("sycl")`.

avik-pal:
Can confirm that even the python intel-extension-for-openxla doesn't ship with all the required libraries. But installing the support_jll and loader_jll gets us all the ones needed there

PR #1278, opened 2025-05-13
Ext: add onehotarrays

wsmoses:
requires https://github.com/FluxML/OneHotArrays.jl/pull/59


mcabbott:
Thanks! And sorry about the noisy robot...

When fiddling I was going to add also
```
Adapt.parent_type(::Type{<:OneHotArray{<:Any, <:Any, <:Any, T}}) where {T} = T
```
as warnings asked for this. And some guardrails to prevent some mistakes which the `<:Integer` restriction may previously have rejected:
```
OneHotArray(indices::AbstractArray{Bool}, L::Int) =
    throw(ArgumentError("OneHotArray(indices, L) requires an array of integer indices, not Bool"))
OneHotArray(indices::AbstractArray{<:CartesianIndex}, L::Int) =
    throw(ArgumentError("OneHotArray(indices, L) requires an array of integer indices, and does not accept CartesianIndex"))
OneHotArray(indices::AbstractArray{<:AbstractFloat}, L::Int) =
    throw(ArgumentError("OneHotArray(indices, L) requires an array of integer indices, not floating point numbers"))
```

PR #1289, opened 2025-05-16
add max length mode for dumps (if using custom opt passes)

wsmoses:


giordano (approved):
Looks good to me, but where are you getting the long names from? :sweat_smile:

wsmoses:
@jit optimize="<the entire optimization string as I try to ablate what is causing a correctness issue>" foo()

PR #1290, opened 2025-05-16
Fix anytracedarray override

wsmoses:
hopefully fixes the sort issue

### DIFF
--- a/src/TracedUtils.jl
+++ b/src/TracedUtils.jl
@@ -22,8 +22,6 @@ ReactantCore.materialize_traced_array(x::AbstractArray) = x
 
 ReactantCore.materialize_traced_array(x::TracedRArray) = x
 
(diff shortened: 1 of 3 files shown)


mofeing (approved):
Im out of town and won't be able to access my laptop until Sunday, but if `methods(sort, (Vector{Union{}},))` doesn't show the reactant method and `sort(Vector{Union{}}())` doesn't crash, then everything should be fine.

wsmoses:
indeed all is well:
```
julia> using Reactant
Precompiling Reactant finished.
  1 dependency successfully precompiled in 16 seconds. 55 already precompiled.

julia> Vector{Union{}} <: Reactant.AnyTracedRArray
false

julia> methods(sort, (Vector{Union{}},))
# 1 method for generic function "sort" from Base:
 [1] sort(v::AbstractVector; kws...)
     @ Base.Sort sort.jl:1489

julia> sort(Vector{Union{}}())
Union{}[]

julia> 
```

PR #1297, opened 2025-05-19
feat: overload LinearAlgebra.lu

avik-pal:


avik-pal:
batched version needs batchopinterface for gather

wsmoses (approved):

PR #1300, opened 2025-05-21
Support abstractfloat track numbers

wsmoses:


jumerckx (approved):
This looks correct to me. What's the use case of only tracking AbstractFloat and not Integer?

PR #1306, opened 2025-05-21
TracedToTypes fixes

jumerckx:
I encountered the make_tracer on `Memory` during function call insertion development, I'm not entirely sure it's still needed so if wanted I can split it off from this and test better.

### DIFF
--- a/src/Tracing.jl
+++ b/src/Tracing.jl
@@ -398,7 +398,10 @@ Base.@nospecializeinfer function traced_type_inner(
             }
         end
         error("Unsupported runtime $runtime")
-    elseif mode == TracedTrack || mode == NoStopTracedTrack || mode == TracedSetPath
+    elseif mode == TracedTrack ||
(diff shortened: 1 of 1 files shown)


src/Tracing.jl
```diff
@@ -1188,6 +1194,23 @@ function make_tracer(
...
+        @nospecialize(prev::Memory),
+        @nospecialize(path),
+        mode;
+        @nospecialize(sharding = Sharding.NoSharding()),
+        kwargs...,
+    )
+        if mode == TracedToTypes
+            return nothing
+        end
+        # TODO: does anything more need to be done here?
```
wsmoses, line 1209: I feel like we should consider this similarly to how we handle arrays?

jumerckx: Honestly I'm not sure at what point I needed this and I think we can do without, I've removed it.

jumerckx:
I *think* CI failure is just flaky aarch64?

PR #1307, opened 2025-05-21
Get rid of :result_and_mutated

jumerckx:
I'd added this but I believe we didn't actually need that.


wsmoses (approved):

jumerckx:
flaky CI?

jumerckx:
@wsmoses I think this can get merged?

PR #1308, opened 2025-05-21
Don't typeconstrain dict in `LinearAlgebra._diagm`

jumerckx:
With automatic function call insertion, `make_tracer(kv_updated, ...)` ends up getting called, which has trouble with the AnyTracedRArray type.


wsmoses:
so this would actually be nice to keep, what is the error being worked around?

jumerckx:
I just tested the function call insertion again without this change and it works. I can only assume whatever was happening has been fixed upstream already. closing this for now.

PR #1310, opened 2025-05-21
allow tobatch !== nothing if shapes match

jumerckx:
`error("This should not happen...")` got hit with automatic function call insertion and this seemed to fix it.
@avik-pal am I correct in thinking that all is well if the shape matches, I'm not understanding the tobatch logic fully.


avik-pal:
I don't think `to_batch` should be passed if the shapes are same no?

jumerckx:
> I don't think to_batch should be passed if the shapes are same no?

Right, I'll take a closer look at what was going on then.

PR #1311, opened 2025-05-22
`@trace if` return nothing in branches

jumerckx:
if the value if the result value in an if-statement is not required.
This is assuming use of the trace macro adheres to the docs:
```
  Currently Supported
  ===================

    •  if conditions (with elseif and other niceties) (@trace if ...)

    •  if statements with a preceeding assignment (@trace a = if ...) (note the positioning of the macro needs to be before the assignment and not before the if)
```


wsmoses:
@Pangoraw can you give this a final review (otherwise lgtm)

Pangoraw (approved):
Looks good!

PR #1321, opened 2025-05-23
Fix `verify_arg_names` when generated function has sizeof != 0

jumerckx:


lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -238,21 +242,23 @@ function trace_for(mod, expr; track_numbers)
...
+                verify_arg_names=$(verify_arg_names_sym),
             )
         end
     end
 
     return quote
-        if $(within_compile)() && $(any)(
-            $(is_traced), $(Expr(:tuple, cond_val.(all_syms.args[(begin + 1):end])...))
-        )
+        if any($(is_traced), $(Expr(:tuple, all_syms.args[(begin + 1):end]...)))
```
wsmoses, line 261: we still need the within compile

wsmoses (commented):
can we add a relevant test?

lib/ReactantCore/src/ReactantCore.jl
jumerckx: oh, that was an accident.

jumerckx:
Interestingly, the test seems to pass without the `Box` fix even.

wsmoses (approved):

PR #1326, opened 2025-05-25
Julia codegen compilation cache

wsmoses:


src/Compiler.jl
```diff
@@ -2989,6 +2989,10 @@ function compile_xla(f, args; client=nothing, serializable::Bool=false, kwargs..
     return results
 end
 
+# inspired by RuntimeGeneratedFunction.jl
+const __thunk_fwd_body_cache = Dict{Symbol,Expr}()
+const __thunk_rev_body_cache = Dict{Expr,Symbol}()
```
mofeing, line 2994: is this a bijective dictionary? i'm in the process of registering [bsc-quantic/BijectiveDicts.jl](https://github.com/bsc-quantic/BijectiveDicts.jl) or updating Bijections.jl to have the functionality of BijectiveDicts.jl, which could be handy.

in any case, lgtm

src/Compiler.jl
```diff
@@ -3102,10 +3104,18 @@ function compile(f, args; sync=false, kwargs...)
         display(mlir_fn_res.donated_args_mask)
     end
 
+    fname = if body in keys(__thunk_rev_body_cache)
+        __thunk_rev_body_cache[body]
```
mofeing, line 3108: what if we compile the same code but with different array sizes? wouldn't the 2nd `Expr` be wrong because it will point to a XLA-compiled function that expects other array sizes?

test/basic.jl
```diff
@@ -22,6 +22,16 @@ using InteractiveUtils
...
 end
 
+@testset "Julia Compilation cache" begin
+    x = @compile -(Reactant.to_rarray(ones(2)))
+    y = @compile -(Reactant.to_rarray(ones(2)))
+
+    @test typeof(x) == typeof(y)
+    # TODO, currently x and y are not equal as x.exec != y.exec
+    # as the executable we generate is itself not cached
+    # (which clearly we should do to improve jit time)
```
mofeing, line 32: following the comment above, what if we add this test...

```suggestion
    x = @compile -(Reactant.to_rarray(ones(2)))
    y = @compile -(Reactant.to_rarray(ones(2)))
    z = @compile -(Reactant.to_rarray(ones(3)))
    
    @test x(Reactant.to_rarray(ones(2))) == -ones(2)
    @test y(Reactant.to_rarray(ones(2))) == -ones(2)
    @test z(Reactant.to_rarray(ones(3))) == -ones(3)

    @test typeof(x) == typeof(y)
    # TODO, currently x and y are not equal as x.exec != y.exec
    # as the executable we generate is itself not cached
    # (which clearly we should do to improve jit time)
```

wsmoses: so that is independent of this pr [just that re compiles produce equivalent fns], so I presume tested elsewhere?

src/Compiler.jl
mofeing: mmm aha. i think i was getting confused because `body` contains the `xla_call_code` Expr and before, the function address was hardcoded into it. But now i see that that's no longer the case so it should be alright.

Issue #1346, opened 2025-05-28
Documentation on core functions


MilesCranmer:
I think `@compile` and `@jit` deserve a bit more documentation. I wasn't sure what these keywords meant:

![Image](https://github.com/user-attachments/assets/a9e92b05-3983-4aaa-9063-9862baabfac9)

mofeing:
Last week we agreed to stat doing a little _doc_-athon after every weekly meeting (probably starting next week).

Keeping up with the docs has been a challenge due to big API changes that have been ongoing.

Just for completion, `@compile` is just a helper macro to call `Reactant.compile`. And `@jit` literally just calls `@compile` and executes.

PR #1348, opened 2025-05-28
docs: document the options to core macros

avik-pal:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/1346

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -2144,11 +2144,18 @@ const COMMON_COMPILE_OPTIONS = Dict{Symbol,Any}(
 )
 
 const COMMON_COMPILE_OPTIONS_DOCS = """
-  - `optimize`: Optimizations passes to run on the traced MLIR code. For most users the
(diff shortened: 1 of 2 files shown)


src/Compiler.jl
```diff
@@ -2127,29 +2127,85 @@ function compile_mlir!(
...
+    :reshape_propagate => :(:up),
+    :optimize_then_pad => true,
+    :optimize_communications => true,
+    :cudnn_hlo_optimize => false,
+)
+
+const COMMON_COMPILE_OPTIONS_DOCS = """
+  - `optimize`: Optimizations passes to run on the traced MLIR code. For most users the
+    default value of `true` (or `:all`) should be sufficient. If you want to disable
+    optimizations, you can set it to `false` or `:none`. Other options are: `:before_kernel`,
```
wsmoses, line 2149: it also takes a string which is the exact optimization passes to run. So I think the three classes are bool / symbol / string

wsmoses (approved):

PR #1350, opened 2025-05-28
feat: support return type annotations for autodiff calls

avik-pal:
@wsmoses PoC of what I mentioned earlier


avik-pal:
We have ignore_derivatives now

PR #1353, opened 2025-05-29
fix: broadcasting with OHA

avik-pal:


avik-pal:
> We probably should add some enzyme jax level optimizations that turn the users of the materialize one hot array into gather scatters

This materialize probably needs to be scatter (probably faster for larger matrices), but currently our opts for that are kind of limited hence this broadcasting trick.

wsmoses (approved):

PR #1361, opened 2025-06-02
feat: fft rev rules + new complex number passes

avik-pal:


avik-pal:
test failures seem like a regression in Enzyme.jl gemm! differentiation (cc @wsmoses).

Reactant tests seem fine. Merging

wsmoses:
oh that sounds like me trying to do something for ffi, can fix that separately definitely merge this

PR #1364, opened 2025-06-03
feat: support for NNlib.scatter

JulianTrommer:
This PR implements the `NNlib.scatter` (and `NNlib.scatter!`) using `stablehlo.scatter`.

Since `stablehlo.scatter`has several restrictions on its inputs I had to `reshape` and `transpose` them based on their dimensions.

One issue currently is that the dispatch for `Statistics.mean` causes an ambiguity due to the [dispatch](https://github.com/FluxML/NNlib.jl/blob/1468582c4db5f18149cc8fff6fb4633c5debe5c5/src/scatter.jl#L85) in NNlib.jl:
```julia
for AT in (AbstractArray, AnyGPUArray)
    @eval function scatter!(op::typeof(mean), dst::$AT, src::$AT, idx::$AT)
        Ns = scatter!(+, zero(dst), one.(src), idx)
        dst_ = scatter!(+, zero(dst), src, idx)
        dst .+= safe_div.(dst_, Ns)
        return dst
    end
end
```
Since `ReactantNNlibExt` has no access to `Statistics` I could not write a specific dispatch. However since `NNlib.jl` uses the `+` operation for `mean` it uses the new working implementations (I also tested this). I was not sure how to handle this exception so I am open to any suggestions.

I also adapted the [test cases](https://github.com/FluxML/NNlib.jl/blob/1468582c4db5f18149cc8fff6fb4633c5debe5c5/test/testsuite/scatter.jl#L108) provided by `NNlib.jl` which all passed during my testing so the functionality is well tested (excluding the `mean` function I mentioned).

Please let me know if there is anything I overlooked!


ext/ReactantNNlibExt/Implementations.jl
```diff
@@ -474,3 +474,127 @@ function _nnlib_gather_impl(src::AnyTracedRArray, idxs::AbstractArray, n_dims::I
         slice_sizes=Int64[size(src)[1:n_dims]..., ones(Int64, ndims(src) - n_dims)...],
     )
 end
+
+# Scatter
+# The mean function currently produces an ambiguity due to
+# https://github.com/FluxML/NNlib.jl/blob/1468582c4db5f18149cc8fff6fb4633c5debe5c5/src/scatter.jl#L85
+# This could be resolved by an explicit dispatch which would require Statistics as dependency
```
avik-pal, line 481: This is where I would say use an `@reactant_overlay` and forward all scatter calls involving traced arrays to overloaded_scatter (we do it for conv)

avik-pal:
Can you give some context on the restrictions of shlo? In principle you can take the gather dims from the NNlib.gather impl and map it to the scatter dims (see https://github.com/EnzymeAD/Reactant.jl/blob/main/src/TracedRArray.jl#L453 for a similar case).

JulianTrommer:
By restrictions I mean the connection between the shapes of (dst, src, idx) and the dimension numbers for shlo.scatter:
```julia
scatter_dimension_numbers = MLIR.API.stablehloScatterDimensionNumbersGet(
        MLIR.IR.context(),
        length(update_window_dims), update_window_dims,
        length(inserted_window_dims), inserted_window_dims,
        length(input_batching_dims), input_batching_dims,
        length(scatter_indices_batching_dims), scatter_indices_batching_dims,
        length(scatter_dims_to_operand_dims), scatter_dims_to_operand_dims,
        index_vector_dim,
    )
```

I tried a lot of different configurations for these dimensions and did not find one or a calculation of it to get the same output as NNlib.scatter in regards to dimensions and values.
This is why the implementation reshapes and transposes the inputs to match the dimension numbers to get a single implementation for the different inputs written in the test cases.

I will take a look at the dims in TracedRArray to see if i can use them instead of static values, thanks for the hint!

wsmoses:
calling into higher level wrappers like ops.scatter (see https://github.com/EnzymeAD/Reactant.jl/blob/b1f56462f558849a327711c1188558a3d1c3ca0d/src/Ops.jl#L1708) should generally be preferred to calling the construct directly unless necessary. those wrappers also should generally handle most of the setup

JulianTrommer:
I tried using `TracedUtils.indices_to_gather_dims` as `scatter_dims` but i am running into the following issue:

For the first testset where the dims of `(dst,src,idx)` match (1D in this case) it also works with the `scatter_dims`. However if the dims do not match (e.g. `dims(idx) -1 = dims(src)` in the second testset) shlo.scatter errors:
```
error: expects updates tensor must be of rank 1 ( == rank-of('scatter_indices') - 1 + size-of('update_window_dims'), where 'scatter_indices' is expanded by a trailing 1 dimension if 'index_vector_dim' == rank-of('scatter_indices')), but got 2.
```
which makes sense if i look at the dimension numbers `indices_to_gather_dims` returns for the 1D indices:
```julia
scatter_dims.start_indices = TracedRArray{Int64,2N}((), size=(5, 1))
update_window_dims = Int64[]
inserted_window_dims = [1]
input_batching_dims = Int64[]
scatter_indices_batching_dims = Int64[]
scatter_dims_to_operand_dims = [1]
index_vector_dim = 2
```
I tried this with:
- no previous changes to the shapes of the arrays,
- adapting the reshaping from https://github.com/EnzymeAD/Reactant.jl/blob/ae1f9ef103789d9717a9a18a697624a64df85357/src/TracedRArray.jl#L392 and
- the reshaping in the current PR,
but could not get a solution without shlo.scatter failing.

ext/ReactantNNlibExt/Implementations.jl
JulianTrommer: Unfortunately this does not work since the scatter call with `mean` involving traced arrays should be forwarded to the implementation in NNlib and not to ReactantNNlibExt.
If I try to use `mean` as the mlir_function inside `_nnlib_scatter_impl` I get a different result than NNlib (since the reduction works differently than `+` for example), hence the additional implementation in NNlib.

Since NNlib already has Statistics as a dependency adding the package as a dependency to ReactantNNlibExt is a possible solution.

avik-pal: I see. Since NNLib already depends on statistics.jl, you can make the weakdep for NNlib into [NNlib, Statistics] and load Statistics here

avik-pal:
> I tried using TracedUtils.indices_to_gather_dims as scatter_dims but i am running into the following issue:

Did you try using the gather dims from https://github.com/JulianTrommer/Reactant.jl/blob/c701f9d3dcde9d16b5aa992b8d4f78a2441fbdee/ext/ReactantNNlibExt/Implementations.jl#L462? Indices to gather dims is very specific for setindex/getindex and the indexing pattern for NNlib.gather/scatter are quite different

JulianTrommer:
> Did you try using the gather dims from https://github.com/JulianTrommer/Reactant.jl/blob/c701f9d3dcde9d16b5aa992b8d4f78a2441fbdee/ext/ReactantNNlibExt/Implementations.jl#L462?

That did the trick for the first and second testset without using any `reshape` or `transpose` function.
The only remaining issue is with the third testset, there seems to be an issue if `src` and `dst` have different dimensions:

```julia
size(dst): (5,)
size(src): (3, 4)
size(idx): (1, 3, 4)
NNlib.scatter_dims: 0
update_window_dims: Int64[]
inserted_window_dims: [1, 2]
input_batching_dims: Int64[]
scatter_indices_batching_dims: Int64[]
scatter_dims_to_operand_dims: [2]
index_vector_dim: 1
```
```
error: Expects rank-of operand to match size-of('update_window_dims') + size-of('inserted_window_dims') + size-of('input_batching_dims') i.e. 2 but got 1.
```

The dimension numbers might need to be adjusted based on the dimension difference.

JulianTrommer:
After more testing i found out that unfortunately the `gather_dims` from https://github.com/JulianTrommer/Reactant.jl/blob/c701f9d3dcde9d16b5aa992b8d4f78a2441fbdee/ext/ReactantNNlibExt/Implementations.jl#L462 do not work.
If the `idx` array contains duplicate entries (which is not the case for the first and second testset, but for the third) they are ignored, i.e. not reduced correctly. An example of this:

```julia
dst = Float32[3, 4, 5, 6, 7]
src = ones(Float32, 5)
idx = [4, 1, 1, 5, 3] # instead of [4, 2, 1, 5, 3]

y = @jit(NNlib.scatter!(+, Reactant.to_rarray(dst), Reactant.to_rarray(src), idx))
# y = Float32[4, 4, 6, 7, 8] instead of Float32[5, 4, 6, 7, 8]
```

ext/ReactantNNlibExt/Implementations.jl
JulianTrommer: Solved by https://github.com/EnzymeAD/Reactant.jl/pull/1364/commits/196cce64f206f7cbcc8c6d3290ec7adf5b031177

avik-pal:
Thanks for getting this going, I added a few patches to your PR and merged it. The failure for the non-unique indices seems to be a bad pass, that I will fix upstream

PR #1368, opened 2025-06-04
fix: use missing instead of empty string for no jll

avik-pal:


src/mlir/MLIR.jl
```diff
@@ -10,7 +10,8 @@ module API
     const mlir_c = if Reactant_jll.is_available()
         Reactant_jll.libReactantExtra
     else
-        ""
+        @error "No compatible Reactant_jll found for $(Reactant_jll.host_platform)"
```
giordano: Not super sure the logging message is useful: for example users who install Reactant on Windows would get the message and can't do anything about it, it's not really actionable.

avik-pal: rn it will also crash on precompile.

Should we check is_available and conditionally register symbol names on julia end to avoid the crash?

giordano: > rn it will also crash on precompile.

It used to work in  #1003, did it regress?

avik-pal: yeah I did not add the check for the LinearAlgebera module which was added later. Let me do that instead

giordano (approved):

PR #1370, opened 2025-06-04
docs: new sharding docs

avik-pal:


docs/src/tutorials/sharding.md
```diff
@@ -0,0 +1,152 @@
...
+
+# TODO more complex tutorial describing replicated
+
+## Sharding in Neural Networks
+
+### 8-way Batch Parallelism
+
+### 4-way Batch & 2-way Model Parallelism
+
+## Related links
```
giordano, line 299: Add also https://jax-ml.github.io/scaling-book/sharding/? I think the visualisations in that book are quite good, better than in the JAX docs

docs/src/tutorials/sharding.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Handling Replicated Tensors
+
+## Sharding in Neural Networks
+
+### 8-way Batch Parallelism
+
+### 4-way Batch & 2-way Model Parallelism
+
+## Related links
```
giordano, line 299: I'm not sure this is relevant here since that's about hardware rather than software (although it's some hardware which is  _very_ related to XLA), but recently I came across this very good overview about TPUs: https://henryhmko.github.io/posts/tpu/tpu.html

docs/src/tutorials/sharding.md
```diff
@@ -0,0 +1,304 @@
...
+<!--
+TODO describe how arrays are the "global data arrays, even though data is itself only stored
+on relevant device and computation is performed only devices with the required data
+(effectively showing under the hood how execution occurs)
+-->
+
+<!--
+TODO make a simple conway's game of life, or heat equation using sharding simulation example
+to show how a ``typical MPI'' simulation can be written using sharding.
+-->
```
giordano, line 132: All the TODOs, also the ones below, are not hidden: https://enzymead.github.io/Reactant.jl/stable/tutorials/sharding

PR #1372, opened 2025-06-04
fix: support newer XLA [JLL]

avik-pal:


wsmoses (approved):

avik-pal:
shardy is busted, needs some investigation

PR #1373, opened 2025-06-04
feat: ignore derivatives

### LINKED ISSUE #1287: Can I ignore derivatives on some variables?
Maybe this is related to #487. I want something like this:

```
Zygote.gradient(x->x-ChainRulesCore.ignore_derivatives(x), 1f0)[1] ≈ 1f0
```

With Enzyme.jl on CPU, I can achieve the equivalent by writing a custom rule (EnzymeRules).
But is it possible with Reactant+Enzyme?

avik-pal:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/1287


wsmoses (commented):
we'll eventually move the ignore_derivatives into the enzyme proper api (and override here), fyi, but otherwise lgtm

PR #1374, opened 2025-06-05
`@trace` while and control flow tutorial

### LINKED ISSUE #1238: Traced while
realized we didn't have this. Since traced for becomes a while anyways I imagine this isn't that bad

### LINKED ISSUE #1317: malformed for loop assignment
Seeing an error when adding `@trace`, the function to reproduce is this
```julia
function temp(h::AbstractMatrix,η::AbstractMatrix,H::AbstractMatrix)
    m,n = size(h)
    @boundscheck (m,n) == size(η) || throw(BoundsError())
    @boundscheck (m,n) == size(H) || throw(BoundsError())

    @inbounds @trace for i in eachindex(η)
        h[i] = η[i] + H[i]
    end
end
```
and you don't even need to call the function, just trying to define it is enough to give the error. The specific message I see is
```
ERROR: LoadError: malformed for loop assignment
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] trace_for(mod::Module, expr::Expr; track_numbers::Type)
   @ ReactantCore ~/.julia/packages/ReactantCore/XJSZ9/src/ReactantCore.jl:173
 [3] var"@trace"(__source__::LineNumberNode, __module__::Module, args::Vararg{Any})
   @ ReactantCore ~/.julia/packages/ReactantCore/XJSZ9/src/ReactantCore.jl:160
in expression starting at REPL[6]:6
```

Pangoraw:
`@trace for` is now implemented with `@trace while`

Closes #1238 
Closes #1317

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Reactant"
 uuid = "3c362404-f566-11ee-1572-e11a4b42c853"
 authors = ["William Moses <wmoses@mit.edu>", "Valentin Churavy <vchuravy@mit.edu>", "Sergio Sánchez Ramírez <sergio.sanchez.ramirez@bsc.es>", "Paul Berg <paul@plutojl.org>", "Avik Pal <avikpal@mit.edu>", "Mosè Giordano <mose@gnu.org>"]
-version = "0.2.127"
+version = "0.2.128"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
@@ -90,7 +90,7 @@ PythonCall = "0.9"
 Random = "1.10"
 Random123 = "1.7"
 ReactantCore = "0.1.12"
-Reactant_jll = "0.0.196"
+Reactant_jll = "0.0.198"
 ScopedValues = "1.3.0"
 Scratch = "1.2"
 Sockets = "1.10"
--- a/deps/ReactantExtra/.bazelrc
+++ b/deps/ReactantExtra/.bazelrc
@@ -23,7 +23,7 @@ build:cuda --repo_env=HERMETIC_CUDA_VERSION="12.8.1"
 build:cuda --repo_env=HERMETIC_CUDNN_VERSION="9.8.0"
 # "sm" means we emit only cubin, which is forward compatible within a GPU generation.
 # "compute" means we emit both cubin and PTX, which is larger but also forward compatible to future GPU generations.
-build:cuda --repo_env HERMETIC_CUDA_COMPUTE_CAPABILITIES="sm_50,sm_60,sm_70,sm_80,compute_90"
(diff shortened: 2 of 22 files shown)


test/control_flow.jl
```diff
@@ -653,6 +653,38 @@ end
     @test x.c ≈ x_ra.c
 end
 
+function for_eachindex(s, x)
+    for i in eachindex(x)
```
avik-pal: missing a `@trace` or was it intentional?

Pangoraw: good catch, thanks!

docs/src/tutorials/control-flow.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+We can confirm by compiling our function and noticing that the result does not
+depend on the argument provided to the compiled function.
+
+```julia
+x = Reactant.ConcreteRArray(randn(Float32, 100))
+
+maybe_square_compiled = @compile maybe_square(true, x)
+maybe_square_compiled(false, x) == maybe_square_compiled(true, x)
```
wsmoses, line 35: Showing code_hlo here would also be good

docs/src/tutorials/control-flow.md
```diff
@@ -0,0 +1,94 @@
...
+    end
+    return x
+end
+```
+
+## Data-dependent Control Flow using [`@trace`](@ref)
+
+During tracing the arrays contain no data and only information about their shape
+and data type. As such, it is not possible to execute conditions that would
+depend on the value of an array. For this cases, ReactantCore provides the
```
wsmoses: For these cases*

docs/src/tutorials/control-flow.md
```diff
@@ -0,0 +1,94 @@
...
+[`@trace`](@ref) macro to allow capturing control flow expressions in the
+compiled program.
+
+Taking our same function from before and adding the [`@trace`](@ref) macro
+before the if expression will allow our compiled function to contain the
+condition.
+
+```@example control_flow_tutorial
+using Reactant
+
```
wsmoses, line 82: I’m not sure this example is what you want fully — since if the cond is a regular and not concrete book it’s still baked into the function

wsmoses: Ah I see it later — we should also explain the semantics of partially evaluating non concrete numbers

lib/ReactantCore/src/ReactantCore.jl
giordano, line 1: I have the feeling this broke GB-25, running the code is timing out now: https://github.com/PRONTOLab/GB-25/actions/runs/15510926515/job/43671788978.  Only difference I see compared to previous successful jobs is upgrade of ReactantCore

PR #1379, opened 2025-06-08
fix: ignore_derivatives need to preserve structure

avik-pal:


avik-pal:
merging to unblock an upstream issue

wsmoses (commented):
retroactive approval

PR #1381, opened 2025-06-09
`@trace`: Promote loop variables to traced numbers

Pangoraw:


lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -292,7 +292,14 @@ function trace_for(expr; track_numbers)
     quote
         local $start_sym, $limit_sym, $step_sym
         $bounds_defs
-        local $counter = 0
+
+        if within_compile()
+            $start_sym = Reactant.TracedUtils.promote_to(Reactant.TracedRNumber{Reactant.unwrapped_eltype(typeof($start_sym))}, $start_sym)
+            $limit_sym = Reactant.TracedUtils.promote_to(Reactant.TracedRNumber{Reactant.unwrapped_eltype(typeof($limit_sym))}, $limit_sym)
+            $step_sym = Reactant.TracedUtils.promote_to(Reactant.TracedRNumber{Reactant.unwrapped_eltype(typeof($step_sym))}, $step_sym)
```
Pangoraw: it's not the best to use Reactant from ReactantCore here, so if anyone has a better idea, i'll take it

wsmoses:
The failures seem new?

wsmoses:
@Pangoraw minor but looking at the docs, can we change the else in the control flow to be x^3 or something, since otherwise Reactant compiles out the identity to not return anything and reuse the input instead of call with mlir:  https://enzymead.github.io/Reactant.jl/dev/tutorials/control-flow

PR #1391, opened 2025-06-11
feat: register various stablehlo passes

avik-pal:


deps/ReactantExtra/API.cpp
```diff
@@ -1189,6 +1193,33 @@ extern "C" void InitializePasses(MlirDialectRegistry creg) {
   xla::sdy::registerStablehloExportPipeline();
   xla::sdy::registerStablehloImportPipeline();
   xla::sdy::registerStablehloImportShardingsPass();
+
+  // SHLO passes
+  stablehlo::registerStablehloAggressiveSimplificationPass();
```
wsmoses: we should probably move a lot of this into a function in RegistryUtils.cpp in Enzyme-JaX so we can more easily debug with enzymexlamlir-opt when things go wrong

avik-pal: https://github.com/EnzymeAD/Enzyme-JAX/pull/1026

avik-pal:
https://github.com/EnzymeAD/Reactant.jl/actions/runs/15593266258/job/43917511693?pr=1391#step:11:349

wsmoses (approved):

wsmoses:
hopefully this also means we can actually run the sharding passes in enzymexlamlir-opt now to debug

PR #1396, opened 2025-06-12
[docs] Improve some information about local builds of libReactant

giordano:


docs/src/tutorials/local-build.md
```diff
@@ -84,8 +84,18 @@ To do that using this `build_local.jl` script pass the options `--extraopt '--st
...
+Since [Bazel interacts badly with ccache](https://github.com/ccache/ccache/discussions/1279) when the latter [masquerades itself as the compiler via symlinks](https://ccache.dev/manual/4.11.3.html#_run_modes), the most reliable way to use ccache with Bazel is to create a small shell wrapper like
 
+```bash
+#!/bin/bash
+/path/to/ccache /path/to/your/compiler "${@}"
+```
+
+replacing `/path/to/ccache` with the path where ccache was installed, and `/path/to/your/compiler` with the path of your compiler.
+Then inform Bazel to use this script as the compiler (with the `build_local.jl` script add the flag `--cc /path/to/ccache/wrapper`, where `/path/to/ccache/wrapper` is the path where you saved the shell wrapper).
+Finally, you may have to add the flag `--extraopt "--sandbox_writable_path=/path/to/ccache/directory"`, otherwise Bazel will not let `ccache` write the cache output to its own directory.
```
giordano, line 98: I think this is the most comprehensive piece of documentation about using ccache with bazel that you can find anywhere on the internet :grin:

PR #1401, opened 2025-06-13
Change ReactantCore so that it doesn't depend on Reactant

ptiede:
I was having an issue using ReactantCore in a package that didn't include Reactant. 

I believe the problem is in the trace_for function since in promote_to it explicitly calls Reactant which isn't loaded in the ReactantCore scope so it causes an error. 

To fix this I made a `promote_to_traced` function which is then defined in Reactant proper. 

Not sure if this is the best way to fix this but it seemed pretty minimal.


wsmoses (approved):

wsmoses:
also please update the various project.toml versions

Project.toml
```diff
@@ -89,7 +89,7 @@ Preferences = "1.4"
 PythonCall = "0.9"
 Random = "1.10"
 Random123 = "1.7"
-ReactantCore = "0.1.12"
+ReactantCore = "0.1.14"
```
giordano: I think this should be
```suggestion
ReactantCore = "0.1.15"
```
and the version of `ReactantCore` should be bumped accordingly

wsmoses: yeah the rebase also bumped so got removed, fixing

PR #1402, opened 2025-06-14
Update jll

wsmoses:


avik-pal:
On phone but remove gather simplify, the consprop pattern supercedes it

PR #1407, opened 2025-06-18
refactor: centralize compile options

avik-pal:


wsmoses (approved):
will the existing string of passes get parsed in here?

avik-pal:
Yeah, there is no change on that end

PR #1408, opened 2025-06-18
perf: run ablations for the paper

avik-pal:


avik-pal:
## DeepONet

image

## Fourier Neural Operator

image

Need to investigate if the pass pipelines are being correctly set the results seem quite strange honestly

avik-pal:
FNO case is heavily dominated by FFTs so nothing much we can do there, but the gradients with passes enabled being consistently slower seems off

wsmoses:
I assume this is all straight line (no control flow like while), and we have no fft specific optimizations yet

in which case this also becomes a question of how well xla’s allocation and other passes may recognize certain patterns

wsmoses:
But also concurrently we built the passes for the primal and post jax differentiation , so we may be missing optimizations relevant to enzyme generated programs

avik-pal:
### Default XLA

image

### Reactant

image

Currently our lack of fft - transpose optimizations is preventing fusions (see the overhead for the elementwise ops)

avik-pal:
image

PR #1430, opened 2025-06-30
[CI] Disambiguate titles and branch names for formatting PRs

giordano:
Ref: https://github.com/EnzymeAD/Reactant.jl/pull/1428#issuecomment-3018068147.


giordano:
This should be quite uncontroversial.  I'm merging so that the jobs running in 10 minutes will have the right names and won't clash.

PR #1441, opened 2025-07-03
Introduce `transmute`, `transmute_type` functions as future names for `make_tracer`, `traced_type`

mofeing:
We also should rename `promote_traced_type` but not sure if to `promote_transmute_type` or `promote_transmuted_type`.

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -1018,7 +1018,7 @@ function Reactant.make_tracer(
 )
     x = Base.unsafe_pointer_to_objref(Base.reinterpret(Ptr{Cvoid}, prev.ptr))
     x = x::TracedRArray
-    Reactant.make_tracer(seen, x, path, mode; kwargs...)
+    Reactant.transmute(seen, x, path, mode; kwargs...)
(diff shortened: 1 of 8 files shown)


Pangoraw:
Can be a good use case for [`@deprecate`](https://docs.julialang.org/en/v1/base/base/#Base.@deprecate). it would require moving the implementation to the new name and marking the old name with it.

mofeing:
> Can be a good use case for [`@deprecate`](https://docs.julialang.org/en/v1/base/base/#Base.@deprecate). it would require moving the implementation to the new name and marking the old name with it.

The main reason why I haven't marked like that yet is because it may trigger a lot of depwarn messages right? Also, we cannot just use `@deprecate` because it will break specializations of `make_tracer` and `traced_type_inner`. We agreed to put these names and finally rename everything on the 0.3 release.

Pangoraw:
> As of Julia 1.5, functions defined by @deprecate do not print warning when julia is run without the --depwarn=yes flag set, as the default value of --depwarn option is no. The warnings are printed from tests run by Pkg.test().

this is the current state of things

PR #1444, opened 2025-07-03
ProbProg: Making trace an operand

sbrantq:


src/ProbProg.jl
```diff
@@ -295,25 +575,28 @@ function call_internal(f::Function, args::Vararg{Any,Nargs}) where {Nargs}
     return result
 end
 
-function generate(f::Function, args::Vararg{Any,Nargs}; constraints=nothing) where {Nargs}
-    trace = ProbProgTrace(f, (args...,))
+function simulate(f::Function, args::Vararg{Any,Nargs}) where {Nargs}
+    old_gc_state = GC.enable(false)
```
wsmoses: move into the runtime functions?

src/Compiler.jl
```diff
@@ -1231,6 +1231,7 @@ end
 # TODO we want to be able to run the more advanced passes via transform dialect as an enzyme intermediate
 # However, this errs as we cannot attach the transform with to the funcop itself [as we run a functionpass].
 const enzyme_pass::String = "enzyme{postpasses=\"arith-raise{stablehlo=true},canonicalize,cse,canonicalize,remove-unnecessary-enzyme-ops,enzyme-simplify-math,canonicalize,cse,canonicalize\"}"
+const probprog_pass::String = "probprog{postpasses=\"arith-raise{stablehlo=true},canonicalize,cse,canonicalize,remove-unnecessary-enzyme-ops,enzyme-simplify-math,canonicalize,cse,canonicalize\"}"
```
wsmoses: do you need these postpasses for probprog?

sbrantq:
@wsmoses So I added `GC.@preserving`s to fix the deadlock issue with the ConcreteRArray that specifies `rng.seed` (for context GC triggered by `EnzymeJaXMapSymbol`'d functions somehow tries to free `only(rng.seed.data).buffer`): https://github.com/EnzymeAD/Reactant.jl/blob/9bd1dee3337b5a61b66e2f6a636bdc66c559ae4a/src/probprog/Modeling.jl#L283-L285

Also GC triggered by mapped functions in `src/probprog/FFI.jl` can cause `jl_gc_wait_for_the_world` to wait forever for the main Julia thread which is `wait`ing during `Base.convert` to Julia Array, like this one https://github.com/EnzymeAD/Reactant.jl/blob/9bd1dee3337b5a61b66e2f6a636bdc66c559ae4a/src/probprog/Modeling.jl#L292
So I added
https://github.com/EnzymeAD/Reactant.jl/blob/9bd1dee3337b5a61b66e2f6a636bdc66c559ae4a/src/probprog/Modeling.jl#L287-L289
(Adding trivial `Base.isready` for ConcretePJRTArray and ConcreteIFRTArray).

Does that sound like an ok-ish fix, or am i missing something?

sbrantq:
@wsmoses Can we merge this large execution trace PR first and add new MH and MCMC interfaces into a new PR?

avik-pal:
We would need a rebase, the test failures should have been fixed a few weeks back. Also let's get the jll changes in first as a separate PR?

PR #1451, opened 2025-07-11
Add copy handler and linkable version

wsmoses:


wsmoses:
@avik-pal the distributed api's changed a bit, we'll need to adapt julia side.

I also got an actual copy api working which is nice

PR #1457, opened 2025-07-14
Don't use `fmap` in traced ifs

jumerckx:
`MissingTracedValue`s are only created as top-level objects in the if-else return tuple AFAICT. This means we don't need to recurse through the whole object structure and can simply map the outer level directly.
(@glou-nes)

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -13,7 +13,6 @@ using ..Reactant:
     MissingTracedValue,
     unwrapped_eltype
 using ReactantCore: ReactantCore
-using Functors: fmap
 
 function mlir_type(x::Union{RNumber,RArray})::MLIR.IR.Type
     return MLIR.IR.TensorType(collect(Int, size(x)), MLIR.IR.Type(unwrapped_eltype(x)))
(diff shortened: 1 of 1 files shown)


src/Ops.jl
```diff
@@ -2376,15 +2375,16 @@ end
...
-    corrected_traced_results = fmap(traced_false_results, traced_true_results) do fr, tr
-        if fr isa MissingTracedValue && tr isa MissingTracedValue
-            return fr
-        elseif fr isa MissingTracedValue
-            return tr
-        else
-            return fr
+    corrected_traced_results =
+        map(zip(traced_false_results, traced_true_results)) do (fr, tr)
+            if fr isa MissingTracedValue && tr isa MissingTracedValue
```
wsmoses, line 2380: I don't think this is right since corrected_traced_results is returned from here? or am I missing something

cc @Pangoraw @avik-pal

jumerckx: I don't see the issue yet. As long as MissingTracedValues are the outermost object, the code with `map` and `fmap` should be equivalent I think?

wsmoses: actually wait I understand, okay this seems reasonable then, but one of @Pangoraw / @avik-pal should also review

wsmoses:
going to just go ahead and merge for now, @Pangoraw @avik-pal please comment with anything post commit

PR #1471, opened 2025-07-22
test: compile with more devices than available

avik-pal:


test/sharding.jl
```diff
@@ -459,3 +459,14 @@ end
         @warn "Not enough addressable devices to run sharding tests"
     end
 end
+
+@testset "Compile-Only with More Devices" begin
+    mesh = Sharding.Mesh(zeros(Int64, 2, 4), (:x, :y))
```
wsmoses, line 464: presumably shouldn't you also check the number of devices here as a prereq?

avik-pal: `zeros` is forcing the "fake"-devices

wsmoses: ah got it

wsmoses (approved):

PR #1478, opened 2025-07-26
[CI] Prepare setup for new runners

giordano:


giordano:
@wsmoses the setup is ready from my point of view, the only problem is that the greatly reduced parallelism would actually make things slower overall in case of multiple concurrent jobs.

giordano:
Ok, this is ready.  New runners aren't currently used, but the setup has been slightly adapted to work with them, so they can just be added later.

Issue #1462, opened 2025-07-16
In-place mutation allocates despite `assert_nonallocating=true`


alexandergunnarson:
I'm trying to simply write a vector to a column-oriented matrix. What I have works for small sizes, but it allocates despite `assert_nonallocating=true`.
```julia
# NOTE — There's probably a simpler way to do this that's not so low-level. But maybe not, because
#        XLA is immutable by default.
function write_row!(xs_mut, i, v)
    v_col = Reactant.Ops.reshape(v, [length(v), 1])
    updated_xs = Reactant.Ops.dynamic_update_slice(xs_mut, v_col, [1, i])

    # Mutatively update xs with the new data
    Reactant.TracedUtils.set_mlir_data!(xs_mut, Reactant.TracedUtils.get_mlir_data(updated_xs))

    nothing
end

# Same issue with this
function write_row_simple!(xs_mut, i, v)
    xs_mut[:, i] = v
    nothing
end
```

I compile and run like so
```Julia
T = Float32
dim = 768
max_n = 10_000_0
vectors_gpu = Reactant.@jit Reactant.Ops.fill(T(0), (dim, max_n))
i_default = 1 # just to have a value to pass into `write_row_rx!`
vector_gpu = Reactant.@jit Reactant.Ops.fill(T(0), (dim,))

write_row_rx! = Reactant.@compile assert_nonallocating=true sync=true write_row!(
    vectors_gpu, i_default, vector_gpu
)

write_row_rx!(vectors_gpu, 1, vector_gpu)
```

And get
```
julia> write_row_rx!(vectors_gpu, 1, vector_gpu)
2025-07-16 22:08:55.628804: W external/xla/xla/tsl/framework/bfc_allocator.cc:501] Allocator (GPU_0_bfc) ran out of memory trying to allocate 28.61GiB (rounded to 30720000000)requested by op
If the cause is memory fragmentation maybe the environment variable 'TF_GPU_ALLOCATOR=cuda_malloc_async' will improve the situation.
Current allocation summary follows.
Current allocation summary follows.
2025-07-16 22:08:55.628872: I external/xla/xla/tsl/framework/bfc_allocator.cc:1049] BFCAllocator dump for GPU_0_bfc
[...]

PR #1496, opened 2025-07-28
Default to post conversion to dynamic update slice

wsmoses:


src/Compiler.jl
```diff
@@ -691,7 +691,7 @@ const SUM_TO_CONV = Ref(false)
 const AGGRESSIVE_SUM_TO_CONV = Ref(false)
 const AGGRESSIVE_PROPAGATION = Ref(false)
 const DUS_SLICE_SIMPLIFY = Ref(true)
-const CONCATS_TO_DUS = Ref(false)
+const CONCATS_TO_DUS = Ref(true)
```
giordano, line 694: 😱

test/indexing.jl
```diff
@@ -47,6 +47,19 @@ end
...
 end
 
+function write_row_simple!(xs_mut, i, v)
+   xs_mut[:, i] = v
+   nothing
+end
+   
+@testset "setindex: DUS" begin
+    x_ra = similar(ConcreteRArray{Float64}, (76, 100))
+    y_ra = similar(ConcreteRArray{Float64}, (76))
```
giordano: If the second argument needs to be a tuple, then you need to use
```suggestion
    y_ra = similar(ConcreteRArray{Float64}, (76,))
```
If not, can restore the previous
```suggestion
    y_ra = similar(ConcreteRArray{Float64}, 76)
```
(which is equivalent to `(76)`)

src/Compiler.jl
wsmoses: benefit is x/ref https://github.com/EnzymeAD/Reactant.jl/issues/1462

wsmoses: [assuming we ofc nicely fix dus comm and other things, so can revert/change/etc things]

jumerckx (approved):
This looks good to me, though I'm not sure what this CI failure is https://github.com/EnzymeAD/Reactant.jl/actions/runs/16577306883/job/46884662450?pr=1496#logs

giordano:
That's #1497.

PR #1502, opened 2025-07-29
Add a new method of getindex

jlk9:
Added a new method of `getindex` for an OffsetVector of tracedRNumber and tracedRArray and Base.oneto

@wsmoses

### DIFF
--- a/ext/ReactantOffsetArraysExt.jl
+++ b/ext/ReactantOffsetArraysExt.jl
@@ -44,6 +44,13 @@ function Base.getindex(
     return getindex(parent(x), offset_indices)
 end
 
+function Base.getindex(
+    x::OffsetVector{Reactant.TracedRNumber{T}, Reactant.TracedRArray{T, 1}}, indices::Base.OneTo{Int}
+) where {T}
+    offset_indices = indices .- x.offsets[1]
+    return getindex(parent(x), offset_indices)
+end
+
 parentindex(r::OffsetArrays.IdOffsetRange, i) = i .- r.offset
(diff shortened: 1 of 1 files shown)


ext/ReactantOffsetArraysExt.jl
```diff
@@ -44,6 +44,13 @@ function Base.getindex(
     return getindex(parent(x), offset_indices)
 end
 
+function Base.getindex(
+    x::OffsetVector{Reactant.TracedRNumber{T}, Reactant.TracedRArray{T, 1}}, indices::Base.OneTo{Int}
+) where {T}
+    offset_indices = indices .- x.offsets[1]
+    return getindex(parent(x), offset_indices)
+end
+
```
wsmoses, line 53: does the same fix work if you put this in the following function like:
```
function Base.getindex(
    a::OffsetArray{<:Reactant.TracedRNumber,N}, indices::Vararg{Union{Int,AbstractArray, Base.OneTo{Int}},N}
) where {N}
```

jlk9: Not quite, but this does:

```
a::OffsetVector{Reactant.TracedRNumber{T}, Reactant.TracedRArray{T, 1}}, indices::Vararg{Union{Int,AbstractArray, Base.OneTo{Int}},N}
) where {N}
```

wsmoses: oh bleh okay leave it as a separate function then [we shouldnt specialize that existing method to be 1dimensional]

PR #1505, opened 2025-07-29
[Experiment] use GPUCompiler instead of OpaqueClosures in `call_with_reactant`

glou-nes:
I messed up my last PR, I recreate it.

Interaction between codegen emitted `spec_func` and Julia is a bit hard (immutable types boxed, signature change) so for now, I'm using wrapped function `func`.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -690,8 +690,6 @@ function call_with_reactant_generator(
         offset += 1
     end
 
-    #TODO: replace 0 with function pointer using mi.cache.invoke
-    @warn mi.cache.invoke
     #force the creation of Any[fn_args...]
     fn_args_vec = push_inst!(
         Expr(:call, GlobalRef(Base, :getindex), GlobalRef(Base, :Any), fn_args...)


src/utils.jl
```diff
@@ -553,35 +481,86 @@ function rewrite_insts!(ir, interp, guaranteed_error)
...
-            else
-                Core.Compiler.setindex!(ir.stmts[idx], Core.Compiler.getindex(urs), :stmt)
-            end
-        end
+@static if isdefined(Core, :BFloat16)
+    nmantissa(::Type{Core.BFloat16}) = 7
+end
+nmantissa(::Type{Float16}) = 10
+nmantissa(::Type{Float32}) = 23
+nmantissa(::Type{Float64}) = 52
```
giordano: Why not using `precision` instead of defining a new function?

wsmoses: I think that is just moving around code from the end of the file, but fail point (and perhaps separate Pr)

PR #1506, opened 2025-07-29
Silence unused-command-line-argument warning, not just the error

giordano:
The warning is absolutely spammy.  We also disable the warning in BinaryBuilder,
it's so annoying:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/99594c5cae8f67b9cdc776418744d88e574d4e05/src/Runner.jl#L449.


giordano:
I don't really like making `--cc` optional, we have no way of telling what's the compiler and that's important for setting the right options.  I restored explicit compiler in [EnzymeAD/Enzyme-JAX@`e652e15` (#1197)](https://github.com/EnzymeAD/Enzyme-JAX/pull/1197/commits/e652e1570595f15e7fbc08a3bb64b7ecc402ae6e), that seemed to work in the little tests I did on tmate.

giordano:
Failures are also on `main`....

PR #1522, opened 2025-08-02
fix: force finalize the buffers after precompile

avik-pal:


wsmoses (approved):

wsmoses:
@avik-pal can you alos add here: https://github.com/EnzymeAD/Reactant.jl/blob/9add8dfdb662e17031b710d0b9a4b7b0d0d19193/ext/ReactantCUDAExt.jl#L1490

avik-pal:
tpu still crashes

wsmoses:
Well one step at a time I suppose

PR #1524, opened 2025-08-03
Don't own a null array

wsmoses:


src/xla/IFRT/AsyncArray.jl
```diff
@@ -3,7 +3,7 @@ mutable struct AsyncArray <: XLA.AbstractAsyncBuffer
     future::Union{Future,Nothing}
 end
 
-const AsyncEmptyArray = AsyncArray(Array(C_NULL), nothing)
+const AsyncEmptyArray = AsyncArray(Array(C_NULL, false), nothing)
```
giordano, line 6: Uh, how does this work?
```julia-repl
julia> Array(C_NULL)
ERROR: MethodError: no method matching Array(::Ptr{Nothing})
The type `Array` exists, but no method is defined for this combination of argument types when trying to construct it.
[...]

julia> Array(C_NULL, false)
ERROR: MethodError: no method matching Array(::Ptr{Nothing}, ::Bool)
The type `Array` exists, but no method is defined for this combination of argument types when trying to construct it.
[...]
```

giordano: Ah, this is https://github.com/EnzymeAD/Reactant.jl/blob/a835bf25da3b4647fb5d387c3f2e50351f252d44/src/xla/IFRT/Array.jl#L1-L9 So if you don't own it you expect the XLA runtime to free it?

wsmoses: or restated: only if we own it, do we have julia free it.

PR #1525, opened 2025-08-03
More precompile memory fixups

wsmoses:


ext/ReactantCUDAExt.jl
```diff
@@ -1489,7 +1489,17 @@ end
                 end
                 y = Reactant.ConcreteRArray([2.0]; client)
                 Reactant.Compiler.compile_mlir(square!, (y,); optimize=false)
-                finalize(y)
+
+                if y isa Reactant.ConcreteIFRTArray
+                    Reactant.XLA.free_buffer(y.data.buffer)
+                    y.data.buffer.buffer = C_NULL
+                else
+                    ntuple(Val(length(y.data))) do i
```
giordano: Why this isn't a simple for loop
```julia
for dat in y.data
```
?

wsmoses: imma be completely honest, I didn't think of that

giordano: One problem of the ntuple iteration is that it unnecessarily allocates memory for collections longer than 10. A for loop is also a lot less cumbersome to read, it took me a while to guess what was going on here.

PR #1526, opened 2025-08-03
Actually use c api

wsmoses:


giordano (approved):
This seems to be the only `std::string` appearing in the signature of a C extern function, so we should be good after this 👍

PR #1530, opened 2025-08-05
CI Try caching precompilation

jumerckx:
I'm not yet sure this does anything. But if this would work, we could split the tests in many more separate jobs.


.github/workflows/CI.yml
```diff
@@ -29,7 +29,32 @@ concurrency:
...
 
 jobs:
+  cache-warmup:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        version: ["1.10", "1.11"]
+        os: [ubuntu-24.04, ubuntu-22.04-arm, macOS-latest]
+        runtime: ["PJRT", "IFRT"]
+        assertions: [false, true]
```
giordano: For what is worth, I personally like expanding lists in the one-per-line format, like in https://github.com/EnzymeAD/Reactant.jl/blob/eca8191c63f926c77ddf409c72adf261b9a45b15/.github/workflows/CI.yml#L67-L87 which makes it easy to add remove individual items without messing up with git-blame.

jumerckx: Thanks, I'll take this into account.
I'm still mainly messing around with things for now.

Do you think it makes sense to have a cache warmup step with cache keys independent of jobs and testgroup?

giordano: If it works, it can make sense.

jumerckx:
packages are still being precompiled in the test step. I'm looking for a culprit, this seems suspicious:
```
┌ Warning: Could not use exact versions of packages in manifest, re-resolving
└ @ Pkg.Operations /opt/hostedtoolcache/julia/1.10.10/aarch64/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:1829

```

giordano:
You may want to drastically reduce the matrix (ideally to a single element) while you investigate how to make this work, otherwise you'll waste loads of resources and everybody's time.

PR #1532, opened 2025-08-05
fix: init for julia and shlo are semantically different

avik-pal:


avik-pal:
lets see if this fixes TPU

avik-pal:
need to do this for reduce window also (in scan)

src/TracedRArray.jl
```diff
@@ -1290,6 +1290,11 @@ function scan_impl!(
             op_in_T = typeof(init)
             input = typeof(init).(input)
         end
+    else
+        # TODO: fix this for TPUs
+        if contains(string(first(Reactant.devices())), "TPU")
```
wsmoses, line 1295: could you not just broadcast the init to the corresponding dimension, and then apply op to it in post?

Assuming the operation itself is commutative

wsmoses: this is somewhat dangerous to special case for TPU since it is out of spec in general for shlo

wsmoses:
this definitely improves things so going to quickly merge, but we should sort out reduce_window shortly

PR #1536, opened 2025-08-06
test: mark tests as broken instead of skipping them

avik-pal:
- all tests are now run with TPUs
- tests are marked with `skip` or `broken`. `skip` for ones which are not supported upstream and `broken` for cases we need to eventually fix
- temporary workaround for #1581
- reduce the number of tests in NN (some of those were not really needed)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,59 +38,59 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version:
-          - "1.10"
-          - "1.11"
-          # - 'nightly'
-        os:
-          - ubuntu-24.04
-          # `ubuntu-22.04-arm` is considered more stable than `ubuntu-24.04-arm`:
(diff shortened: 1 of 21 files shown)


avik-pal:
One final crash remaining https://github.com/EnzymeAD/Reactant.jl/actions/runs/17003494579/job/48209393626?pr=1536#step:12:1002

avik-pal:
Seems to crash in `@code_hlo` for enzyme gradient even before any TPU stuff... But again doesn't happen on a non-tpu machine (reproducible on colab as well)

avik-pal:
Something is broken in https://github.com/EnzymeAD/Enzyme-JAX/blob/095cdf65c04d1ebdcfb06cfec3175c0653d9c7ba/src/enzyme_ad/jax/Passes/ArithRaising.cpp#L293-L304

avik-pal:
All tests are finally passing now :tada:

wsmoses (approved):

PR #1553, opened 2025-08-13
No static libstdc++

wsmoses:


deps/ReactantExtra/WORKSPACE
```diff
@@ -83,6 +83,9 @@ sed -i.bak0 "1s/^/#define _USE_MATH_DEFINES\\n/g" xla/fp_util.h xla/hlo/builder/
 """,
     """
 sed -i.bak0 "s/Windows\\.h/windows\\.h/g" xla/tsl/platform/windows/port.cc xla/tsl/platform/windows/wide_char.cc xla/tsl/platform/windows/env.cc xla/tsl/platform/windows/windows_file_system.cc
+""",
+    """
+sed -i.bak0 "/D_FORTIFY_SOURCE/d" third_party/gpus/crosstool/cc_toolchain_config.bzl.tpl tools/toolchains/cross_compile/cc/BUILD xla/tools/toolchains/clang6/CROSSTOOL.tpl
```
giordano: As far as I understand, the file `xla/tools/toolchains/clang6/CROSSTOOL.tpl
Collapse comment` isn't there yet when this command is run, the other two exist instead.

giordano:
```suggestion
""",
    """
sed -i.bak0 "/D_FORTIFY_SOURCE/d" third_party/gpus/crosstool/cc_toolchain_config.bzl.tpl tools/toolchains/cross_compile/cc/BUILD
```

giordano: Now this compiles for me, but `-D_FORTIFY_SOURCE` is still there.  I don't know what's up with `xla/tools/toolchains/clang6/CROSSTOOL.tpl` 🤔

wsmoses: Ah it should be tools/ not xla/tools/

deps/ReactantExtra/API.cpp
```diff
@@ -207,6 +207,8 @@ template <typename T> HeldValue<T> *capture(T obj) {
 
 } // namespace reactant
 
+#define REACTANT_ABI extern "C" MLIR_CAPI_EXPORTED
```
giordano, line 210: I agree [`MLIR_CAPI_EXPORTED`](https://github.com/llvm/llvm-project/blob/e27e4f3a99911675215ca79eb95290a3ab77fe74/mlir/include/mlir-c/Support.h#L39-L43) is what you want to use, but if you enable this then you're also exposing LLVM MLIR API?

PR #1557, opened 2025-08-15
feat: persistent compilation cache

avik-pal:


deps/ReactantExtra/API.cpp
```diff
@@ -932,10 +942,34 @@ xla::CompileOptions
...
+
+    if (process_id <= 0) {
+      debug_options->set_xla_gpu_experimental_autotune_cache_mode(
+          xla::DebugOptions::AutotuneCacheMode::
+              DebugOptions_AutotuneCacheMode_AUTOTUNE_CACHE_MODE_UPDATE);
+    } else {
+      debug_options->set_xla_gpu_experimental_autotune_cache_mode(
+          xla::DebugOptions::AutotuneCacheMode::
+              DebugOptions_AutotuneCacheMode_AUTOTUNE_CACHE_MODE_READ);
+    }
```
giordano: So this should already work right now by setting `XLA_FLAGS`?

avik-pal: Yes. XLA however doesn't validate caches which is why it is not very safe to let the end user handle this.

docs/src/tutorials/persistent_compile_cache.md
```diff
@@ -0,0 +1,28 @@
...
+- `persistent_autotune_cache_enabled`: Whether to enable the autotuning cache. Defaults to
+  `true`.
+
+## Clearing the cache
+
+To clear the cache, you can reuse `Scratch.jl`'s `clear_scclear_scratchspaces!` function:
+
+```julia
+using Scratch, Reactant
+Scratch.clear_scratchspaces!(Reactant)
```
wsmoses: can we make a separate function for this, since we use the scratch space for other things like libtpu [and in the future intel runtime, etc]

giordano:
Anything holding this?

avik-pal:
Feel free to merge if tests pass. I will add a function for clearing
specific parts of the cache on Monday

docs/src/tutorials/persistent_compile_cache.md
avik-pal: done

wsmoses (approved):

PR #1558, opened 2025-08-15
Fix KA.allocate in ReactantKernelAbstractionsExt

oschulz:


ext/ReactantKernelAbstractionsExt.jl
```diff
@@ -22,7 +22,7 @@ function Base.getproperty(x::ReactantBackend, sym::Symbol)
     end
 end
 
-KA.allocate(n::ReactantBackend, ::Type{T}, dims::Tuple) where {T} = KA.zeros(b, T, dims)
+KA.allocate(b::ReactantBackend, ::Type{T}, dims::Tuple) where {T} = KA.zeros(b, T, dims)
```
wsmoses: While definitely correct, we incidentally now actually support similar(ConcreteRArray), if you wouldn't mind changing this to (and adding a clearly missing test for this)

wsmoses (approved):

ext/ReactantKernelAbstractionsExt.jl
oschulz: So something like

```
KA.allocate(b::ReactantBackend, ::Type{T}, dims::Tuple) where {T} = similar(ConcreteRArray, T, dims)
```

?

One question about this: In both versions, the new array will be created on the current default Reactant backend. In principle, one can use Reactant on both CPU and GPU in the same process, right? Shouldn't `ReactantBackend` carry some information on the device, like `MLDataDevices.ReactantDevice` does, so that the new array can be created on the correct device (also what about sharding config, etc.)?

Oh, and while we're at it - should we change the implementation of

```
function KA.zeros(::ReactantBackend, ::Type{T}, dims::Tuple) where {T}
    return Reactant.to_rarray(zeros(T, dims))
end
```

to use `KA.allocate` followed by a `fill!` or so, instead of first allocating an array on the host?

wsmoses: yeah that would be a good idea

wsmoses: and yeah also that seems wise too

oschulz: Ok, I'll see if I can figure this out. :-)

oschulz: It looks like some of the code to do this is in https://github.com/LuxDL/Lux.jl/blob/main/lib/MLDataDevices/ext/MLDataDevicesReactantExt.jl

Would it make sense to have a Reactant device type that includes client, device and sharding information in Reactant itself, so that KernelAbstractions and MLDataDevices could just kind of wrap it? I get the impression that doing array creation and conversion correctly needs "expert" Reactant knowledge, so maybe it would be best to host this code in Reactant?

oschulz: Maybe we could have an undef-ctor like

```julia
ConcreteRArray(
    ::UndefInitializer, ::Type{T}, dims::Dims;
    device::Reactant.XLA.AbstractDevice = Reactant.XLA.default_device(),
    client::Reactant.XLA.AbstractClient = Reactant.XLA.client(device)),
    sharding::Sharding.AbstractSharding = Sharding.NoSharding()
)
```

or so as a starting point?  ~~I'm not sure the default values for device and client are correct like that. I'm also not sure how to implement it correctly - looking at how `similar` is implemented in all of it's stages (calling `similar` again for the data and their buffer and so on) it doesn't seem trivial to do correctly, at least for a Reactant non-expert. ;-)~~

oschulz: I've added a draft of an undef-ctor (Copilot helped ...) - is this implemented correctly?

src/Types.jl
```diff
@@ -453,12 +453,103 @@ Base.@deprecate_binding ConcreteRNG ReactantRNG
...
+
+
+function ConcretePJRTArray(
+    ::UndefInitializer, ::Type{T}, shape::Dims;
+    client::Union{Nothing,XLA.AbstractClient} = nothing,
+    idx::Union{Int,Nothing} = nothing,
+    device::Union{Nothing,XLA.AbstractDevice} = nothing,
+    sharding::Sharding.AbstractSharding = Sharding.NoSharding(),
+) where {T}
+    theclient, thedevice = _select_client_and_device(client, idx, device, sharding)
```
wsmoses, line 479: we should just delete the similar(Concrete...) definition and have that fwd to here

src/Types.jl
```diff
@@ -453,12 +453,103 @@ Base.@deprecate_binding ConcreteRNG ReactantRNG
...
+function ConcreteIFRTArray(
+    ::UndefInitializer, ::Type{T}, shape::Dims;
+    client::Union{Nothing,XLA.AbstractClient} = nothing,
+    idx::Union{Int,Nothing} = nothing,
+    device::Union{Nothing,XLA.AbstractDevice} = nothing,
+    sharding::Sharding.AbstractSharding = Sharding.NoSharding(),
+) where {T}
+    theclient, thedevice = _select_client_and_device(client, idx, device, sharding)
+    # ToDo: How to avoid allocating dummy array on host?
+    dummy_array = Array{T}(undef, shape)
```
wsmoses, line 498: yeah we dont yet have an api call we can use for that I think [cc @avik-pal ]

src/Types.jl
oschulz: Done (I think, take a look ...).

src/Types.jl
oschulz: Should I open a seperate issue for that, for future API extensions?

oschulz:
@wsmoses Ok, tests pass now (approval-dependent ones haven't run yet). I think it's good to go from my side.

This PR should be completely non-breaking.

Adding client/device/sharding-into to the KernelAbstractions `ReactantBackend` I'll try in a separate PR, it may not be straightforward.

src/Types.jl
```diff
@@ -453,12 +410,106 @@ Base.@deprecate_binding ConcreteRNG ReactantRNG
 Base.@deprecate_binding TracedRNG ReactantRNG
 
 ## Aliases based on the set preferences
-if XLA.REACTANT_XLA_RUNTIME == "PJRT"
+@static if XLA.REACTANT_XLA_RUNTIME == "PJRT"
+    """
+        const ConcreteRArray = ConcretePJRTArray
+
+        ConcreteRArray(
```
wsmoses: why not add these docstrings on ConcretePJRTRArray / ConcreteIFRTArray ?

src/Types.jl
wsmoses: we still need to delete https://github.com/EnzymeAD/Reactant.jl/blob/e4e51e464107dac5012f4e0448c279ee9a2549d3/src/ConcreteRArray.jl#L371 and have that call this [to avoid duplicate implementation

src/Types.jl
oschulz: If we attach the docstring to ConcretePJRTRArray / ConcreteIFRTArray, it won't show up when users do `?ConcreteRArray`. User code should almost always use ConcreteRArray instead of ConcretePJRTRArray / ConcreteIFRTArray, correct?

wsmoses: can we add a separate docstring for ConcreteRArray then which just says see ConcretePJRTArray?

oschulz: I've changed it around a bit, avoiding the duplicate docstring and adding a short note on implementation. Is this better now?

src/Types.jl
oschulz: Done.

Question: This is inline - should I make the undef-ctors inline as well?

oschulz:
Let's see if I fixed docgen now ...

oschulz:
I added a docstring for `ConcreteRNumber` too now, since it's exported and mentioned in the docs.

src/Types.jl
oschulz: Created #1578.

PR #1562, opened 2025-08-17
windows ci

wsmoses:


gbaraldi:
i'm not super aware of what this is doing. But preserving random pointers will not do anything. Preserves are for the julia object that gets turned into a pointer.

PR #1571, opened 2025-08-20
Fix typedef and undef ctors for ConcreteRArray and ConcreteRNumber

oschulz:
I don't know what I was thinking in #1558, but Julia convention is of course
`ArrayType{T}(undef, dims)` and `ArrayType{T}(undef, dims...)`, but *not* `ArrayType(undef, T, dims)`.

Also, the typedef for `ConcreteRArray` was missing `T` and `N` parameters and the typedefs for `ConcreteRNumber` was missing `T`,

This PR fixes that (with deprecation for `ConcreteRArray(undef, T, dims; kwargs...)`) and adds tests.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -440,10 +440,6 @@ end
     ::UndefInitializer, shape::Integer...; kwargs...
 ) where {T<:Number} = ConcreteRArray{T}(undef, Dims(shape); kwargs...)
 
-@deprecate ConcreteRArray(
-    ::UndefInitializer, ::Type{T}, shape::Dims; kwargs...
-) where {T<:Number} ConcreteRArray{T}(undef, shape; kwargs...)
-
 """
     ConcreteRNumber(
         x::Number;
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -478,7 +478,7 @@ function ConcretePJRTArray{T}(
     idx::Union{Int,Nothing}=nothing,
     device::Union{Nothing,XLA.AbstractDevice}=nothing,
     sharding::Sharding.AbstractSharding=Sharding.NoSharding(),
-) where {T<:Number}
+) where {T}
     theclient, thedevice = _select_client_and_device(client, idx, device, sharding)
     sharded_data, shardinfo = sharding(theclient, thedevice, T, shape)
     N = length(shape)
@@ -493,7 +493,7 @@ function ConcreteIFRTArray{T}(
     idx::Union{Int,Nothing}=nothing,
     device::Union{Nothing,XLA.AbstractDevice}=nothing,
     sharding::Sharding.AbstractSharding=Sharding.NoSharding(),
-) where {T<:Number}
+) where {T}
     theclient, thedevice = _select_client_and_device(client, idx, device, sharding)
     N = length(shape)
     # ToDo: How to avoid allocating dummy array on host?
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -452,18 +452,16 @@ Depending on the Reactant `xla_runtime` preference setting, `ConcreteRArray`
 is an alias for `ConcretePJRTNumber` or `ConcreteIFRTNumber`. User code should
 use `ConcreteRNumber`.
 """
-const ConcreteRNumber = @static if XLA.REACTANT_XLA_RUNTIME == "PJRT"
-    ConcretePJRTNumber
+const ConcreteRNumber{T} = @static if XLA.REACTANT_XLA_RUNTIME == "PJRT"
+    ConcretePJRTNumber{T}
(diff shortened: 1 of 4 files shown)


wsmoses (approved):
I wouldn't bother deprecating and just deleting the bad one here.

src/Types.jl
```diff
@@ -452,46 +462,42 @@ Depending on the Reactant `xla_runtime` preference setting, `ConcreteRArray`
...
+function ConcreteIFRTArray{T}(
     ::UndefInitializer,
-    ::Type{T},
     shape::Dims;
     client::Union{Nothing,XLA.AbstractClient}=nothing,
     idx::Union{Int,Nothing}=nothing,
     device::Union{Nothing,XLA.AbstractDevice}=nothing,
     sharding::Sharding.AbstractSharding=Sharding.NoSharding(),
-) where {T}
+) where {T<:Number}
```
wsmoses: This should be primitive not number

oschulz: So just `T` then?

oschulz: I took the `<:Number` out.

oschulz:
> I wouldn't bother deprecating and just deleting the bad one here.

Ok, it's gone. :-)

oschulz:
> I wouldn't bother deprecating and just deleting the bad one here.

Ok, it's gone. :-)

src/Types.jl
```diff
@@ -428,12 +426,20 @@ Depending on the Reactant `xla_runtime` preference setting, `ConcreteRArray`
 is an alias for `ConcretePJRTArray` or `ConcreteIFRTArray`. User code should
 use `ConcreteRArray`.
 """
-const ConcreteRArray = @static if XLA.REACTANT_XLA_RUNTIME == "PJRT"
-    ConcretePJRTArray
+const ConcreteRArray{T,N} = @static if XLA.REACTANT_XLA_RUNTIME == "PJRT"
```
wsmoses: I don't think you need this specialization?

src/Types.jl
```diff
@@ -428,12 +426,20 @@ Depending on the Reactant `xla_runtime` preference setting, `ConcreteRArray`
...
-    ConcreteIFRTArray
+    ConcreteIFRTArray{T,N}
 end
 
+@inline ConcreteRArray{T}(
+    ::UndefInitializer, shape::Integer...; kwargs...
+) where {T<:Number} = ConcreteRArray{T}(undef, Dims(shape); kwargs...)
+
+@deprecate ConcreteRArray(
+    ::UndefInitializer, ::Type{T}, shape::Dims; kwargs...
```
wsmoses: remove the deprecate

src/Types.jl
```diff
@@ -452,46 +458,42 @@ Depending on the Reactant `xla_runtime` preference setting, `ConcreteRArray`
 is an alias for `ConcretePJRTNumber` or `ConcreteIFRTNumber`. User code should
 use `ConcreteRNumber`.
 """
-const ConcreteRNumber = @static if XLA.REACTANT_XLA_RUNTIME == "PJRT"
-    ConcretePJRTNumber
+const ConcreteRNumber{T} = @static if XLA.REACTANT_XLA_RUNTIME == "PJRT"
```
wsmoses: same comment removing {T}

src/Types.jl
oschulz: I think it would be quite natural. It's hard to predict what users may need to dispatch on, in Reactant-extensions of their packages, for example - but element type and dimensionality are very common means of specialization, after all.

src/Types.jl
oschulz: Argh, I did remove the wrong deprecation - fixed now, sorry.

src/Types.jl
oschulz: Couldn't it be useful to be able to dispatch on the actual number type, low-level Reactant-specific code (package extensions and so on)?

src/Types.jl
giordano: I think the point of the comment is that you don't need to add `{T, N}`
```julia-repl
julia> const MyAlias = Array
Array

julia> MyAlias{Float64, 3}
Array{Float64, 3}
```

oschulz: Oh, of course! Sorry, I usually have typedefs with Unions, I completely forgot that a single typedef "inherits" all type parameters. Ok gone.

src/Types.jl
oschulz: `{T}` is gone now.

oschulz:
Are the test failures related to this PR?

giordano:
No (and one is simply the coverage step failing to start at all)

oschulz:
> No (and one is simply the coverage step failing to start at all)

Thanks. I'm good with this from my side then.

PR #1579, opened 2025-08-22
[build_local] Also symlink `libdevice` to appropriate path

giordano:
I'm testing something along these lines in https://github.com/EnzymeAD/Enzyme-JAX/pull/1335.


avik-pal:
should we switch our local libReactant runners to only use the gcp runners?

wsmoses:
Yes

giordano:
Confirmed to work in https://github.com/EnzymeAD/Enzyme-JAX/pull/1338.  Changes to CI setup were taken out in #1584

wsmoses:
We also might need to symlink fatbin as well

PR #1587, opened 2025-08-24
A comic number of backslashes to support rewriting patch files define…

wsmoses:
…d in enzyme-jax

will let https://github.com/EnzymeAD/Enzyme-JAX/pull/1337 and future patches pass

### DIFF
--- a/deps/ReactantExtra/WORKSPACE
+++ b/deps/ReactantExtra/WORKSPACE
@@ -20,6 +20,11 @@ http_archive(
     sha256 = ENZYMEXLA_SHA256,
     strip_prefix = "Enzyme-JAX-" + ENZYMEXLA_COMMIT,
     urls = ["https://github.com/EnzymeAD/Enzyme-JAX/archive/{commit}.tar.gz".format(commit = ENZYMEXLA_COMMIT)],
+    patch_cmds = [
+    """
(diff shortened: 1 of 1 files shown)


deps/ReactantExtra/WORKSPACE
```diff
@@ -20,6 +20,11 @@ http_archive(
     sha256 = ENZYMEXLA_SHA256,
     strip_prefix = "Enzyme-JAX-" + ENZYMEXLA_COMMIT,
     urls = ["https://github.com/EnzymeAD/Enzyme-JAX/archive/{commit}.tar.gz".format(commit = ENZYMEXLA_COMMIT)],
+    patch_cmds = [
+    """
+sed -i.bak0 "s/\\\\\\\\\\\\\\\\\\/\\\\\\\\\\\\\\\\\\/:patches/@enzyme_ad\\\\\\\\\\\\\\\\\\/\\\\\\\\\\\\\\\\\\/:patches/g" workspace.bzl
```
giordano, line 25: Why not using non-backlash separators (like `?`) in the sed command instead?

wsmoses: unfortunately I don't think that would save much here, but if you have a better idea, go for it!

Issue #1589, opened 2025-08-25
Overlayed functions not detected inside another overlayed function


unit-of-inductance:
```julia
using LinearAlgebra, Reactant, Lux, Random

const xdev = reactant_device()

model = Dense(2 => 1)
points = [rand(2),rand(2)] |> xdev
parameters, states = Lux.setup(Random.default_rng(), model) |> xdev

function f(parameters, points)
    return mapreduce(p -> Lux.LuxCore.stateless_apply(model, p, parameters), +, points)
end

function g(parameters, points)
    return reduce(+, map(p -> Lux.LuxCore.stateless_apply(model, p, parameters), points))
end

# works without issues
g_compiled = @compile g(parameters, points)

# raises error
f_compiled = Reactant.allowscalar() do
   @compile f(parameters, points)
end
```
The above code raises the following error (this appears seperate to me to the issue from #1554, which seems to necessitate the `allowscalar()` call here) :
```
1-element ExceptionStack:
MethodError: no method matching Float64(::Reactant.TracedRNumber{Float64})
The type `Float64` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  (::Type{T})(::T) where T<:Number
   @ Core boot.jl:900
  Float64(::Irrational{:SQRT_HALF})
   @ Random irrationals.jl:251
  Float64(::Irrational{:catalan})
   @ Base irrationals.jl:251
  ...
```
<details><summary>Stacktrace</summary>
<p>

```
Stacktrace:
  [1] setindex!(a::Reactant.TracedRArray{Float64, 2}, v::Reactant.TracedRNumber{Float64}, index::CartesianIndex{2})
    @ Reactant.TracedRArrayOverrides ~/.julia/packages/Reactant/ZEral/src/TracedRArray.jl:370
  [2] _modify!
    @ /nix/store/0f8r9l0fn26mjj96j08p92ak8gmbgwas-julia-bin-1.11.6/share/julia/stdlib/v1.11/LinearAlgebra/src/generic.jl:91 [inlined]
  [3] _generic_matmatmul!(C::Reactant.TracedRArray{Float64, 2}, A::Reactant.TracedRArray{Float32, 2}, B::Base.ReshapedArray{Reactant.TracedRNumber{Float64}, 2, Reactant.TracedRArray{Float64, 1}, Tuple{}}, _add::LinearAlgebra.MulAddMul{true, true, Bool, Bool})
[...]

PR #1596, opened 2025-08-28
fix: overlay searchsorted functions

avik-pal:


wsmoses (approved):

avik-pal:
the patch fails because of https://github.com/EnzymeAD/Reactant.jl/issues/1589. calling `sum` (aka mapreduce) from inside an overlayed function

wsmoses:
hm that shouldn't be an issue, as long as within the overlay we call back into call_with_reactant, for the function being mapped

wsmoses:
e.g. we probably want to have that be map(Base.Fix1(call_with_reactant), f), ...) or something on the inside

PR #1607, opened 2025-08-30
fix: add overloads for Base.complex

avik-pal:


src/TracedRNumber.jl
```diff
@@ -136,6 +136,21 @@ end
 Base.Complex(x::TracedRNumber{<:Real}) = @opcall complex(x, zero(x))
 Base.Complex(x::TracedRNumber{<:Complex}) = x
 
+# Base.complex
+Base.complex(::Type{TracedRNumber{T}}) where {T} = TracedRNumber{complex(T)}
+Base.complex(x::TracedRNumber{<:Real}) = complex(x, zero(x))
+function Base.complex(x::TracedRNumber{<:Real}, y::TracedRNumber{<:Real})
```
wsmoses, line 142: we probably also need the case where only one of x/y are traced

PR #1613, opened 2025-09-01
[CI] Run job updating MLIR bindings on Julia v1.11

giordano:


deps/ReactantExtra/Project.toml
```diff
@@ -4,3 +4,6 @@ BinaryBuilderBase = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 
 [compat]
 Clang = "0.18"
+
+[sources]
+Clang = {repo = "https://github.com/wsmoses/Clang.jl", rev = "patch-1"}
```
giordano: This should be reverted after https://github.com/JuliaInterop/Clang.jl/pull/552 is merged and makes it into a new release.

PR #1615, opened 2025-09-02
feat: use scoped settings for global configurable options

### LINKED ISSUE #1608: Use `ScopedSettings.jl` to allow global configuration of flags
https://github.com/oschulz/ScopedSettings.jl

We can switch the current ScopedValue to this, and more importantly make the myriad number of global Refs into scoped settings so that we can globally modify them or use scopes

avik-pal:
- [x] replace `ScopedValue` and `Ref` with `ScopedSetting`
- [ ] needs https://github.com/oschulz/ScopedSettings.jl/pull/3
- [ ] documentation for the options
- [ ] allow specifying the values via `with_config`

fixes #1608

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -27,9 +27,42 @@ import Reactant: OptimizeCommunicationOptions, ShardyPropagationOptions, Compile
 
 import ..ReactantCore: correct_maybe_bcast_call
 
-const DEBUG_PRINT_CODEGEN = Ref(false)
-const DEBUG_DISABLE_RESHARDING = Ref(false)
-const DEBUG_ALIASED_BUFFER_ASSIGNMENT_ERROR = Ref(false)
+using ScopedSettings: ScopedSetting, GetPreference
+
+const DEBUG_PRINT_CODEGEN = ScopedSetting(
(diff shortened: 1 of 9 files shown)


src/Compiler.jl
```diff
@@ -27,9 +27,42 @@ import Reactant: OptimizeCommunicationOptions, ShardyPropagationOptions, Compile
...
+const DEBUG_PRINT_CODEGEN = ScopedSetting(
+    GetPreference(Reactant, "debug_print_codegen", false)
+)
+const DEBUG_DISABLE_RESHARDING = ScopedSetting(
+    GetPreference(Reactant, "debug_disable_resharding", false)
+)
+const DEBUG_ALIASED_BUFFER_ASSIGNMENT_ERROR = ScopedSetting(
+    GetPreference(Reactant, "debug_aliased_buffer_assignment_error", false)
+)
+const DEBUG_KERNEL = ScopedSetting(GetPreference(Reactant, "debug_kernel", false))
```
wsmoses, line 41: these compile time attributes really don't make sense to be scoped, since they're set once per the entire compilation

avik-pal: I just searched through and replaced them :sweat:. will go through all of them once

Issue #1616, opened 2025-09-02
Could not find unique name for Reactant.TracedUtils.TypeCast{Float64}()_broadcast_scalar when compiling function with lots of matrix multiplications


unit-of-inductance:
Consider the following MWE
```
using Reactant

n_points = 10000
dim = 10

points = Reactant.to_rarray(eachcol(rand(dim,n_points)))
matrix = Reactant.to_rarray(rand(1, dim))

function f(points)
  return sum(matrix * point for point in points)
end

f_compiled = @compile f(points)
```
This raises
```
Could not find unique name for Reactant.TracedUtils.TypeCast{Float64}()_broadcast_scalar
```
<details><summary>Stracktrace</summary>
<p>

```
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] __lookup_unique_name_in_module(mod::Reactant.MLIR.IR.Module, name::String)
    @ Reactant.TracedUtils ~/.julia/packages/Reactant/EsmaI/src/TracedUtils.jl:998
  [3] (::Reactant.TracedUtils.var"#13#20"{…})()
    @ Reactant.TracedUtils ~/.julia/packages/Reactant/EsmaI/src/TracedUtils.jl:842
  [4] block!(f::Reactant.TracedUtils.var"#13#20"{…}, blk::Reactant.MLIR.IR.Block)
    @ Reactant.MLIR.IR ~/.julia/packages/Reactant/EsmaI/src/mlir/IR/Block.jl:214
[...]

PR #1617, opened 2025-09-02
Refactor unique name lookup loop

jumerckx:
should hopefully fix https://github.com/EnzymeAD/Reactant.jl/issues/1616

### DIFF
--- a/src/TracedUtils.jl
+++ b/src/TracedUtils.jl
@@ -990,12 +990,15 @@ end
 function __lookup_unique_name_in_module(mod, name)
     new_name = name
     tab = MLIR.IR.SymbolTable(MLIR.IR.Operation(mod))
-    for i in 0:10000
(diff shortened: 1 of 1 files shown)


src/TracedUtils.jl
```diff
@@ -990,12 +990,15 @@ end
 function __lookup_unique_name_in_module(mod, name)
     new_name = name
     tab = MLIR.IR.SymbolTable(MLIR.IR.Operation(mod))
-    for i in 0:10000
+    i = 0
+    while true
```
wsmoses, line 994: erm I'm not sure we should make this change.

clearly this is already a catastrophic situation, and early terminating feels better than to infinte loop

jumerckx: ok, closing in favor of something better then.

PR #1622, opened 2025-09-03
Update Clang compatibility version to 0.19

giordano:
The job is failing because of the compat bound, but someone should really tests the script still works (I'm really not familiar with the binding generation stuff), since it's a breaking change _some_ breakage should be expected


wsmoses:
we locally tested this with the given branch earlier, so im fine to merge

Issue #1555, opened 2025-08-14
Compile errors when using AcceleratedKernels.mapreduce on a StructVector of Tuples


oschulz:
~~When Reactant-compiling an `AcceleratedKernels.mapreduce` over a `StructArray`, Reactant seems to just store the results, and doesn't recompute for new data.~~

~~Reactant just replays the result from the first run (computed during compilation somehow?). Is this a bug or am I doing something wrong here?~~ (Edit: yes I was).

wsmoses:
@oschulz the linked code has a syntax error? neutral=?

wsmoses:
also your code is also clearly wrong:
```
julia> foo(AB) = AcceleratedKernels.mapreduce(splat(+), +, cuAB; init = zero(Float32), neutral = 
```

you bake in cuAB as a global into it there, instead of using the argument AB.

Reactant partially evaluates in any non-reactant data. aka

```
using Reactant
function add(a, b)
   a + b
end

x = ConcreteRNumber(3)
y = ConcreteRNumber(4)

addxy = @compile add(x, y)

addxy(x, y)
# 7, as we compiled a function with both x and y as varying, nothing partially evaluated into addxy

addxy(x, ConcreteRNumber(10))
# 13, as we compiled a function with both x and y as varying, nothing partially evaluated into addxy

addx4 = @compile add(x, 4)

addx4(x, 4)
# 7, as we compiled a function with both x as varying but y as something to be baked in, using the same value 4 for y

addx4(x, 10)
# 7, as we compiled a function with both x as varying but y as something to be baked in.


# for more we can look at the compiled IR
@code_hlo add(x, y)
module @reactant_add attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<i64>, %arg1: tensor<i64>) -> tensor<i64> {
    %0 = stablehlo.add %arg0, %arg1 : tensor<i64>
    return %0 : tensor<i64>
  }
}

@code_hlo add(x, 4)
module @reactant_add attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<i64>) -> tensor<i64> {
    %c = stablehlo.constant dense<4> : tensor<i64>
    %0 = stablehlo.add %arg0, %c : tensor<i64>
    return %0 : tensor<i64>
  }
}
```

[...]

PR #1624, opened 2025-09-03
Add partial evaluation tutorial

oschulz:
See discussion in #1555.

### DIFF
--- a/docs/src/tutorials/partial-evaluation.md
+++ b/docs/src/tutorials/partial-evaluation.md
@@ -12,15 +12,20 @@ on this.
 For example, the function
 
 
-```@example partial_evaluation_tutorial
+```jldoctest partial_evaluation_tutorial
+using Reactant
 function add(a, b)
    a + b
-end
+end;
+
+# output
+
+add (generic function with 1 method)
 ```
 
 when compiled with two `ConcreteRNumber` arguments
 
-```@example partial_evaluation_tutorial
+```jldoctest partial_evaluation_tutorial; filter = r"I000.*" => s""
 using Reactant
 
 x = ConcreteRNumber(3)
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/index.md
```diff
@@ -6,5 +6,6 @@
 - [Control flow](@ref control-flow).
 - [Sharding](@ref sharding).
 - [Persistent Compilation Cache](@ref persistent_compile_cache).
+- [Partial evaluation](@ref partial-evaluation).
```
wsmoses: This is kind of important I might put this first, then the control flow one

docs/src/tutorials/partial-evaluation.md
```diff
@@ -0,0 +1,70 @@
+# [Partial Evaluation](@id partial-evaluation)
+
+When compiling functions with Reactant, the function arguments (and possible
+closure fields) may contain non-Reactant values, so numbers and arrays that
```
wsmoses: Instead of so, can you use i.e.

docs/src/tutorials/partial-evaluation.md
```diff
@@ -0,0 +1,70 @@
...
+
+```@example partial_evaluation_tutorial
+using Reactant
+
+x = ConcreteRNumber(3)
+y = ConcreteRNumber(4)
+
+addxy = @compile add(x, y)
+
+addxy(x, y)
```
wsmoses, line 36: Can we add doc tests here to confirm the result (and also show the result)

docs/src/tutorials/index.md
oschulz: Done

docs/src/tutorials/partial-evaluation.md
oschulz: Haha, I had "i.e." and then changed it to "so". :-) Changed back now.

docs/src/tutorials/partial-evaluation.md
oschulz: For all of them, or just for this one?

Also, should I convert the result to a `Float64` - otherwise the full type of the RNumber will become part of the doctest.

wsmoses: I'd say all of them, and honestly that full type is fine here

docs/src/tutorials/index.md
```diff
@@ -3,6 +3,7 @@
 - [Profiling](@ref profiling).
 - [Multi-Host Environments](@ref distributed).
 - [Local build of ReactantExtra](@ref local-build).
+- [Partial evaluation](@ref partial-evaluation).
```
wsmoses, line 3: oh I'd even go further and move the partial eval to be the top of the list, then control flow, then sharding, then profiling, the rest of the list

oschulz: Ok, done.

docs/src/tutorials/partial-evaluation.md
oschulz: Ok, turned it all into `jldoctest`s. Some of the outputs look like

```
I0000 00:00:1756978474.009076  615672 dump.cc:108] [MaybeDumpCompileInputs] Dumping PJRT inputs for module: reactant_add
I0000 00:00:1756978474.009107  615672 dump.cc:120]   Dump path not set via xla_dump_to, skipping dump.
ConcretePJRTNumber{Int64, 1, Reactant.Sharding.ShardInfo{Reactant.Sharding.NoSharding, Nothing}}(7)
```

I'm currently using the `jldoctest` output filter option to ignore everything after `I0000` - is it possible to suppress this compiler info/output?

giordano: Avik mentioned this should be fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/12002 (but that will take a few hours to come in)

oschulz: Doc build on CI fails due to an additional output

```
AssertionError("ReactantExtra was not built with GPU support")
```

That's more difficult to filter - the `jldoctest` filter opion seems to work line-by-line, so I don't know if lines can be removed completely.

wsmoses: We should potentially add an early check Julia side to only try setting up GPU backends if compiled with a jll for a GPU?

that would avoid the error

oschulz: In Reactant in general or for the tutorial? If the latter, how do I do that?

wsmoses: https://github.com/EnzymeAD/Reactant.jl/pull/1647 may do it

oschulz: Thanks, I'll rebase this PR as soon as #1647 is merged then.

PR #1626, opened 2025-09-04
Create ROCm client via plugin

giordano:
This is ***very*** preliminary at the moment, but I managed to get access to the devices, so it looks promising:
```julia-repl
julia> Reactant.devices()
I0000 00:00:1756985457.845658 3031425 pjrt_api.cc:118] GetPjrtApi was found for rocm at /cosma/home/do006/dc-gior1/tmp/pjrt-plugin/jax_plugins/xla_rocm60/xla_rocm_plugin.so
I0000 00:00:1756985457.863487 3031425 pjrt_api.cc:96] PJRT_Api is set for device type rocm
I0000 00:00:1756985457.863510 3031425 pjrt_api.cc:167] The PJRT plugin has PJRT API version 0.55. The framework PJRT API version is 0.75.
I0000 00:00:1756985463.767717 3031425 pjrt_c_api_client.cc:133] PjRtCApiClient created.
4-element Vector{Reactant.XLA.PJRT.Device}:
 Reactant.XLA.PJRT.Device(Ptr{Nothing} @0x000000001bc36820, "ROCM:0 AMD Instinct MI300A")
 Reactant.XLA.PJRT.Device(Ptr{Nothing} @0x000000001bc36c70, "ROCM:1 AMD Instinct MI300A")
 Reactant.XLA.PJRT.Device(Ptr{Nothing} @0x000000001bc36fb0, "ROCM:2 AMD Instinct MI300A")
 Reactant.XLA.PJRT.Device(Ptr{Nothing} @0x000000001bc372f0, "ROCM:3 AMD Instinct MI300A")
```

For the record, for some reason the env var `ROCM_PATH` wasn't set in my environment and I had to manually do it to help XLA find libdevice, otherwise I was getting errors like
```
INTERNAL: bitcode module not found at ./opencl.bc
```
As usual, setting the environment variables
```sh
TF_CPP_MIN_VLOG_LEVEL=0
TF_CPP_MAX_VLOG_LEVEL=3
```
was very useful for debugging.

### DIFF
--- a/src/accelerators/ROCm.jl
+++ b/src/accelerators/ROCm.jl
@@ -1,0 +1,62 @@
+module ROCm
+
+using Reactant: Reactant
+using Scratch: @get_scratch!
+using Downloads
+
+const rocm_pjrt_plugin_dir = Ref{Union{Nothing,String}}(nothing)
+
+function __init__()
+    @static if Sys.islinux()
+        Reactant.precompiling() || setup_rocm_pjrt_plugin!()
+    end
+end
+
+has_rocm() = true
+
+function setup_rocm_pjrt_plugin!()
+    path_from_env = get(ENV, "ROCM_LIBRARY_PATH", nothing)
+    if path_from_env !== nothing && ispath(path_from_env)
+        rocm_pjrt_plugin_dir[] = path_from_env
+    else
+        rocm_pjrt_plugin_dir[] = @get_scratch!("pjrt_rocm_plugin")
+    end
+    # download_rocm_pjrt_plugin_if_needed(rocm_pjrt_plugin_dir[])
+    return nothing
+end
+
+get_rocm_pjrt_plugin_dir() = rocm_pjrt_plugin_dir[]
+
+function get_rocm_pjrt_plugin_path()
+    return joinpath(get_rocm_pjrt_plugin_dir(), "xla_rocm_plugin.so")
+end
+
+# function download_rocm_pjrt_plugin_if_needed(path=nothing)
+#     path === nothing && (path = get_rocm_pjrt_plugin_dir())
+#     @assert path !== nothing "rocm_pjrt_plugin_dir is not set!"
+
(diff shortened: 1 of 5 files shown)


src/accelerators/ROCm.jl
```diff
@@ -0,0 +1,62 @@
...
+#         )
+#         run(`unzip -qq $(zip_file_path) -d $(tmp_dir)`)
+#         mv(
+#             joinpath(tmp_dir, "jax_plugins", "rocm_plugin", "pjrt_plugin_rocm_14.dylib"),
+#             rocm_pjrt_plugin_path,
+#         )
+#         rm(tmp_dir; recursive=true)
+#         rm(zip_file_path; recursive=true)
+#     end
+# end
```
giordano, line 60: We need to download the wheel from https://rocm.docs.amd.com/projects/radeon/en/latest/docs/install/native_linux/install-jax.html, the problem being that there are different builds for different distributions, we'd need first to identify them.  At the moment I'm working with a local copy and I set the `ROCM_LIBRARY_PATH` env var (is there a standard name?)

src/accelerators/ROCm.jl
```diff
@@ -0,0 +1,62 @@
...
+
+const rocm_pjrt_plugin_dir = Ref{Union{Nothing,String}}(nothing)
+
+function __init__()
+    @static if Sys.islinux()
+        Reactant.precompiling() || setup_rocm_pjrt_plugin!()
+    end
+end
+
+has_rocm() = true
```
giordano, line 15: Of course this needs to be made more realistic than this 😄

src/accelerators/ROCm.jl
owainkenwayucl: I believe it's doable to build JAX from source for ROCm.

I actually need to try this for another thing so I'll update you on how easy/painful it is.

giordano:
Closing this because ROCm is supported natively in libreactant now, but this PR can still be a good reference in the future for other plugins (I'd like to do this for Tenstorrent, if I only I manage to install all their dependencies....)

PR #1632, opened 2025-09-06
ci: update the continuous benchmarking ci to use the new machines

avik-pal:


avik-pal:
@wsmoses / @giordano would you happen to know why the cuda runners are failing to load reactant https://github.com/EnzymeAD/Reactant.jl/actions/runs/17509546097/job/49739166333?pr=1632#step:6:423

giordano:
If you ask me, I never understood why Reactant_jll depends on CUDA_Driver_jll in the first place.  In https://github.com/EnzymeAD/ReactantBuilder/pull/1 I also noted that `CUDA_Driver_jll` is loaded unconditionally on all non-macOS platforms, also when there isn't a GPU: https://github.com/JuliaPackaging/Yggdrasil/blob/500450fbc8748ee4c29c0beda3d21246c65b5bb9/R/Reactant/build_tarballs.jl#L501-L503. The fact that [libraries in CUDA_Driver_jll aren't dlopened automatically](https://github.com/JuliaPackaging/Yggdrasil/blob/500450fbc8748ee4c29c0beda3d21246c65b5bb9/C/CUDA/CUDA_Driver/build_tarballs.jl#L55-L58) and no one is dlopening them convinces me that that dependency is completely useless here.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,103 @@
...
+      image: ${{ contains(matrix.os, 'linux') && 'ghcr.io/enzymead/reactant-docker-images@sha256:7004a6ebbdd77bd047900b2bffc542e8576864056dc27a9c94d30666d6f7ea01' || '' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - linux-x86-n2-32
+          - linux-x86-ct6e-180-4tpu
+        # XXX: temporarily disabled since the JLLs are failing
+        #      https://github.com/EnzymeAD/Reactant.jl/actions/runs/17509546097/job/49739166333?pr=1632#step:6:423
+        #   - linux-x86-a2-48-a100-4gpu
```
giordano: It makes a me bit uncomfortable to overuse these runners, when we have so few available at the moment, with this change every single PR or push to `main` in this repo will clog both queues.

avik-pal: I can make this opt-in for PRs with a label (specifically we want this for PRs that update the jll version). For main, I can do a nightly job.

giordano: That sounds better 👍

wsmoses:
> If you ask me, I never understood why Reactant_jll depends on CUDA_Driver_jll in the first place.  In https://github.com/EnzymeAD/ReactantBuilder/pull/1 I also noted that `CUDA_Driver_jll` is loaded unconditionally on all non-macOS platforms, also when there isn't a GPU: https://github.com/JuliaPackaging/Yggdrasil/blob/500450fbc8748ee4c29c0beda3d21246c65b5bb9/R/Reactant/build_tarballs.jl#L501-L503. The fact that [libraries in CUDA_Driver_jll aren't dlopened automatically](https://github.com/JuliaPackaging/Yggdrasil/blob/500450fbc8748ee4c29c0beda3d21246c65b5bb9/C/CUDA/CUDA_Driver/build_tarballs.jl#L55-L58) and no one is dlopening them convinces me that that dependency is completely useless here.

Yeah it's sole purpose atm is for figuring out cuda version to select the relevant jll. If we can do that in another way we should get rid of it!

giordano:
> figuring out cuda version to select the relevant jll

Where does that happen? 🤔

.github/workflows/benchmark.yml
avik-pal: done

avik-pal:
For some reason benchmarking the backward pass for ViT on TPUs seems to take forever

wsmoses:
> > figuring out cuda version to select the relevant jll
> 
> 
> 
> Where does that happen? 🤔

https://github.com/JuliaPackaging/Yggdrasil/blob/52447ed000894dbd026c7b035a3c24d750590862/R/Reactant/platform_augmentation.jl#L45

giordano:
> > > figuring out cuda version to select the relevant jll
> > 
> > 
> > Where does that happen? 🤔
> 
> https://github.com/JuliaPackaging/Yggdrasil/blob/52447ed000894dbd026c7b035a3c24d750590862/R/Reactant/platform_augmentation.jl#L45

That file doesn't load CUDA_Driver_jll at all, it's using the local driver.

wsmoses:
Oh cool we should nix the dep then

benchmark/Project.toml
```diff
@@ -1,33 +1,12 @@
...
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
-Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
-
-[compat]
-BenchmarkTools = "1.5"
-Boltz = "1"
-Enzyme = "0.13"
-Lux = "1.1"
-Random = "1.10"
-julia = "1.10"
```
giordano: Keep the relevant compat bounds?  That'd avoid immediate issues when upstream packages go through a breaking change.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          version: "1"
+      - uses: julia-actions/cache@v2
+      - name: "Run Benchmarks"
+        run: |
+          julia --color=yes --project=benchmark -e '@info "Instantiating project"
+            using Pkg
+            Pkg.instantiate()'
+
+          julia --color=yes --project=benchmark benchmark/runbenchmarks.jl
```
giordano: Since you mentioned a job taking forever, can you add a reasonable timeout, to prevent jobs going haywire?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,103 @@
...
+          path: "benchmark/results/*"
+          retention-days: 90
+          overwrite: false
+
+  benchmark-aggregate:
+    if: ${{ (github.ref == 'refs/heads/main' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')) || (github.event_name == 'pull_request' && contains(join(github.event.pull_request.labels.*.name, ','), 'run benchmarks')) }}
+    needs: benchmark
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
giordano: On this one which uses `ubuntu-latest` you can use v5 (the self-hosted agents are still on an older version of the runner software)
```suggestion
      - uses: actions/checkout@v5
```

avik-pal:
need to verify this but I think a lot of time is from the TPU compilation. Need to figure out a way to get `@setup` to only compile once.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,116 @@
...
+        with:
+          version: "1"
+      - uses: julia-actions/cache@v2
+      - name: "Run Benchmarks"
+        run: |
+          julia --color=yes --project=benchmark -e '@info "Instantiating project"
+            using Pkg
+            Pkg.instantiate()'
+
+          julia --color=yes --project=benchmark benchmark/runbenchmarks.jl
```
giordano:
```suggestion
      - name: "Instantiate benchmarks environment"
        shell: julia --color=yes --project=benchmark {0}
        run: |
          using Pkg
          Pkg.instantiate()
      - name: "Run Benchmarks"
        run: |
          julia --color=yes --project=benchmark benchmark/runbenchmarks.jl
```

wsmoses (approved):

src/Compiler.jl
```diff
@@ -1135,7 +1135,8 @@ function optimization_passes(
                 "transpose_batch_norm_grad",
                 "transpose_if",
                 "transpose_fft",
-                "transpose_reshape",
+                # XXX: https://github.com/EnzymeAD/Enzyme-JAX/issues/1374
+                # "transpose_reshape",
```
wsmoses: can this be re-enabled as the linked issue is closed?

avik-pal: Still waiting for jll build to go through

wsmoses: ah lets do https://github.com/EnzymeAD/Enzyme-JAX/pull/1381#issuecomment-3282878013 merge that, then start jll'ing

avik-pal:
```suggestion
                "transpose_reshape",
```

PR #1658, opened 2025-09-11
feat: overlay Zygote.gradient and use Enzyme instead

avik-pal:
I know @ChrisRackauckas will like this :sweat_smile:

### DIFF
--- a/ext/ReactantZygoteExt.jl
+++ b/ext/ReactantZygoteExt.jl
@@ -18,10 +18,7 @@ using Enzyme: Enzyme, Reverse, Active, Const, Duplicated
                not use `Zygote.gradient` and instead use `Enzyme.gradient` inside \
                `Reactant.@compile`. If this behavior is undesirable, set the \
                `overlay_zygote_calls` scoped value via `Reactant.with_config` to `false`."
-        dargs = map(Enzyme.make_zero, args)
-        duplicated = map(Duplicated, args, dargs)
-        Reactant.overload_autodiff(Reverse, Const(f), Active, duplicated...)
-        return dargs
+        return Enzyme.gradient(Reverse, Const(f), args...)
     else
         return Base.inferencebarrier(Zygote.gradient)(CallWithReactant(f), args...)
     end
--- a/ext/ReactantZygoteExt.jl
+++ b/ext/ReactantZygoteExt.jl
@@ -13,6 +13,9 @@ using Enzyme: Enzyme, Reverse, Active, Const, Duplicated
 @reactant_overlay function Zygote.gradient(f::F, args...) where {F}
     # TODO: check `f` as well once #1642 is merged
     if use_overlayed_version(args)
+        @warn "Reactant doesn't support using Zygote for computing gradients. Replacing \
+               `Zygote.gradient` with `Enzyme.autodiff` call. Please update your code to \
+               not use `Zygote.gradient` inside `Reactant.@compile`." maxlog = 1
         dargs = map(Enzyme.make_zero, args)
         duplicated = map(Duplicated, args, dargs)
         Reactant.overload_autodiff(Reverse, Const(f), Active, duplicated...)
--- a/ext/ReactantZygoteExt.jl
+++ b/ext/ReactantZygoteExt.jl
@@ -17,7 +17,9 @@ using Enzyme: Enzyme, Reverse, Active, Const, Duplicated
                `Zygote.gradient` with `Enzyme.autodiff` call. Please update your code to \
                not use `Zygote.gradient` and instead use `Enzyme.gradient` inside \
                `Reactant.@compile`. If this behavior is undesirable, set the \
-               `overlay_zygote_calls` scoped value via `Reactant.with_config` to `false`."
+               `overlay_zygote_calls` scoped value via `Reactant.with_config` to \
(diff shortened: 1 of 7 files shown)


ext/ReactantZygoteExt.jl
```diff
@@ -0,0 +1,25 @@
...
+using Enzyme: Enzyme, Reverse, Active, Const, Duplicated
+
+# TODO: overload the following as well
+#       - Zygote.pullback
+#       - Zygote.jacobian
+#       - Zygote.hessian
+
+@reactant_overlay function Zygote.gradient(f::F, args...) where {F}
+    # TODO: check `f` as well once #1642 is merged
+    if use_overlayed_version(args)
```
wsmoses: If we do this, we should quite aggressively yell that we're gonig to do this -- I would even be okay saying to do this for each call [not even each callsite]

avik-pal: done

wsmoses: I'm also kind of debating if we want to have this behind a feature flag as well [as perhaps it is useful to compare the performance of zygote as a frontend vs us inside the compiler]

ext/ReactantZygoteExt.jl
```diff
@@ -0,0 +1,28 @@
...
+        @warn "Reactant doesn't support using Zygote for computing gradients. Replacing \
+               `Zygote.gradient` with `Enzyme.autodiff` call. Please update your code to \
+               not use `Zygote.gradient` inside `Reactant.@compile`." maxlog = 1
```
wsmoses: I'd potentially even be evil and not do maxlog=1, or at least something where we have maxlog1 from the callsite of zygote

wsmoses: also maybe we should say "not use Zygote.gradient and instead use Enzyme.gradient"

ext/ReactantZygoteExt.jl
avik-pal: That is what I was testing first, but we need to fix our broadcasting quirks before we can work through ChainRules rrules.

wsmoses: yeah....we should definitely fix that

ext/ReactantZygoteExt.jl
```diff
@@ -0,0 +1,29 @@
...
+
+@reactant_overlay function Zygote.gradient(f::F, args...) where {F}
+    # TODO: check `f` as well once #1642 is merged
+    if use_overlayed_version(args) && Reactant.OVERLAY_ZYGOTE_CALLS[]
+        @warn "Reactant doesn't support using Zygote for computing gradients. Replacing \
+               `Zygote.gradient` with `Enzyme.autodiff` call. Please update your code to \
+               not use `Zygote.gradient` and instead use `Enzyme.gradient` inside \
+               `Reactant.@compile`."
+        dargs = map(Enzyme.make_zero, args)
+        duplicated = map(Duplicated, args, dargs)
```
wsmoses: Is there a reason not to just call Enzmye.gradient here?

ext/ReactantZygoteExt.jl
avik-pal: This will error inside Zygote but we have an option

```julia
Reactant.with_config(; overlay_zygote_calls=false) do
    @jit Zygote.gradient(sumabs2, x)
end
```

src/Configuration.jl
```diff
@@ -379,3 +387,6 @@ function DotGeneralAlgorithm(
 
     return nothing
 end
+
+# Overlay Zygote.jl
+const OVERLAY_ZYGOTE_CALLS = ScopedValue(true)
```
wsmoses, line 392: I feel like it might be better to default to false, at least to start? could be convinced either way

avik-pal: I don't like the idea of a default that errors out almost always. Rn we should always work with the switching. If anyone really disagrees with switching they can easy opt-out in which case their code will just crash

avik-pal:
> Or the idea is to let people keep their existing code?

The idea is to make people progressively switch to Enzyme inside Reactant. Earlier today one of our students had switched the models and parameters to use reactant (xref https://github.com/SciML/NeuralPDE.jl/pull/967) but did not switch the AD she was using.

Instead of crashing we now do the switching with a loud enough warning that users know that they need to eventually fix this behavior.

src/Configuration.jl
wsmoses: I suppose that's fair, and I'm okay with this. I guess the specific caveat being that I think we should reserve the right to swap the default (and do so once either more downstream things are set to use enzyme properly and/or we fix broadcasting or other limitations)

avik-pal:
marking as draft to avoid accidental merge before others have a chance to comment

src/Configuration.jl
wsmoses: and yeah I do agree it's a lot better to get an early error message with a backtrace where it's at least possible to see where to do the switch

avik-pal: Updated the text a bit more.

ChrisRackauckas:
Best way to fix people's code and performance 10/10

avik-pal:
@giordano @vchuravy any final thoughts on this?

giordano:
I'm mildly in favour of this.

PR #1662, opened 2025-09-12
feat: caching code in mapreduce style ops

avik-pal:


src/TracedRArray.jl
```diff
@@ -1492,29 +1500,57 @@ end
 
 struct BroadcastIterator{F}
     f::F
+
+    BroadcastIterator{F}(f::F) where {F} = new{F}(f)
+    BroadcastIterator(f::F) where {F} = new{F}(f)
 end
 
-(fn::BroadcastIterator)(args...) = Reactant.call_with_reactant(fn.f, (args...,))
+(fn::BroadcastIterator)(args...) = fn.f((args...,))
```
jumerckx, line 1517: Trying to understand what this is doing. Why isn't `call_with_reactant` required anymore, or why was it required before?

avik-pal: This is meant to be only used instead `broadcast` which with force the `call_via_reactant` in `elem_apply` (it should have not been added in the first place)

jumerckx:
So this looks reasonable to me (though not sure if the CI failure is related to this pr?)

One major caveat is that the caching mechanism for `Ops.call` isn't correct if the function mutates a non-traced Julia object. e.g.:
for 
```jl
function f(X, y)
	push!(X, y)
end
# X::Vector{TracedRArray}
# y::TracedRArray
```
because in this case tracing through `f` has a side-effect that isn't captured when simply inserting a `func.call` to the existing `func.func`.

Billy mentioned the need for an analysis that tells whether such a mutation occurs. I've looked into this a little bit but haven't yet found a good way to do this.

avik-pal:
Ah I see, in that case, let's hold off on the caching part. I will separate out the non-caching fixes in a separate PR

PR #1666, opened 2025-09-15
mem effects

wsmoses:


src/TracedUtils.jl
```diff
@@ -269,6 +269,22 @@ mutable struct CompiledMlirFnResult{F,TR,Re,Rt,LA,LR,PA,CR,M,MA,RS,GD,DA}
...
+    is_pure::Bool
+end
+
+function is_pure(func)
+    attr = MLIR.IR.attr(func, "enzymexla.memory_effects")
+    # conservatively assume is not pure
+    if attr isa Nothing
+        return false
+    end
+    for i in 1:length(attr)
```
giordano: I'm on the phone, but I think this entire for loop until the final return can be written as
```julia
any(at -> String(at) == "write", attr)
```

wsmoses: problem is that attr doesn't define an iterator

giordano: It has indices but isn't iterable? That sounds unfortunate, the basic iteration interface is a single function: https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration

wsmoses: yeah I mean ironically the definition of that is in this repo (attribute.jl)

jumerckx: I can push this if you want?
```diff
diff --git a/src/TracedUtils.jl b/src/TracedUtils.jl
index 64619741..7d6d4de1 100644
--- a/src/TracedUtils.jl
+++ b/src/TracedUtils.jl
@@ -278,12 +278,7 @@ function is_pure(func)
     if attr isa Nothing
         return false
     end
-    for i in 1:length(attr)
-        sa = Base.String(attr[i])
-        if sa == "write"
-            return false
-        end
-    end
+    any(at -> String(at) == "write", attr) && return false
     return true
 end
 
diff --git a/src/mlir/IR/Attribute.jl b/src/mlir/IR/Attribute.jl
index 1321c3e9..ec12785d 100644
--- a/src/mlir/IR/Attribute.jl
+++ b/src/mlir/IR/Attribute.jl
@@ -802,6 +802,14 @@ function Base.getindex(attr::Attribute, i)
     end
 end
 
+function Base.iterate(attr::Attribute, state=1)
+    if state > length(attr)
+        nothing
+    else
+        (attr[state], state + 1)
+    end
+end
+
 function Base.getindex(attr::Attribute)
     @assert isdenseelements(attr) "attribute $(attr) is not a dense elements attribute"
     @assert issplat(attr) "attribute $(attr) is not splatted (more than one different elements)"

```

wsmoses: go for it!

src/Compiler.jl
```diff
@@ -3209,8 +3217,9 @@ Generate Julia code to call the XLA executable.
 
 - `flatten_names`: A list of `Symbol`s representing the names of the flattened linear arguments.
 - `nresults`: The number of results to expect.
+- `is_pure`: Whether the function being compiled is pure (e.g. has a side effect, like MPI.Send)
```
romanlee:
```suggestion
- `is_pure`: Whether the function being compiled is pure (i.e., has no side effects)
```
a little confusing as is

src/TracedUtils.jl
```diff
@@ -416,6 +427,7 @@ function make_mlir_fn(
         missing,
         global_device_ids,
         nothing, # populated later in `compile_mlir!`
+        is_pure(func2)
```
romanlee, line 430: Minor: but not necessary to actually calculate this here since it's recalculated in `compile_mlir!` I think. Could just pass `false` instead.

romanlee:
Looks good from my side besides the comments above

src/TracedUtils.jl
wsmoses: the mlir func may go out of scope so we do need to precompute

romanlee: Oh I see

giordano:
This broke building the docs https://github.com/EnzymeAD/Reactant.jl/actions/runs/17802463542/job/50605626751

wsmoses:
oh let me fix that in post, I thought docs failure was expected from some ssl issue but thats clearly not

Issue #1659, opened 2025-09-11
Restructure Testing + Downstream Testing


avik-pal:
Currently we have an ever-growing list of test deps, most being used only for a subset of tests. As a starting point, we can split up the tests as is into

- [x] `core`: this uses the `test/Project.toml`
- [x] `integration`: for extensions that reside in `ext/`
- [x] `nn`: move these into `test/downstream/Flux` & `test/downstream/Lux` with separate Project.toml

avik-pal:
## Downstream Testing

- [x] Lux & LuxLib
- [x] Flux
- [ ] Oceananigans

mofeing:
we could also take the time to organize the tests in the ever-growing https://github.com/EnzymeAD/Reactant.jl/blob/main/test/basic.jl file

PR #1675, opened 2025-09-17
Alias `TracedStepRangeLen` inside `TracedRNumberOverrides` for compatibility

giordano:
Ref https://github.com/CliMA/Oceananigans.jl/pull/4792 and https://github.com/CliMA/Oceananigans.jl/pull/4793#issuecomment-3303516473

### DIFF
--- a/src/TracedRNumber.jl
+++ b/src/TracedRNumber.jl
@@ -5,6 +5,11 @@ using ..Ops: @opcall
 using ReactantCore: ReactantCore, @trace
 using Adapt: Adapt
 
+# This isn't technically necessary in this module, but this type used to be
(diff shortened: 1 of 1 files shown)


avik-pal:
We should start adding some kind of downstream testing like we have started adding in Enzyme. Tracking in https://github.com/EnzymeAD/Reactant.jl/issues/1659#issuecomment-3303846364

giordano:
Tested successfully in https://github.com/EnzymeAD/Enzyme-JAX/pull/1399, I'm going to merge this even if CI hasn't finished yet (the change here doesn't really affect much the rest of the code) so that Enzyme-JAX can work again.

PR #1681, opened 2025-09-18
apply_type_with_promotion

jumerckx:
While in my testing `traced_type` now does the typevar promotion we talked about, it doesn't yet work for `make_tracer` as the result from `traced_types` isn't used to guide `make_tracer` AFAICT.

```jl
using Reactant

struct Bar{T}
    b::T
end
struct Foo{T, B<:Bar{T}, AT<:AbstractArray{T}}
    a::AT
    b::B
end

foo = Reactant.to_rarray(Foo(rand(3), Bar(1.)))

Reactant.traced_type(typeof(foo), Val(Reactant.ConcreteToTraced), Union{}, nothing, nothing)
# returns: Foo{Reactant.TracedRNumber{Float64}, Bar{Reactant.TracedRNumber{Float64}}, Reactant.TracedRArray{Float64, 1}}

Reactant.make_tracer(Reactant.OrderedIdDict(), foo, (), Reactant.ConcreteToTraced) # fails
```
```
ERROR: AssertionError: Could not recursively make tracer of object of type Foo{Float64, Bar{Float64}, ConcretePJRTArray{Float64, 1, 1, Reactant.Sharding.ShardInfo{Reactant.Sharding.NoSharding, Nothing}}} into Foo{Reactant.TracedRNumber{Float64}, Bar{Reactant.TracedRNumber{Float64}}, Reactant.TracedRArray{Float64, 1}} at field 2 (named b), need object of type Bar{Reactant.TracedRNumber{Float64}} found object of type Bar{Float64} 
Stacktrace:
 [1] make_tracer_unknown(seen::Reactant.OrderedIdDict{…}, prev::Any, path::Any, mode::Reactant.TraceMode; track_numbers::Type, sharding::Any, runtime::Any, kwargs::@Kwargs{})
   @ Reactant ~/Reactant2/src/Tracing.jl:1197
 [2] make_tracer_unknown
   @ ~/Reactant2/src/Tracing.jl:1095 [inlined]
 [3] #make_tracer#139
   @ ~/Reactant2/src/Tracing.jl:1232 [inlined]
 [4] make_tracer(seen::Reactant.OrderedIdDict{Any, Any}, prev::Any, path::Any, mode::Reactant.TraceMode)
   @ Reactant ~/Reactant2/src/Tracing.jl:1222
 [5] top-level scope
   @ ~/Reactant2/make_tracer_debuggin.jl:12
```


src/Tracing.jl
```diff
@@ -745,6 +745,14 @@ Base.@nospecializeinfer function traced_type_inner(
     if fieldcount(T) == fieldcount(TT2)
         legal = true
         for f in 1:fieldcount(T)
+            if isa(Base.unwrap_unionall(T.name.wrapper).types[f], TypeVar)
+                # The field is constrained by a TypeVar directly,
+                # so we don't need to check.
+                # (The check below would fail if the typevar was promoted as
+                # we don't get the same result when calling traced_type_inner
```
wsmoses: don't we still need to check below that the subtype matches, even if constrained? just to confirm we did do the typevar solve correctly?

src/Tracing.jl
```diff
@@ -857,6 +865,79 @@ const traced_type_cache = Dict{Tuple{TraceMode,Type,Any},Dict{Type,Type}}()
 #     $(Expr(:meta, :generated, traced_type_generator))
 # end
 
+"""
+This function tries to apply the param types to the wrapper type.
+When there's a constraint conflict, it tries to resolve it by promoting the conflicting types. The new param type is then propagated in any param type that depends on it.
+"""
+function apply_type_with_promotion(wrapper, params, relevant_typevars=typevar_dict(wrapper))
```
wsmoses, line 933: can we add some tests?

src/Tracing.jl
jumerckx: I'm not sure...
The reason I added this early return is for e.g.:
`Foo{Float64, Bar{Float64}, ConcretePJRTArray{Float64, 1}}`
Which, with `ConcreteToTraced` is now converted to:
`Foo{TracedRNumber{Float64}, Bar{TracedRNumber{Float64}}, TracedRArray{Float64, 1}}`

The field `b` of the original type is `Bar{Float64}`. When tracing this, it just returns `Bar{Float64}` because of lack of constraints. So what should we compare the fieldtype against?

wsmoses:
```
struct T
   d::Array
end
```

wsmoses:
```
struct T{D<:Array}
   d::D
end


struct T{D<:AbstractArray{Float64}}
   d::D
end
```

wsmoses: check can be skipped when an unconstrained typevar
+ must be skipped when that typevar is itself a dependency of another type. If this is true in any case, we need the fancy make_tracer
+ may be run when that typevar is itself not a dependency of another type. If this is true in all cases, we don't need the upgrade

wsmoses: case where need to skip checj
```
struct MyNewArray{FT, N, D <: AbstractArray{FT, N}}
    data::D
    off::FT
end
```

jumerckx:
I added a bunch of debug statements that seem handy to keep around.
However, they clutter CI. Should I just get rid of them, or look for a solution?

src/Tracing.jl
```diff
@@ -710,11 +732,18 @@ Base.@nospecializeinfer function traced_type_inner(
         return T
     end
 
+    @debug "traced_type_inner: Processing type with field changes" T = T subTys = subTys
```
wsmoses: I would probably get rid of the debug statements here, if no longer necessary

wsmoses:
@giordano adding you as a reviewer.

this should remedy many of the custom type annotations weve historically needed for oceananigans and fixedsizearray

jumerckx:
Not sure how I should fix the CI docs failure. It's an internal method but still useful to have a docstring imo.

giordano:
Add it to https://enzymead.github.io/Reactant.jl/stable/api/api#Internal-utils

avik-pal (approved):

jumerckx:
This looks like it might actually be related to the changes. I'll search for these output dumps in a bit.
```
...
┌ Error: Compilation failed, MLIR module written to /__w/Reactant.jl/Reactant.jl/tmp/reactant_27DS3r/module_017_reactant_copysign_post_xla_compile.mlir
└ @ Reactant.MLIR.IR /__w/Reactant.jl/Reactant.jl/src/mlir/IR/Pass.jl:119
┌ Error: Compilation failed, MLIR module written to /__w/Reactant.jl/Reactant.jl/tmp/reactant_27DS3r/module_018_reactant_copysign_post_xla_compile.mlir
└ @ Reactant.MLIR.IR /__w/Reactant.jl/Reactant.jl/src/mlir/IR/Pass.jl:119
┌ Error: Compilation failed, MLIR module written to /__w/Reactant.jl/Reactant.jl/tmp/reactant_27DS3r/module_019_reactant_copysign_post_xla_compile.mlir
└ @ Reactant.MLIR.IR /__w/Reactant.jl/Reactant.jl/src/mlir/IR/Pass.jl:119

```

PR #1691, opened 2025-09-20
adpt to activity changes

wsmoses:


src/Enzyme.jl
```diff
@@ -27,7 +27,7 @@ macro register_make_zero_inplace(sym)
         end
 
         @inline function $sym(prev::RArray{T,N}, seen::ST)::Nothing where {T,N,ST}
-            if Enzyme.Compiler.guaranteed_const_nongen(T, nothing)
```
avik-pal, line 30: we probably need a min-version bump?

wsmoses: Ah yeah, sorry away from computer but feel free to take the branch if unblocks

Issue #1673, opened 2025-09-17
JIT session error: Failed to materialize symbols in windows tests


avik-pal:
```julia
JIT session error: Failed to materialize symbols: { (enzymejitdl_9, { __real@fffffffff8000000, __real@4170000000000000, __real@3f847ae147ae147b, __real@4012000000000000, __real@4012051eb851eb85, __real@4340000000000000, __xmm@7fffffffffffffff7fffffffffffffff, __real@3ff0000000000000, __real@c3e0000000000000, __real@43e0000000000000 }) }
 lookupError Failed to materialize symbols: { (enzymejitdl_9, { entry }) }
```

https://github.com/EnzymeAD/Reactant.jl/actions/runs/17796071021/job/50583891541#step:11:924

avik-pal:
```
JIT session error: Failed to materialize symbols: { (enzymejitdl_10, { __real@3fe0000000000000 }) }
 lookupError Failed to materialize symbols: { (enzymejitdl_10, { entry }) }
```

https://github.com/EnzymeAD/Reactant.jl/actions/runs/17796071021/job/50583891582#step:11:670

wsmoses:
cc @giordano presumably jll [xla] bump related

PR #1693, opened 2025-09-21
ci: use a shared workflow

avik-pal:


.github/workflows/CommonCI.yml
```diff
@@ -0,0 +1,205 @@
...
+      #   # We have to use `${GITHUB_WORKSPACE}` instead of `github.workspace` because GitHub
+      #   # is terrible and the two don't match inside containers:
+      #   # https://github.com/actions/runner/issues/2058
+      #   if: ${{ inputs.os != 'windows-latest' }}
+      #   run: |
+      #     echo "TMPDIR=${GITHUB_WORKSPACE}/tmp" >> ${GITHUB_ENV}
+      # - name: Make TMPDIR
+      #   if: ${{ inputs.os != 'windows-latest' }}
+      #   run: |
+      #     mkdir -p ${{ env.TMPDIR }}
```
avik-pal: I need to figure out why this doesn't create the directory

avik-pal:
The main motivation apart from reducing redundancy here, is that I am working on a PR to separate the integration testing into separate workflows to avoid the bazillion test deps. It would be nice to not copy around the full workflow

avik-pal:
CI seems happy
1. windows CI failures are due to https://github.com/EnzymeAD/Reactant.jl/issues/1673
2. assertions CI is disabled till https://github.com/EnzymeAD/Reactant.jl/actions/runs/17895941712/job/50882582947#step:11:1675 is resolved. Seems to originate from CUDA.jl

.github/workflows/CommonCI.yml
```diff
@@ -0,0 +1,215 @@
...
+          export TMPDIR="${GITHUB_WORKSPACE}/tmp"
+          mkdir -p "$TMPDIR"
+          echo "TMPDIR=$TMPDIR" >> $GITHUB_ENV
+
+      - uses: julia-actions/setup-julia@v2
+        if: ${{ ! inputs.assertions }}
+        with:
+          version: ${{ inputs.julia_version }}
+      - uses: julia-actions/cache@v2
+        id: julia-cache
```
giordano, line 71: One quick issue I noticed is that there are cache conflicts: https://github.com/EnzymeAD/Reactant.jl/actions/runs/17896155900/job/50883073314?pr=1693#step:45:9.  That's is an upstream issue: https://github.com/julia-actions/cache/issues/161, this is how I addressed it in GB-25: https://github.com/PRONTOLab/GB-25/pull/71.

avik-pal: Should be fixed now https://github.com/EnzymeAD/Reactant.jl/actions/runs/17897156838/job/50885295765?pr=1693#step:45:14

.github/workflows/CommonCI.yml
```diff
@@ -0,0 +1,217 @@
...
+          bazelisk-cache: true
+          # Store build cache per workflow.
+          disk-cache: ${{ github.workflow }}-${{ inputs.os }}-${{ inputs.julia_version }}
+          # Share repository cache between workflows.
+          repository-cache: true
+          bazelisk-version: 1.x
+      - name: Prepare build on macOS
+        if: ${{ startsWith(inputs.os, 'macOS-') && inputs.localjll }}
+        run: |
+          echo "SDKROOT=$(xcrun --show-sdk-path)" >> "${GITHUB_ENV}"
```
giordano, line 96: How's this used?  Is that picked up automatically by Bazel?  Otherwise this seems to be completely unused.

.github/workflows/CommonCI.yml
```diff
@@ -0,0 +1,217 @@
...
+          dev_pks = Pkg.PackageSpec[]
+          for path in ("lib/ReactantCore",)
+              push!(dev_pks, Pkg.PackageSpec(; path))
+          end
+          Pkg.develop(dev_pks)
+        shell: julia --color=yes --code-coverage=user --depwarn=yes --project=. {0}
+        # Only in Julia v1.10 we need to install `ReactantCore` manually.
+        if: ${{ inputs.julia_version == '1.10' || inputs.julia_version == 'lts' }}
+        env:
+          JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
```
giordano: Maybe this can be moved up to the job-level environment, instead of having to repeat this var every time?

.github/workflows/CommonCI.yml
```diff
@@ -0,0 +1,217 @@
...
+        timeout-minutes: 10
+        if: ${{ always() }}
+        with:
+          name: "mlir-${{ inputs.julia_version }}-${{ inputs.os }}-${{ inputs.runtime }}-assertions=${{ inputs.assertions }}-${{ github.event_name }}-test_group=${{ inputs.test_group }}-${{ inputs.localjll }}"
+          path: "**/*.mlir"
+          retention-days: 90
+          overwrite: false
+
+      - name: Save Julia depot cache on cancel or failure
+        if: ${{ inputs.localjll && (cancelled() || failure()) }}
```
giordano: I think it'd be worth to always do the upload in case of failure, not just when doing a local build (although with the upcoming cache restrictions we'll need to rethink our caching entirely):
```suggestion
        if: cancelled() || failure()
```

giordano (commented):
Looks overall good to me, I'm always happy to remove duplication 👍

.github/workflows/CommonCI.yml
avik-pal: not sure. I picked it from the old local jll script. That said the macos local jll has been broken for ages so this might very well be unused

giordano (approved):

PR #1694, opened 2025-09-21
test: reduce test dependencies and setup integration CI

avik-pal:
Time to instantiate and precompile down from ~10mins to ~3mins

- [x] enable all testing
- [x] Merge `ap/cleanup_ci`
- [ ] compat
- [x] runtimes are not being correctly set for integration testing
- [ ] move some of the CUDA tests for integration


avik-pal:
This should be ready for a review now

PR #1700, opened 2025-09-24
fix: try to make searchsortedfirst raising work again

avik-pal:
I can confirm that we are now hitting the custom dispatch for `searchsortedfirst` but raising still seems to fail

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -4,7 +4,7 @@ using Base: Broadcast
 using Base.Broadcast: Broadcasted, AbstractArrayStyle, instantiate
 
 using ..Reactant: Reactant, TracedRArray, TracedRNumber, AnyTracedRArray, AnyTracedRVector
-using ..Reactant: MLIR, unwrapped_eltype
+using ..Reactant: MLIR, unwrapped_eltype, CuTracedRNumber, CuTracedArray
 using ..Ops: @opcall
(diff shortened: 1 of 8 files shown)


wsmoses:
if it hits raising that means we're likely not hitting a relevant overload?

src/TracedRArray.jl
```diff
@@ -1181,14 +1181,13 @@ end
...
+    (AbstractRange{<:CuTracedRNumber}, Real),
     (AbstractRange{<:TracedRNumber}, TracedRNumber{<:Real}),
+    (AbstractRange{<:CuTracedRNumber}, CuTracedRNumber{<:Real}),
 )
-    @eval function Base.searchsortedfirst(
-        a::$(aType), x::$(xType), o::Base.DirectOrdering
-    )::TracedRNumber{keytype(a)}
-        x = TracedUtils.promote_to(TracedRNumber{Reactant.unwrapped_eltype(a)}, x)
-
+    @eval function Base.searchsortedfirst(a::$(aType), x::$(xType), o::Base.DirectOrdering)
```
avik-pal, line 1190: It is hitting this (confirmed with the good old way of changing the function and checking the traced mlir). not sure if we want to call into some other function

wsmoses: hm who is still calling the kernel then, that's weird

PR #1707, opened 2025-09-26
Use Enzyme.ignore_derivatives, now that landed

wsmoses:


src/Enzyme.jl
```diff
@@ -586,14 +586,14 @@ function overload_autodiff(
...
-    ignore_derivatives(args...)
+    Enzyme.ignore_derivatives(args...)
 
 Prevents the flow of gradients (and higher-order derivatives) by creating a new value that
 is detached from the original value. This is an identity operation on the primal. This can
 be applied on a nested structure of arrays and we will apply the operation on each of the
 leaves.
 """
-function ignore_derivatives(args...)
+@reactant_overlay function Enzyme.ignore_derivatives(args...)
```
avik-pal: let's move this into the Overlay file to avoid messing up Revise

src/Enzyme.jl
```diff
@@ -586,14 +586,14 @@ function overload_autodiff(
 end
 
 """
-    ignore_derivatives(args...)
+    Enzyme.ignore_derivatives(args...)
```
avik-pal: Also lets add an
```
const ignore_derivatives = Enzyme.ignore_derivatives
```

somewhere in Reactant module to prevent downstream breakages

avik-pal (approved):

PR #1669, opened 2025-09-16
FixedSizeArrays support (And Memory Support)

Qfl3x:
#1649 
Struggling with the Memory object. This is what I'm getting currently:
```
ERROR: NoFieldMatchError:
Cannot convert type Memory{Float32}, best attempt Memory{ConcretePJRTNumber{Float32, 1, Reactant.Sharding.ShardInfo{Reactant.Sharding.NoSharding, Nothing}}} failed.
This could be because the type does not capture the fieldtypes that should be converted in its type parameters.
name=length idx=1 Derived: ConcretePJRTNumber{Int64, 1, Reactant.Sharding.ShardInfo{Reactant.Sharding.NoSharding, Nothing}} Existing: Int64 Best Attempt: Int64
name=ptr idx=2 Derived: Ptr{Nothing} Existing: Ptr{Nothing} Best Attempt: Ptr{Nothing}

Stacktrace:
  [1] traced_type_inner(T::Type, seen::Dict{Type, Type}, mode::Reactant.TraceMode, track_numbers::Type, sharding::Any, runtime::Any)
    @ Reactant ~/projects/Reactant.jl/src/Tracing.jl:759
  [2] traced_type(T::Type, ::Val{Reactant.ArrayToConcrete}, track_numbers::Type, sharding::Reactant.Sharding.NoSharding, runtime::Val{:PJRT})
    @ Reactant ~/projects/Reactant.jl/src/Tracing.jl:870
  [3] make_tracer_unknown(seen::Reactant.OrderedIdDict{Any, Any}, prev::Any, path::Any, mode::Reactant.TraceMode; track_numbers::Type, sharding::Any, runtime::Any, kwargs::@Kwargs{device::Nothing, client::Nothing})
    @ Reactant ~/projects/Reactant.jl/src/Tracing.jl:1032
  [4] make_tracer_unknown
    @ ~/projects/Reactant.jl/src/Tracing.jl:1009 [inlined]
  [5] #make_tracer#191
    @ ~/projects/Reactant.jl/src/Tracing.jl:1146 [inlined]
  [6] make_tracer
    @ ~/projects/Reactant.jl/src/Tracing.jl:1136 [inlined]
  [7] make_tracer(seen::Reactant.OrderedIdDict{…}, prev::FixedSizeArray{…}, path::Any, mode::Reactant.TraceMode; kwargs::@Kwargs{…})
[...]

PR #1712, opened 2025-09-27
Allow passing custom size to Array-like `make_tracer` methods

giordano:
This allows creating traced arrays with shape different than the original data, without having to add extra `reshape` instructions.  Ref: https://github.com/EnzymeAD/Reactant.jl/pull/1669#discussion_r2373575794.


wsmoses:
I'm not convinced this is the best design. It feels like it would be better to specialize fixedsizeareays to retain the shape information than to pipe through arrays for all trace overrides. Eg I'm unclear who other than fixed size arrays would need (and also am somewhat worried about the soundness of an outer array propagation into)

PR #1721, opened 2025-10-01
Add new optimization patterns to Compiler.jl

avik-pal:


src/Compiler.jl
```diff
@@ -922,6 +924,7 @@ function optimization_passes(
                 "concat_insert_dim_reduce",
                 "concat_insert_dim_sort",
                 "concat_insert_dim_reduce_window",
+                "concat_insert_dim_elementwise",
```
avik-pal, line 927:
```suggestion
                # XXX: busted `map of slices` tests, needs upstream fix
                # "concat_insert_dim_elementwise",
```

wsmoses (approved):

PR #1755, opened 2025-10-15
fix: loading reactant on 1.12

avik-pal:
Enables loading on 1.12 mostly to get automerge to pass. I am printing a warning if someone tries to use Reactant on 1.12

### DIFF
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -1677,5 +1677,5 @@ end
     x_ra2 = Reactant.to_rarray(rand(Float64, 2))
 
     fn = @compile sum(x_ra1)
-    @test_throws Reactant.MisMatchedThunkTypeError fn(x_ra2)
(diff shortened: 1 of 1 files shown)


src/Reactant.jl
```diff
@@ -330,6 +330,19 @@ function __init__()
         end
     end
 
+    @static if VERSION > v"1.11"
+        if ccall(:jl_generating_output, Cint, ()) == 1
+            @warn """
+            Reactant.jl currently doesn't support versions of Julia 1.12 or newer. We are
+            actively working on adding support for newer versions of Julia. For the time
+            being we recommend using 1.10 (current Julia LTS) or 1.11.
```
wsmoses: let's probably say we recommend using Julia 1.11 or LTS [since currently 1.11 is actually more stable/better for reactant than 1.10, in contrast to enzyme]

wsmoses (approved):

PR #1760, opened 2025-10-16
[WIP] GPUCompiler + custom JIT

glou-nes:
Extract from #1754, use a custom GPUCompiler infra / ORC JIT to replace `OpaqueClosure`.
- Redefine `ReactantInterpreter` without depending on `Enzyme` interpreter. 
- Incorporate dynamic call rewrite in the type inference and remove fragile custom inference.

I only test this on 1.11,  it probably don't run on 1.10. Now, I mostly want to check if that approach can work at scale. I let all debug stuff atm.

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -1465,7 +1465,7 @@ end
 @static if !Sys.isapple()
     @setup_workload begin
         Reactant.initialize_dialect()
-        init_jit()
+        Reactant.init_jit()
         if Reactant.XLA.REACTANT_XLA_RUNTIME == "PJRT"
             client = Reactant.XLA.PJRT.CPUClient(; checkcount=false)
         elseif Reactant.XLA.REACTANT_XLA_RUNTIME == "IFRT"


src/JIT.jl
```diff
@@ -0,0 +1,632 @@
...
+@kwdef struct DebugData
+    enable_log::Bool = true
+    enable_runtime_log::Bool = true
+    rewrite_call::Set = Set()
+    non_rewrite_call::Set = Set()
+end
+
+struct ReactantToken end
+
+@kwdef struct ReactantInterpreter <: CC.AbstractInterpreter
```
wsmoses, line 83: it still may be useful to have this inherit from the enzyme interp (but we can still specialize the methods below if set to reactant_abi or w/e).

Is it possible/reasonable to do that?

src/JIT.jl
```diff
@@ -0,0 +1,632 @@
+using GPUCompiler
+CC = Core.Compiler
+
+#leak each argument to a global variable
+macro lk(args...)
```
wsmoses, line 5: can we rename this leak?

also I don't quite follow why this is needed?

wsmoses:
@gbaraldi can you help review?

src/JIT.jl
glou-nes: Sorry, I don't have cleanup my debug stuff. I will remove it.

src/JIT.jl
glou-nes: Is that possible to elaborate? I don't get what can be used from the enzyme interp.

wsmoses: longer term we need to retain the noinline/etc characteristics from the enzyme custom rules, so its preferable to build off the enzyme interp, and perhap specialize it on these methods for the reactant version (e.g. when we have reactant-specific type parameters)

cc @avik-pal

ext/ReactantCUDAExt.jl
```diff
@@ -1465,7 +1465,7 @@ end
 @static if !Sys.isapple()
     @setup_workload begin
         Reactant.initialize_dialect()
-
+        init_jit()
```
avik-pal:
```suggestion
        Reactant.init_jit()
```

glou-nes: Thanks!

wsmoses (commented):
bumping here just because code will get stale over time if not merged (relatedly the 1.12 fixes we did not conflict here)

glou-nes:
I got three kinds of issues atm. `ccall needs the compiler` inside some `call_with_reactant`; type inference changes leading to `MethodError` and `ccall_deferred` symbol unknown. It passes half Enzyme Reactant unit tests, and 2/3 of basics.

PR #1762, opened 2025-10-16
feat: julia api to access device properties

avik-pal:


src/xla/IFRT/Device.jl
```diff
@@ -31,6 +31,14 @@ function XLA.get_local_device_id(::Device)
     return error("Not implemented for ifrt devices")
 end
 
+function XLA.get_local_hardware_id(::Device)
```
avik-pal:
```suggestion
function XLA.get_local_hardware_id(device::Device)
```

avik-pal:
This should now be good to go

PR #1770, opened 2025-10-18
feat: convert Random.AbstractRNG to ReactantRNG in make_tracer

### LINKED ISSUE #1583: Convert Random.TaskLocalRNG in to_rarray?
MLDataDevices adapts `Random.TaskLocalRNG` to Random.TaskLocalRNG

```julia
adapt(ReactantDevice(), Random.default_rng()) isa Reactant.ReactantRNG
```

but `to_rarray`

```julia
Reactant.to_rarray(Random.default_rng()) isa Random.TaskLocalRNG
```

currently doesn't. So when users have, a data structure that contains both an RNG and some data, for example

```julia
dummy_mcmc_chain = (current = zeros(Float32, 10), rng = Random.default_rng(), buffer = zeros(Float32, 10))
dummy_mcmc_step!(chain) = chain.current .+= randn!(chain.rng, chain.buffer)
dummy_mcmc_step!(chain)
dummy_mcmc_step!(chain)
...
```

And move it to Reactant

```
reactant_mcmc_chain = Reactant.to_rarray(dummy_mcmc_chain)
reactant_mcmc_step! = @compile dummy_mcmc_step!(reactant_mcmc_chain)
reactant_mcmc_step!(reactant_mcmc_chain)
```

the random numbers drawn are always the same, because the state of `Random.default_rng()` at captured at compilation time and stored in `reactant_mcmc_step!`.

If users don't know the details of the code they are compiling, this may be an easy trap to fall into. Should we change `Reactant.to_rarray` to convert RNGs, at least when it's a `Random.TaskLocalRNG`?

### LINKED ISSUE #1765: Should ReactantDevice (and to_rarray) move all AbstractRNG to ReactantRNG?
See https://github.com/LuxDL/Lux.jl/issues/1515

Currently only `Random.default_rng` is being moved to `ReactantRNG`, while MersenneTwister and StableRNG are preserved

avik-pal:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/1765
fixes https://github.com/EnzymeAD/Reactant.jl/issues/1583

### DIFF
--- a/ext/ReactantRandom123Ext.jl
+++ b/ext/ReactantRandom123Ext.jl
@@ -4,8 +4,14 @@ using Random123: Threefry4x, Threefry2x, Philox4x, Philox2x
 using Reactant: TracedRandom
 
 TracedRandom.rng_algorithm(::Threefry4x) = "THREE_FRY"
-TracedRandom.rng_algorithm(::Threefry2x) = "THREE_FRY"
 TracedRandom.rng_algorithm(::Philox4x) = "PHILOX"
+
+TracedRandom.rng_algorithm(::Threefry2x) = "THREE_FRY"
+TracedRandom.should_warn_if_not_natively_supported(::Threefry2x) = nothing
+TracedRandom.make_seed(rng::Threefry2x) = UInt64[rng.key1, rng.key2]
+
 TracedRandom.rng_algorithm(::Philox2x) = "PHILOX"
+TracedRandom.should_warn_if_not_natively_supported(::Philox2x) = nothing
+TracedRandom.make_seed(rng::Philox2x) = UInt64[rng.ctr1, rng.ctr2, rng.key]
 
 end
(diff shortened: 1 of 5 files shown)


wsmoses (approved):
I'm overall rather skeptical of this as an arbitrary rng may have different properties like seeding distribution etc, which means we will give different (wrong) answers when converting to a generic reactant rng.

At minimum we should aggressively yell, but I'd prefer that we implement the specification of the rngs in terms of reactant native primitives

avik-pal:
Yeah that makes sense. For all RNGs we warn except, `ThreeFry2x` and `Philox2x` which are natively supported

bvdmitri:
> I'm overall rather skeptical of this as an arbitrary rng may have different properties like seeding distribution etc, which means we will give different (wrong) answers when converting to a generic reactant rng.

fair point, but I also think its fair for an RNG to produce different results on different hardware. The discrepancy can be explained that `StableRNG` is producing different streams on CPU and TPU/GPU but the streams are of course identical if executed within the same hardware family 

a warning is good in either case

**edit**: ah, you've probably meant if the original RNG is skewed towards particular values and `ReactantRNG` isnt the same...subtle indeed

PR #1782, opened 2025-10-21
[deps] Make CUDA build more robust

giordano:
* actually run code for symlinking ptxas, fatbinary, libdevice (the condition `build_backend == "cuda"` was now always false)
* check the source of the symlinks exists, instead of creating broken symlinks, if that part of the code ever runs (which at the moment it doesn't because of point above)
* rework backend handling, to use more structure instead of string comparisons (which lead to issues like the first point)

Ref: https://github.com/EnzymeAD/Enzyme-JAX/pull/1491.  ~Not fully tested end-to-end, but I did test argument parsing and that seems to be working correctly.~ Tested in https://github.com/EnzymeAD/Enzyme-JAX/pull/1496.

### DIFF
--- a/deps/build_local.jl
+++ b/deps/build_local.jl
@@ -115,36 +115,50 @@ source_dir = joinpath(@__DIR__, "ReactantExtra")
 # --@local_config_cuda//:cuda_compiler=nvcc
 # --crosstool_top="@local_config_cuda//crosstool:toolchain"
 
-build_kind = parsed_args["debug"] ? "dbg" : "opt"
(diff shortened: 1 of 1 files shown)


giordano:
This seems to be working well in https://github.com/EnzymeAD/Enzyme-JAX/pull/1496: https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/18685814501/job/53282009739#step:12:491.  Symlinks are created correctly, and since now we throw an error if source file isn't found, we should be alerted early if something is off.

PR #1798, opened 2025-10-29
clean up symm

snonk:
clean up commit history from previous PR 😅


avik-pal:
https://github.com/EnzymeAD/Reactant.jl/pull/2800

PR #1810, opened 2025-10-31
ProbProg: Static HMC JLL changes

sbrantq:


wsmoses:
honestly those belong even in enzyme proper, not even enzyme-jax

sbrantq:
Yeah will do that

PR #1815, opened 2025-10-31
fix: ensure randn respects exclusivity of maxval

avik-pal:


avik-pal:
finally resolves the long standing issue with https://github.com/LuxDL/Lux.jl/pull/1158

src/Ops.jl
```diff
@@ -1628,7 +1678,7 @@ distribution with rate 1. Returns a NamedTuple with the following fields:
     shape;
     algorithm::String="DEFAULT",
     location=mlir_stacktrace("rand", @__FILE__, @__LINE__),
-) where {T}
+) where {T<:AbstractFloat}
```
wsmoses, line 1693: Do we need to do union{abstractfloat,tracedabstractfloat}

avik-pal: I dont think passing in a traced float here ever worked

wsmoses: I suppose that's fair, I'm just wondering if it's reasonable to end up with, eg if someone said eltype of traced array.

In any case not a blocker, otherwise LGTM

PR #1822, opened 2025-11-03
fix: correct type for ad annotation

### LINKED ISSUE #1821: Incorrect activity guessed for ForwardMode
```
using Reactant, Enzyme

function f(x)
    return sum(abs2, x; dims=(2, 3))
end

x = Reactant.to_rarray(rand(Float32, 3, 4, 5, 6))
bx = Reactant.to_rarray(rand(Float32, 3, 4, 5, 6))

@code_hlo Enzyme.autodiff(Forward, f, Duplicated(x, bx))
```

`A` inside `overloaded_autodiff` is `Const{Union{}}`.

avik-pal:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/1821


src/Enzyme.jl
```diff
@@ -295,6 +295,16 @@ function act_attr(val)
     return MLIR.IR.Attribute(val)
 end
 
+function overload_autodiff(
+    mode::CMode, f::FA, args::Vararg{Annotation,Nargs}
+) where {CMode<:Mode,FA<:Annotation,Nargs}
+    # need to guess the correct activity here. Execute the function, we will DCE it
+    res = call_with_reactant(f.val, [x.val for x in args]...)
```
wsmoses: er, not always -- if f mutates this is wrong

wsmoses: the slightly more ideal case would be calling like https://github.com/EnzymeAD/Enzyme.jl/blob/bc031934aa2deefdcd3be139e4ad9a28e3692d14/src/typeutils/inference.jl#L73, except with our interpreter

PR #1823, opened 2025-11-03
fix: add checks for running shardy without raising

avik-pal:


wsmoses (approved):

src/Compiler.jl
```diff
@@ -1738,6 +1738,8 @@ function compile_mlir!(
             ()
         end
 
+    legal_to_run_shardy_passes = compile_options.optimization_passes !== :before_raise
```
wsmoses: we also may need before_jit

PR #1824, opened 2025-11-04
feat: enable new licm + loop unroll passes + inlining

avik-pal:


src/Compiler.jl
```diff
@@ -693,6 +693,7 @@ const AGGRESSIVE_SUM_TO_CONV = Ref(false)
 const AGGRESSIVE_PROPAGATION = Ref(false)
 const DUS_SLICE_SIMPLIFY = Ref(true)
 const CONCATS_TO_DUS = Ref(false)
+const WHILE_UNROLL_THRESHOLD = Ref(5)
```
avik-pal: @wsmoses is this limit too aggressive? 1 would probably be the safest limit

wsmoses: I'd probably pick a power of two like 4

avik-pal:
```suggestion
const WHILE_UNROLL_THRESHOLD = Ref(4)
```

PR #1826, opened 2025-11-05
Replace unzip_jll with p7zip_jll

nhz2:
`unzip_jll` is no longer maintained, and `p7zip_jll` is a standard library.


nhz2:
It turns out that p7zip_jll has a bug on Linux https://github.com/JuliaLang/julia/issues/60071

I tested, and the ".whl" files being extracted here do not trigger the problem, so I think it is okay to not revert this PR.

PR #1847, opened 2025-11-10
fix: finite diff gradient accidental type promotion + generalize inputs

avik-pal:
precision issue will take a look later

### DIFF
--- a/test/nn/luxlib.jl
+++ b/test/nn/luxlib.jl
@@ -85,12 +85,12 @@ end
     end
 
     @testset "Activation: $act" for (act, gradfn) in (
-            (identity, (Ω, x) -> one(Ω)),
-            (relu, (Ω, x) -> (Ω > 0)),
-            (sigmoid, (Ω, x) -> conj((1 - Ω) * Ω)),
-            (tanh, (Ω, x) -> conj(1 - Ω^2)),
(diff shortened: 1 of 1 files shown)


src/TestUtils.jl
```diff
@@ -33,23 +34,121 @@ function default_epslion(::Val{fdtype}, ::Type{T}) where {fdtype,T}
...
+    )
+end
+
+function finite_difference_gradient(f::F, args...) where {F}
+    argprefix = gensym("finitediffarg")
+    resprefix = gensym("finitediffresult")
+    resargprefix = gensym("finitediffresarg")
+
+    # TODO: can we detect and prevent using functions that mutate their arguments?
+    mlir_fn_res = TracedUtils.make_mlir_fn(
```
wsmoses, line 67: do we need to do the full make_mlir_fn here, or can we reuse the equivalent of traced call or traced functions to achieve the same effect?

avik-pal: I am using make_mlir_fn mostly to linearlize the arguments. not sure how to use traced_call here though

wsmoses: ah so it was linearization, not efficiency

wsmoses (approved):

PR #1855, opened 2025-11-11
feat: protoype for dynamically sized arrays

avik-pal:
```mlir
module @reactant_sum attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<?x2xf32> {enzymexla.memory_effects = [], tf.aliasing_output = 1 : i32}) -> (tensor<f32>, tensor<?x2xf32>) attributes {enzymexla.memory_effects = []} {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<?x2xf32>) -> tensor<2x?xf32>
    %1 = stablehlo.reduce(%0 init: %cst) applies stablehlo.add across dimensions = [0, 1] : (tensor<2x?xf32>, tensor<f32>) -> tensor<f32>
    %2 = stablehlo.transpose %0, dims = [1, 0] : (tensor<2x?xf32>) -> tensor<?x2xf32>
    return %1, %2 : tensor<f32>, tensor<?x2xf32>
  }
}
```

### DIFF
--- a/ext/ReactantNNlibExt/Implementations.jl
+++ b/ext/ReactantNNlibExt/Implementations.jl
@@ -52,7 +52,7 @@ function overloaded_conv!(
     end
 
     result = @opcall convolution(
-        collect(Int64, size(y)),
+        TracedUtils.collect_dynamic_size(y),
         x,
         weight;
         window_strides=collect(Int64, NNlib.stride(cdims)),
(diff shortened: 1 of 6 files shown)


avik-pal:
generally most ops wont work with unbounded dynamism. One thing we can do is to store the trace and on input run shape refinement.

PR #1860, opened 2025-11-12
Create TT client via plugin for TensTorrent devices

giordano:
Based on the template of #1626. This is very preliminary, but ***very*** basic stuff works if we ignore the result of a matmul is off by ~0.1%, which doesn't look terribly good (accuracy is ***a lot*** better on CPU):
```julia-repl
julia> using Reactant

julia> Reactant.devices()
32-element Vector{Reactant.XLA.PJRT.Device}:
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001bd189f0), "TT:0 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d000), "TT:1 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d0b0), "TT:2 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d160), "TT:3 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d210), "TT:4 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d2c0), "TT:5 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d370), "TT:6 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d420), "TT:7 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d4d0), "TT:8 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d580), "TT:9 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d630), "TT:10 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d6e0), "TT:11 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d790), "TT:12 Wormhole_b0")
 Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000001c16d840), "TT:13 Wormhole_b0")
[...]

PR #1866, opened 2025-11-13
Define `XLA.DEFAULT_DEVICE` to set default device ID

giordano:


src/xla/Client.jl
```diff
@@ -13,4 +13,8 @@ function get_device end
 function get_addressable_device end
 function platform_name end
 
-default_device(client::AbstractClient) = first(addressable_devices(client))
+function default_device(client::AbstractClient)
+    return addressable_devices(client)[something(
+        tryparse(Int, get(ENV, "REACTANT_DEFAULT_DEVICE", "1")), 1
```
giordano: Should `REACTANT_DEFAULT_DEVICE` be 0-based?

wsmoses: yeah I would say so

giordano:
To demonstrate this works, on top of #1860 I get
```julia-repl
julia> using Reactant

julia> Reactant.XLA.default_device()
Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x0000000042369bd0), "TT:0 Wormhole_b0")

julia> ENV["REACTANT_DEFAULT_DEVICE"] = "31"
"31"

julia> Reactant.XLA.default_device()
Reactant.XLA.PJRT.Device(Ptr{Nothing}(0x000000004d618490), "TT:31 Wormhole_b0")
```

src/xla/Client.jl
```diff
@@ -13,4 +13,10 @@ function get_device end
 function get_addressable_device end
 function platform_name end
 
-default_device(client::AbstractClient) = first(addressable_devices(client))
+function default_device(client::AbstractClient)
+    return addressable_devices(client)[something(
+        # `REACTANT_DEFAULT_DEVICE` is interpreted as 0-based.
+        tryparse(Int, get(ENV, "REACTANT_DEFAULT_DEVICE", "0")) + 1,
```
wsmoses: can we do something like const REACTANT_DEFAULT_DEVICE = ...

just because default_device may be called in the hot path

giordano: Yeah, I suggested this could be also a global `const`, I can change that (and we don't need to deal with the parsing)

src/xla/Client.jl
```diff
@@ -13,10 +13,13 @@ function get_device end
 function get_addressable_device end
 function platform_name end
 
+"""
+    DEFAULT_DEVICE :: Ref{Int}
+
+0-based index of default device to use, by default 0 (first available device).
+"""
+const DEFAULT_DEVICE = Ref{Int}(0)
```
wsmoses, line 22: if we're not doing an env var we can leave this as 1-indexed. [reason for env var is cuz cuda_visible_devices and friends are 0-indexed]

wsmoses: also if you want we can still do DEFAULT_DEVICE = Ref{Int}(something(tryparse, ...)

giordano: Done.

giordano:
Just confirming this still works:
```julia-repl
julia> using Reactant

julia> Reactant.XLA.default_device()
Reactant.XLA.PJRT.Device(Ptr{Nothing} @0x00000000321be750, "TT:0 Wormhole_b0")

julia> Reactant.XLA.DEFAULT_DEVICE[] = 16
16

julia> Reactant.XLA.default_device()
Reactant.XLA.PJRT.Device(Ptr{Nothing} @0x0000000032833c10, "TT:16 Wormhole_b0")

julia> Reactant.XLA.DEFAULT_DEVICE[] = 31
31

julia> Reactant.XLA.default_device()
Reactant.XLA.PJRT.Device(Ptr{Nothing} @0x0000000032834660, "TT:31 Wormhole_b0")
```
Yay!

PR #1883, opened 2025-11-17
feat: more linear algebra operations (non factorization changes)

avik-pal:
factorization ops are a bit messed up upstream, need to polish those before supporting them.

- [x] `det`/`logabsdet`
- [x] `inv`
- [x] `cross`
- [x] `rmul!`, `lmul!`, `rdiv!`, `ldiv!` -- array and number
- [x] `normalize`/`normalize!` (via `__normalize!`)
- [x] fixes accidental promotion in `norm`
- [x] `issymmetric`, `ishermitian`, `isbanded` (by extension `isdiag`)
- [x] `cholesky`

Non Linear Algebera additions

- [x] `nextfloat`
- [x] `prevfloat`

### DIFF
--- a/test/integration/linear_algebra.jl
+++ b/test/integration/linear_algebra.jl
@@ -493,9 +493,7 @@ end
         @testset for T in (Float32, Float64, ComplexF32, ComplexF64)
             (T == ComplexF64 || T == Float64) && RunningOnTPU && continue
 
-            A = stack(
-                random_matrix_with_cond(T, 4, 4, 1.001) for _ in 1:6
-            )
+            A = stack(random_matrix_with_cond(T, 4, 4, 1.001) for _ in 1:6)
             A = reshape(stack((r * r' for r in eachslice(A; dims=3))), 4, 4, 3, 2)
             A_ra = Reactant.to_rarray(A)
 


src/TracedRArray.jl
```diff
@@ -29,6 +29,10 @@ Base.elsize(::Type{TracedRArray{T,N}}) where {T,N} = sizeof(T)
 # we use it
 Base.convert(T::Type{<:TracedRArray}, x::AbstractArray) = Reactant.promote_to(T, x)
 
+# Base.first is very common usecase for getting first element to get the type
+# inside LinearAlgebra.jl
+Base.first(x::TracedRArray{T,N}) where {T,N} = @allowscalar(x[1])
```
wsmoses: do we want to put the allowscalar here? I would've expected not (but we look to what gpuarrays do)

avik-pal: let me remove this and check if anything breaks. I ended up wrapping more of the functionality manually (mostly because of conditionals)

src/TracedRArray.jl
```diff
@@ -29,6 +29,10 @@ Base.elsize(::Type{TracedRArray{T,N}}) where {T,N} = sizeof(T)
...
+
```
avik-pal:
```suggestion
```

PR #1908, opened 2025-11-25
feat: new syrk passes + lowering

avik-pal:


src/Compiler.jl
```diff
@@ -909,8 +910,16 @@ function optimization_passes(
...
         "divide_negated_operands_simplify",
         "multiply_negated_operands_simplify",
+        "transpose_syrk_to_syrk",
+        "fuse_mul_into_syrk",
+        "fuse_add_into_syrk",
     ]
 
+    if !is_sharded
+        # these passes don't have optimized sharding implementations
+        append!(transform_passes_list, ["dot_general_to_syrk"])
```
wsmoses: can't we just (inefficiently) lower syrk back to a dot general at the end, if not supported

avik-pal: we do that as fallback.

But shardy passes are run post this transform, so a transform here would mean an implicit all-gather on CPU.

wsmoses (approved):

PR #1920, opened 2025-11-29
Add docs page on GPU kernel compilation and raising

gdalle:
Following the docs PR invitation in https://discourse.julialang.org/t/kernelabstractions-cuda-reactant-how-to-get-minimal-working-example/134164/

### DIFF
--- a/docs/src/tutorials/kernels.md
+++ b/docs/src/tutorials/kernels.md
@@ -58,7 +58,7 @@ yr = square_compiled(xr)
 @assert yr == xr .^ 2  # hide
 ```
 
-The Reactant-compiled function `square_compiled` now runs on whatever device Reactant requests, including CPU, GPU, TPU or distributed settings.
+The Reactant-compiled function `square_compiled` now runs on whatever device you request from Reactant, including CPU, GPU, TPU or distributed settings.
 It will not run on the device it was written for, nor will it require a CUDA-enabled device.
 
 ## Kernel raising
@@ -68,14 +68,13 @@ For more information, see the [corresponding documentation](@ref "Raising").
 
 ## Kernel differentiation
 
-If you want to compute derivatives of your kernel, combining Reactant with [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) is the best choice.
+You can combine with [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) to take derivatives of kernels. For more information of differntiating generic code with Enzyme+Reactant, see the [corresponding documentation](@ref "automatic-differentiation").
 
 ```@example
 import Enzyme
 ```
 
-Currently, you must use the `raise = true` and `raise_first = true` compilation options to make sure the kernel is raised before Enzyme performs automatic differentiation on the program.
-This restriction will be removed in future versions.
+When differentiating computational kernels, Enzyme can either differentiate the kernel directly, or differentiated the raised linear algebra. Using Enzyme outside of Reactant, Enzyme.jl differentiates the kernel directly. Currently within Reactant, only differentiating the raised code is supported (though supporting both is in progress). As a result, you must use the `raise = true` and `raise_first = true` compilation options to make sure the kernel is raised before Enzyme performs automatic differentiation on the program.
 
 ```@example
 sumsquare(x) = sum(square(x))


docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,73 @@
...
+  1.0
+  4.0
+  9.0
+ 16.0
+ 25.0
+```
+
+## Kernel compilation
+
+To compile such kernels with Reactant, you need to pass the option `raise=true` to the `@compile` or `@jit` macro.
```
wsmoses: you don't need to perform raise=true to compile with reactant, they'll just run natively as the existing kernel. I you want to convert it to a tensor form to enable linear algebra optimizations, raise needs to be set to true

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,73 @@
...
+Furthermore, the [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) package needs to be loaded (even on non-NVIDIA hardware).
+
```
wsmoses, line 43: hm this is a bug, in principle we should force load this for you [though in the bg it will load cuda.jl]

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,73 @@
...
+  1.0
+  4.0
+  9.0
+ 16.0
+ 25.0
+```
+
+## Differentiated kernel
+
+In addition, if you want to compute derivatives of your kernel with [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl), the option `raise_first=true` also becomes necessary.
```
wsmoses: can you explain more what raising is. I think this doc would be better written as a tutorial on raising rather than kernels

wsmoses: re this specifically, currently we only support differentiation rules for the raised tensors [the internal kernel representaiton derivatives are in progress cc @Pangoraw @avik-pal etc].

For now raising will enable differentiation to succeed, but also raising must be performed prior to differentiation, hence raisefirst

docs/src/tutorials/kernels.md
gdalle: Without loading CUDA first, I get the error I mentioned on Discourse:

```julia
julia> y = @jit raise=true square(x)
ERROR: MethodError: no method matching ka_with_reactant(::Int64, ::Nothing, ::KernelAbstractions.Kernel{…}, ::Reactant.TracedRArray{…}, ::Reactant.TracedRArray{…})
The function `ka_with_reactant` exists, but no method is defined for this combination of argument types.
Attempted to raise a KernelAbstractions kernel with Reactant but CUDA.jl is not loaded.
Load CUDA.jl using `using CUDA`. You might need to restart the Julia process (even if Revise.jl is loaded).
Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/Reactant/zlIsO/src/utils.jl:0 [inlined]
  [2] call_with_reactant(::typeof(Reactant.ka_with_reactant), ::Int64, ::Nothing, ::KernelAbstractions.Kernel{…}, ::Reactant.TracedRArray{…}, ::Reactant.TracedRArray{…})
    @ Reactant ~/.julia/packages/Reactant/zlIsO/src/utils.jl:944
  [3] (::KernelAbstractions.Kernel{…})(::Any, ::Vararg{…}; ndrange::Int64, workgroupsize::Nothing)
    @ ReactantKernelAbstractionsExt ~/.julia/packages/Reactant/zlIsO/ext/ReactantKernelAbstractionsExt.jl:118
...
```

docs/src/tutorials/kernels.md
gdalle: I'm on a Mac M3 and when I try kernel compilation without raising I get

```julia
julia> y = @jit square(x)
ERROR: CUDA driver not found
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] functional
    @ ~/.julia/packages/CUDA/x8d2s/src/initialization.jl:24 [inlined]
  [3] task_local_state!()
    @ CUDA ~/.julia/packages/CUDA/x8d2s/lib/cudadrv/state.jl:77
  [4] active_state
    @ ~/.julia/packages/CUDA/x8d2s/lib/cudadrv/state.jl:110 [inlined]
  [5] cufunction(f::typeof(gpu_square_kernel!), tt::Type{Tuple{…}}; kwargs::@Kwargs{})
    @ CUDA ~/.julia/packages/CUDA/x8d2s/src/compiler/execution.jl:366
  [6] cufunction(f::typeof(gpu_square_kernel!), tt::Type{Tuple{…}})
    @ CUDA ~/.julia/packages/CUDA/x8d2s/src/compiler/execution.jl:365
  [7] launch_configuration(f::ReactantCUDAExt.LLVMFunc{…}; shmem::Int64, max_threads::Int64)
    @ ReactantCUDAExt ~/.julia/packages/Reactant/zlIsO/ext/ReactantCUDAExt.jl:619
  [8] ka_with_reactant
    @ ~/.julia/packages/Reactant/zlIsO/ext/ReactantCUDAExt.jl:525 [inlined]
  [9] (::Nothing)(none::typeof(Reactant.ka_with_reactant), none::Int64, none::Nothing, none::KernelAbstractions.Kernel{…}, none::Tuple{…})
    @ Reactant ./<missing>:0
 [10] launch_config
    @ ~/.julia/packages/Reactant/zlIsO/ext/ReactantKernelAbstractionsExt.jl:68 [inlined]
 [11] ka_with_reactant
    @ ~/.julia/packages/Reactant/zlIsO/ext/ReactantCUDAExt.jl:504 [inlined]
 [12] call_with_reactant(::typeof(Reactant.ka_with_reactant), ::Int64, ::Nothing, ::KernelAbstractions.Kernel{…}, ::Reactant.TracedRArray{…}, ::Reactant.TracedRArray{…})
    @ Reactant ~/.julia/packages/Reactant/zlIsO/src/utils.jl:0
 [13] (::KernelAbstractions.Kernel{…})(::Any, ::Vararg{…}; ndrange::Int64, workgroupsize::Nothing)
    @ ReactantKernelAbstractionsExt ~/.julia/packages/Reactant/zlIsO/ext/ReactantKernelAbstractionsExt.jl:118
 [14] #kwcall
    @ ~/.julia/packages/Reactant/zlIsO/ext/ReactantKernelAbstractionsExt.jl:113 [inlined]
 [15] (::Nothing)(none::typeof(Core.kwcall), none::@NamedTuple{…}, none::KernelAbstractions.Kernel{…}, none::Tuple{…})
    @ Reactant ./<missing>:0
 [16] call_with_reactant(::typeof(Core.kwcall), ::@NamedTuple{…}, ::KernelAbstractions.Kernel{…}, ::Reactant.TracedRArray{…}, ::Reactant.TracedRArray{…})
    @ Reactant ~/.julia/packages/Reactant/zlIsO/src/utils.jl:523
 [17] square
    @ ~/Documents/GitHub/Julia/Reactant.jl/test/playground.jl:15 [inlined]
 [18] (::Nothing)(none::typeof(square), none::Reactant.TracedRArray{Float64, 1})
    @ Reactant ./<missing>:0
 [19] getproperty
    @ ./Base.jl:49 [inlined]
 [20] size
    @ ~/.julia/packages/Reactant/zlIsO/src/TracedRArray.jl:259 [inlined]
 [21] axes
    @ ./abstractarray.jl:98 [inlined]
 [22] similar
    @ ./abstractarray.jl:821 [inlined]
 [23] similar
    @ ./abstractarray.jl:820 [inlined]
 [24] square
    @ ~/Documents/GitHub/Julia/Reactant.jl/test/playground.jl:12 [inlined]
 [25] call_with_reactant(::typeof(square), ::Reactant.TracedRArray{Float64, 1})
    @ Reactant ~/.julia/packages/Reactant/zlIsO/src/utils.jl:0
 [26] make_mlir_fn(f::typeof(square), args::Tuple{…}, kwargs::@NamedTuple{}, name::String, concretein::Bool; toscalar::Bool, return_dialect::Symbol, args_in_result::Symbol, construct_function_without_args::Bool, do_transpose::Bool, input_shardings::Nothing, output_shardings::Nothing, runtime::Val{…}, verify_arg_names::Nothing, argprefix::Symbol, resprefix::Symbol, resargprefix::Symbol, num_replicas::Int64, optimize_then_pad::Bool)
    @ Reactant.TracedUtils ~/.julia/packages/Reactant/zlIsO/src/TracedUtils.jl:345
 [27] make_mlir_fn
    @ ~/.julia/packages/Reactant/zlIsO/src/TracedUtils.jl:275 [inlined]
 [28] compile_mlir!(mod::Reactant.MLIR.IR.Module, f::typeof(square), args::Tuple{…}, compile_options::CompileOptions, callcache::Dict{…}, sdycache::Dict{…}, sdygroupidcache::Tuple{…}; fn_kwargs::@NamedTuple{}, backend::String, runtime::Val{…}, legalize_stablehlo_to_mhlo::Bool, client::Reactant.XLA.PJRT.Client, kwargs::@Kwargs{})
    @ Reactant.Compiler ~/.julia/packages/Reactant/zlIsO/src/Compiler.jl:1614
 [29] compile_mlir!
    @ ~/.julia/packages/Reactant/zlIsO/src/Compiler.jl:1576 [inlined]
 [30] compile_xla(f::Function, args::Tuple{…}; before_xla_optimizations::Bool, client::Nothing, serializable::Bool, kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/.julia/packages/Reactant/zlIsO/src/Compiler.jl:3524
 [31] compile_xla
    @ ~/.julia/packages/Reactant/zlIsO/src/Compiler.jl:3496 [inlined]
 [32] compile(f::Function, args::Tuple{…}; kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/.julia/packages/Reactant/zlIsO/src/Compiler.jl:3600
 [33] top-level scope
    @ ~/.julia/packages/Reactant/zlIsO/src/Compiler.jl:2669
```

docs/src/tutorials/kernels.md
gdalle: > can you explain more what raising is. I think this doc would be better written as a tutorial on raising rather than kernels

No I can't. I don't know the first thing about how raising works (besides what @yolhan83 has taught me on Discourse), I only know that it seems to be necessary on my laptop for handwritten kernels to get compiled.

The goals of this PR are:

1. Showing that interoperability with custom kernels is an important aspect for Reactant users, that needs to be documented.
2. Getting the page started so that people who actually know what they're talking about can finish writing proper documentation.

yolhan83:
Hey all, indeed a doc about raising may be interesting but a kernel part may be as important, I think RArray auto trigger CUDABackend on KA which mean without raising it may only work on CUDA, a thing to test is when not raising, setting the backend of KA by hand and the Reactant backend too but not sure Reactant will be able to send the Metal or AMD device code to the XLA backend anyway.

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,73 @@
+# Kernels
```
Pangoraw: Just a random thought, but maybe the tutorial can be called something like "Compute Kernels" or "GPU Kernels" ? IMHO, kernel has a lot of different meanings in computing.

docs/src/tutorials/kernels.md
Pangoraw: wrote a small section about raising, if you want to include or i can open a PR after this one is merged: https://github.com/EnzymeAD/Reactant.jl/commit/5fe6e01bb52b43d7265c5ca53358a42c70ade1a8

gdalle:
So, as I understand it:

- Reactant compilation should work without `raise=true`, but on my Mac it doesn't
- Reactant compilation with `raise=true` should work without `using CUDA`, but on my Mac it doesn't

In light of this, should we wait for this to be fixed or should we document the behavior that we know for sure works across platforms?

Pangoraw:
> Reactant compilation should work without raise=true, but on my Mac it doesn't

#1923 would fix this for KA kernels, it should already work for CUDA.jl kernels.

gdalle:
Thanks! From what I see, the fix lives in Reactant's CUDA extension, so it will remain necessary to run `using CUDA` even on platforms where it is not functional?

Pangoraw:
Yes it uses CUDA.jl to extract the kernel IR. So it will need to be loaded (either by the user or the KA Reactant extension)

> we should force load this for you

@wsmoses do you know how we can achieve that?

gdalle:
By making CUDA a hard dep of Reactant? Why isn't that the case if you need it for kernels?

gdalle:
I understand the need to keep it as a weak dep though, and I don't think Pkg manipulations should be performed under the hood. So the current error message might be good enough, I was just a bit surprised that installing CUDA on my Mac would actually solve the issue (which it did).

Pangoraw:
> By making CUDA a hard dep of Reactant? Why isn't that the case if you need it for kernels?

We need for CUDA kernels and KA kernels, so it can be fair to require `using CUDA` even in the case of a non-CUDA setup.

wsmoses:
@gdalle with the other raising PR in, can you rebase this on that

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,73 @@
...
+yr = @jit raise=true square(xr)
+
+# output
+
+5-element ConcretePJRTArray{Float64,1}:
+  1.0
+  4.0
+  9.0
+ 16.0
+ 25.0
```
avik-pal: can we write these as `@example` and add asserts with `#hide` instead.

jldoctests rely on the exact printing details which are somewhat frustrating to deal with.

docs/src/tutorials/kernels.md
gdalle: Done

gdalle:
Gentle bump, this should be ready to ship

wsmoses:
I think you still need to integrate this into the existing kernel docs page, right?

wsmoses:
ah sorry that was the raising docs, let me read a fresh

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,106 @@
+# [GPU Kernels](@id gpu-kernels)
```
wsmoses: Change to computationa kernels (including GPU, TPU, and other accelerators)

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,106 @@
...
+
+```@meta
+ShareDefaultModule = true
+```
+
```
wsmoses, line 6: custom computational kernels

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,106 @@
...
+```@example
+xr = ConcreteRArray(x)
+square_compiled = @compile square(xr)
+```
+
+```@example
+yr = square_compiled(xr)
+@assert yr == xr .^ 2  # hide
+```
+
```
wsmoses, line 60: Add a comment that this is now run on whatever device is set using reactant, including CPU, TPU, and distributed (and will not run on CUDA or the default and does not matter if you do not have a cuda device)

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,106 @@
...
+## Kernel raising
+
```
wsmoses, line 63: This should just link to the existing raising docs

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,106 @@
...
+## Kernel differentiation
+
+If you want to compute derivatives of your kernel, combining Reactant with [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) is the best choice.
+
+```@example
+import Enzyme
+```
+
+You must use the `raise_first = true` compilation option to make sure the kernel is raised before Enzyme performs automatic differentiation on the program.
```
wsmoses: Currently***

This restriction will be removed in later versions

gdalle:
This seems ready

docs/src/tutorials/kernels.md
wsmoses: sorry I meant to say, why have this subsection when we have the whole raising doc page?

gdalle: So that people have the whole workflow on one page, instead of having to go to another part of the docs? I added the link to the raising page and removed the blurb suggested by Paul, but I left the code example so that people can just copy-paste it

gdalle: But if you really mind I can just remove it

gdalle:
Gentle bump, what do you think @wsmoses @avik-pal?

docs/src/tutorials/kernels.md
wsmoses: we should avoid having docs in two places to avoid them getting out of sync, or worse, out of date. 

Maybe just reference the relevant section, and for information on raising kernels see ...

gdalle: Done in the latest commit.

gdalle:
Final comments taken into account

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,88 @@
...
+square_compiled = @compile square(xr)
+```
+
+```@example
+yr = square_compiled(xr)
+@assert yr == xr .^ 2  # hide
+```
+
+The Reactant-compiled function `square_compiled` now runs on whatever device Reactant uses, including CPU, GPU, TPU or distributed settings.
+It will not run on the default device, nor will it require a CUDA-enabled device.
```
wsmoses: Keep a header for kernel raising, and just add the link to forward to the raising section? Maybe with a brief "Reactant supports the ability to detect and optimize high level tensor representations of existing kernels. For more information see ..."

docs/src/tutorials/kernels.md
```diff
@@ -0,0 +1,90 @@
...
```
wsmoses: Sorry one last comment then I think we’re good.

perhaps instead of “will not run on the default device” => will not necessarily run on the device it was written for, but instead the device requested by reactant (such as a cpu gpu TPU or other accelerator)

gdalle:
```suggestion
The Reactant-compiled function `square_compiled` now runs on whatever device Reactant requests, including CPU, GPU, TPU or distributed settings.
It will not run on the device it was written for, nor will it require a CUDA-enabled device.
```

gdalle:
All comments have been addressed

Pangoraw (approved):

gdalle:
Is there a reason why the docs page doesn't appear on the website?

Pangoraw:
> Is there a reason why the docs page doesn't appear on the website?

The tutorial needs to be added to the config.mts and tutorials/index.md. 

https://github.com/EnzymeAD/Reactant.jl/pull/1935/changes#diff-6f9e51d59e17c18f185924ee9fc6aab5bd52c56901e5ce2147356ddd8f9c6dcd

PR #1923, opened 2025-12-01
Fix KA when CUDA is non-functional & re-enable tests

Pangoraw:


wsmoses:
nice!

test/integration/kernelabstractions.jl
```diff
@@ -21,23 +21,14 @@ function matmul!(output, a)
...
-            else
-                @code_hlo optimize = :before_kernel matmul!(out, A)
-            end
-        end
-    end
+@testset "KernelAbstractions Matmul" begin
+    A = Reactant.to_rarray(ones(100, 100))
+    out = Reactant.to_rarray(ones(100, 100))
+    raise = false
+    @jit raise=raise matmul!(out, A)
```
wsmoses: this should likely still be left to the default though since it'll be needed to be true on tpu atm

Pangoraw: TPU CI fails with a raising issue whereas I can raise locally: https://github.com/EnzymeAD/Enzyme-JAX/issues/1672

So i'll disable for TPU

wsmoses: oh weird, and yeah we should clearly fix that raising (end to end this becoming a matmul again would be really nice)

Pangoraw: I managed to raise up to a matmul locally so maybe Julia version producing different ir is at play here.

https://github.com/Pangoraw/Enzyme-JAX/blob/010b8a637becd755e1e87b0f7593117d68e7739c/test/lit_tests/reduce_mul_broadcast_to_dot_general.mlir

PR #1931, opened 2025-12-04
Add support for FFT plans

### LINKED ISSUE #1926: Support FFT plans
```julia
using Reactant, FFTW

x_host = rand(ComplexF32, 16, 16)
x_r = Reactant.to_rarray(x_host)

### Non-planned, out-of-place FFT

fft_compiled = @compile fft(x_r)
y_r = fft_compiled(x_r)

# Check Reactant FFT works
@assert y_r ≈ fft(x_host)

### Planned, in-place FFT

copied_x = copy(x_host)
copied_x_r = Reactant.to_rarray(copied_x)

function planned_fft(x)
    plan = plan_fft!(x)
    return plan * x
end

planned_fft_compiled = @compile planned_fft(x_r)

# Just a sanity check the non-Reactant version works
@assert planned_fft(x_host)       ≈ fft(copied_x)
# Check Reactant FFT works
@assert compiled_planned_fft(x_r) ≈ fft(copied_x)
```
At the moment Reactant works fine with a non-planned FFT, but the planned one fails with
```
ERROR: LoadError: MethodError: no method matching plan_fft!(::Reactant.TracedRArray{ComplexF32, 2}, ::UnitRange{Int64})
The function `plan_fft!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  plan_fft!(::StridedArray{T, N}, ::Any; flags, timelimit, num_threads) where {T<:Union{ComplexF64, ComplexF32}, N}
   @ FFTW ~/.julia/packages/FFTW/uNgRy/src/fft.jl:788
  plan_fft!(::AbstractArray; kws...)
   @ AbstractFFTs ~/.julia/packages/AbstractFFTs/4iQz5/src/definitions.jl:68
  plan_fft!(::StridedArray{<:Union{ComplexF64, ComplexF32}}; kws...)
   @ FFTW ~/.julia/packages/FFTW/uNgRy/src/fft.jl:802
```

I think one way to deal with this is to extend `plan_fft!(::TracedRArray)` to return a simple struct (`<: AbstractFFTs.Plan`) which is only used for dispatch, so that `*(::AbstractReactantPlan, ::TracedRArray)` does the actual fft.  We need to do this for a variety of plans though (`fft`, `ifft`, `rfft`, etc.)

giordano:
Should fix #1926, but for some reason I don't quite understand, the in-place plans aren't in-place (the added tests should fail), even though the code in https://github.com/EnzymeAD/Reactant.jl/issues/1926#issuecomment-3609300530 works as expected.  I now get
```julia-repl
julia> using Reactant, FFTW

julia> x_host = rand(ComplexF32, 16, 16); x_r = Reactant.to_rarray(x_host);

julia> planned_fft!(x) = plan_fft!(x) * x
planned_fft! (generic function with 1 method)

julia> @code_hlo planned_fft!(x_r)
```
```mlir
module @reactant_planned... attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<16x16xcomplex<f32>> {enzymexla.memory_effects = []}) -> tensor<16x16xcomplex<f32>> attributes {enzymexla.memory_effects = []} {
    %0 = stablehlo.fft %arg0, type =  FFT, length = [16, 16] : (tensor<16x16xcomplex<f32>>) -> tensor<16x16xcomplex<f32>>
    return %0 : tensor<16x16xcomplex<f32>>
  }
}
```
Note that contrary to https://github.com/EnzymeAD/Reactant.jl/issues/1926#issuecomment-3609300530, the input is not aliasing the output.  @avik-pal do you have any clue of what's going on here?

***Edit***: I simply [missed a `!`](https://github.com/EnzymeAD/Reactant.jl/pull/1931#discussion_r2589885071) :facepalm:

### DIFF
--- a/ext/ReactantAbstractFFTsExt.jl
+++ b/ext/ReactantAbstractFFTsExt.jl
@@ -57,7 +57,7 @@ for op in (:rfft, :fft, :ifft)
         @eval AbstractFFTs.$(plan_f!)(::Reactant.TracedRArray{T}) where {T} =
             $(plan_name!){T}()
         @eval Base.:*(::$(plan_name!){T}, x::Reactant.TracedRArray{T}) where {T} =
-            x .= AbstractFFTs.$(op)(x)
+            copyto!(x, AbstractFFTs.$(op)(x))
     end
 end
 


ext/ReactantAbstractFFTsExt.jl
```diff
@@ -40,6 +40,22 @@ for op in (:rfft, :fft, :ifft)
...
+    @eval struct $(plan_name){T} <: AbstractFFTs.Plan{T} end
+    @eval AbstractFFTs.$(plan_f)(::Reactant.TracedRArray{T}) where {T} = $(plan_name){T}()
+    @eval Base.:*(::$(plan_name){T}, x::Reactant.TracedRArray{T}) where {T} = AbstractFFTs.$(op)(x)
+
+    # In-place plan
+    if op !== :rfft
+        plan_name! = Symbol("Reactant", uppercase(string(op)), "InPlacePlan")
+        plan_f! = Symbol("plan_", op, "!")
+        @eval struct $(plan_name!){T} <: AbstractFFTs.Plan{T} end
+        @eval AbstractFFTs.$(plan_f!)(::Reactant.TracedRArray{T}) where {T} = $(plan_name){T}()
```
avik-pal:
```suggestion
        @eval AbstractFFTs.$(plan_f!)(::Reactant.TracedRArray{T}) where {T} = $(plan_name!){T}()
```

giordano: Thank you, I really couldn't see it!

ext/ReactantAbstractFFTsExt.jl
```diff
@@ -40,6 +40,25 @@ for op in (:rfft, :fft, :ifft)
...
+
+    # In-place plan
+    if op !== :rfft
+        plan_name! = Symbol("Reactant", uppercase(string(op)), "InPlacePlan")
+        plan_f! = Symbol("plan_", op, "!")
+        @eval struct $(plan_name!){T} <: AbstractFFTs.Plan{T} end
+        @eval AbstractFFTs.$(plan_f!)(::Reactant.TracedRArray{T}) where {T} =
+            $(plan_name!){T}()
+        @eval Base.:*(::$(plan_name!){T}, x::Reactant.TracedRArray{T}) where {T} =
+            x .= AbstractFFTs.$(op)(x)
```
avik-pal: Does the inplace plan reuse the buffer that is input to the matrix multiply or the buffer with which the plan was created?

wsmoses: also copyto is a lot less expensive than broadcasting [for compile/trace time]

giordano: I changed it to `copyto!`, but it doesn't change much in terms of code HLO:
```julia-repl
julia> using Reactant, FFTW

julia> x_host = rand(ComplexF32, 16, 16); x_r = Reactant.to_rarray(x_host);

julia> planned_fft!(x) = plan_fft!(x) * x
planned_fft! (generic function with 1 method)

julia> @code_hlo planned_fft!(x_r)
```
```mlir
module @reactant_planned... attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<16x16xcomplex<f32>> {enzymexla.memory_effects = [], tf.aliasing_output = 0 : i32}) -> tensor<16x16xcomplex<f32>> attributes {enzymexla.memory_effects = []} {
    %0 = stablehlo.fft %arg0, type =  FFT, length = [16, 16] : (tensor<16x16xcomplex<f32>>) -> tensor<16x16xcomplex<f32>>
    return %0 : tensor<16x16xcomplex<f32>>
  }
}
```
It's exactly the same as https://github.com/EnzymeAD/Reactant.jl/issues/1926#issuecomment-3609300530

wsmoses: yeah hlo should be the same. It's just that doing a broadcast forces an extra function tracing (where we reset all arguments and redo the equivlent of to_rarray), so might as well avoid it when not necessary

giordano: > Does the inplace plan reuse the buffer that is input to the matrix multiply or the buffer with which the plan was created?

It modifies the array the plan is applied to:
```julia-repl
julia> using FFTW

julia> N = 16;

julia> x = randn(ComplexF32, N, N); copied_x = copy(x);

julia> plan = plan_fft!(x);

julia> y = randn(ComplexF32, N, N); copied_y = copy(y);

julia> plan * y;

julia> x == copied_x
true

julia> y == copied_y
false
```
The array the plan was created on is left untouched.

giordano:
I'd hope XLA is able to reuse plans applied to different arrays of the same shape.  Since all the planning is completely opaque to the frontend we can't do much about it, but reusing plans is quite a crucial optimisation (and the whole point plans exist)

wsmoses:
I'd assume so as well, but also since we already separate compile from run, during compile even if we make a different plan for all the fft instructions, they'd still all be planned during runtime ?  It already does this with matmul by trying all possible matmul plans I know

giordano:
The requisites for a plan are the shape and the eltype of the target array, and XLA should known them all at compile-time, right?  So I'd speculate it's possible it does all the planning during compilation.  But still, planning can be quite expensive, not repeating it unnecessarily would be welcome in large applications, whether it's during compile- or run-time.

wsmoses:
yeah Id assume it has a cache, but we can benchmark, check, and fix, as needed as a follow up

ext/ReactantAbstractFFTsExt.jl
```diff
@@ -40,6 +40,25 @@ for op in (:rfft, :fft, :ifft)
...
+    # Out-of-place plan
+    plan_name = Symbol("Reactant", uppercase(string(op)), "Plan")
+    plan_f = Symbol("plan_", op)
+    @eval struct $(plan_name){T} <: AbstractFFTs.Plan{T} end
+    @eval AbstractFFTs.$(plan_f)(::Reactant.TracedRArray{T}) where {T} = $(plan_name){T}()
+    @eval Base.:*(::$(plan_name){T}, x::Reactant.TracedRArray{T}) where {T} =
+        AbstractFFTs.$(op)(x)
+
+    # In-place plan
+    if op !== :rfft
```
wsmoses, line 53: whats up with rfft?

giordano: Rfft returns an array of different shape than the input one, it can't be done in-place. There's no `plan_rfft!`

wsmoses: ah got it!

PR #1948, opened 2025-12-09
Fix stricter module name resolution in Julia 1.12

### LINKED ISSUE #1943: UndefVarError: `Ops` not defined in `Reactant`
On latest released version of Reactant, julia v1.12

Since this is an undefvar error, I figured the stack trace is enough?
```julia
ERROR: UndefVarError: `Ops` not defined in `Reactant`
Suggestion: define the const at top-level before running function that uses it (stricter Julia v1.12+ rule).
Note: the binding state changed since the error occurred (was: 11, now: 0).
Stacktrace:
  [1] should_rewrite_call(ft::Any)
    @ Reactant ~/.julia/packages/Reactant/woboD/src/utils.jl:245
  [2] should_rewrite_invoke
    @ ~/.julia/packages/Reactant/woboD/src/utils.jl:280 [inlined]
  [3] call_with_reactant_generator(world::UInt64, source::Union{LineNumberNode, Method}, self::Any, redub_arguments::Any)
    @ Reactant ~/.julia/packages/Reactant/woboD/src/utils.jl:735
  [4] make_mlir_fn(f::typeof(create), args::Tuple{…}, kwargs::@NamedTuple{…}, name::String, concretein::Bool; toscalar::Bool, return_dialect::Symbol, args_in_result::Symbol, construct_function_without_args::Bool, do_transpose::Bool, input_shardings::Nothing, output_shardings::Nothing, runtime::Val{…}, verify_arg_names::Nothing, argprefix::Symbol, resprefix::Symbol, resargprefix::Symbol, num_replicas::Int64, optimize_then_pad::Bool)
    @ Reactant.TracedUtils ~/.julia/packages/Reactant/woboD/src/TracedUtils.jl:347
  [5] make_mlir_fn
    @ ~/.julia/packages/Reactant/woboD/src/TracedUtils.jl:275 [inlined]
  [6] compile_mlir!(mod::Reactant.MLIR.IR.Module, f::typeof(create), args::Tuple{…}, compile_options::CompileOptions, callcache::Dict{…}, sdycache::Dict{…}, sdygroupidcache::Tuple{…}; fn_kwargs::@NamedTuple{…}, backend::String, runtime::Val{…}, legalize_stablehlo_to_mhlo::Bool, client::Reactant.XLA.PJRT.Client, kwargs::@Kwargs{})
    @ Reactant.Compiler ~/.julia/packages/Reactant/woboD/src/Compiler.jl:1627
  [7] compile_xla(f::Function, args::Tuple{…}; before_xla_optimizations::Bool, client::Nothing, serializable::Bool, kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/.julia/packages/Reactant/woboD/src/Compiler.jl:3550
  [8] compile_xla
    @ ~/.julia/packages/Reactant/woboD/src/Compiler.jl:3522 [inlined]
  [9] compile(f::Function, args::Tuple{…}; kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/.julia/packages/Reactant/woboD/src/Compiler.jl:3626
 [10] macro expansion
    @ ~/.julia/packages/Reactant/woboD/src/Compiler.jl:2695 [inlined]
 [11] create_r(A::SMatrix{…}, B::SMatrix{…}, C::SMatrix{…}, Q::SMatrix{…}, R::SMatrix{…}, Rr::SMatrix{…}, Ts::Float64, umin::SVector{…}, umax::SVector{…}; Np::Int64)
    @ Main ~/.julia/dev/LinearMPC/example/bench_creation.jl:56
```

mofeing:
fixes #1943

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -242,9 +242,9 @@ function should_rewrite_call(@nospecialize(ft))
         if hasfield(typeof(ft), :name) && hasfield(typeof(ft.name), :module)
             mod = ft.name.module
             # Don't rewrite primitive ops, tracing utilities, or any MLIR-based functions
-            if has_ancestor(mod, Ops) ||
-                has_ancestor(mod, TracedUtils) ||
-                has_ancestor(mod, MLIR)
+            if has_ancestor(mod, Reactant.Ops) ||
+                has_ancestor(mod, Reactant.TracedUtils) ||
+                has_ancestor(mod, Reactant.MLIR)
                 return false
             end
             if string(mod) == "CUDA"


baggepinnen:
The error went away :tada: 

Still got tripped up on 
```julia
julia> allowscalar(true)
┌ Warning: It's not recommended to use allowscalar([true]) to allow scalar indexing.
│ Instead, use `allowscalar() do end` or `@allowscalar` to denote exactly which operations can use scalar operations.
└ @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:184

julia> cr = @compile create(Ar, Br, Cr, Q1r, Q2r, Q3r, Ts, umin_r, umax_r; Np=20)
ERROR: MethodError: no method matching exp!(::Reactant.TracedRArray{Float64, 2})
The function `exp!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  exp!(::StridedMatrix{T}) where T<:Union{Float32, Float64, ComplexF64, ComplexF32}
   @ LinearAlgebra ~/.julia/juliaup/julia-1.12.2+0.x64.linux.gnu/share/julia/stdlib/v1.12/LinearAlgebra/src/dense.jl:726
```
but that's another story

mofeing:
It seems like the ExplicitImport "Self Qualified Accesses" tests are failing on exactly the solution to this bug:

```
Module `Reactant` has self-qualified accesses:
  - `MLIR` was accessed as `Reactant.MLIR` inside `Reactant` at `/Users/runner/work/Reactant.jl/Reactant.jl/src/utils.jl:247:44`
  - `Ops` was accessed as `Reactant.Ops` inside `Reactant` at `/Users/runner/work/Reactant.jl/Reactant.jl/src/utils.jl:245:43`
  - `TracedUtils` was accessed as `Reactant.TracedUtils` inside `Reactant` at `/Users/runner/work/Reactant.jl/Reactant.jl/src/utils.jl:246:44`
```

Problem is that the "utils.jl" is included before the `Ops` and `TracedUtils` modules are defined, and apparently, in Julia 1.12, this is no longer valid due to some stricter module name resolution (checkout error message in #1943).

It should also be fixable by moving the `include("utils.jl")` to the end but that looks a lil bit fragile to me...

@avik-pal is it ok if we relax the "Self Qualified Accesses" test?

avik-pal:
@ericphanson would you happen to know if there is a fix to this other than disabling the check?

PR #1951, opened 2025-12-10
`similar` based on `TracedRArray` type and size

jumerckx:


src/TracedRArray.jl
```diff
@@ -266,6 +266,10 @@ function Base.similar(::TracedRArray, ::Type{T}, dims::Dims{N}) where {T,N}
     return @opcall fill(zero(unwrapped_eltype(T)), dims)
 end
 
+function Base.similar(::Type{<:TracedRArray{T}}, dims::Dims{N}) where {T,N}
+    return @opcall fill(zero(T), dims)
```
wsmoses, line 270: @jumerckx there's also a version that is ArrayType, ElType, dims as well

jumerckx: Ah, will add tomorrow

jumerckx: > ArrayType, ElType, dims

I don't see that one in the docs or method tables
```
[3] similar(::Type{T}, dims::NTuple{N, Int64} where N) where T<:AbstractArray
   @ abstractarray.jl:867
[4] similar(::Type{T}, shape::Tuple{Union{Integer, Base.OneTo}, Vararg{Union{Integer, Base.OneTo}}}) where T<:AbstractArray
   @ abstractarray.jl:866
[5] similar(::Type{T}, dims::Union{Integer, AbstractUnitRange}...) where T<:AbstractArray
   @ abstractarray.jl:865
```

PR #1953, opened 2025-12-10
Update Reactant_jll version to 0.0.269

wsmoses:


avik-pal (approved):

avik-pal:
this will probably break cuda dot general

avik-pal:
we need the ffi from https://github.com/EnzymeAD/Reactant.jl/pull/1944

wsmoses:
ah fair enough. I'll let you handle the rest of the jll bump then

Issue #1954, opened 2025-12-10
Using Global ConcreteRArrays rather than arguments during compilation


lassepe:
I've been exploring some of the interop between Reactant and JAX and ran into the following error:

```julia
import Reactant as Rx
import Flux as Fx
import CUDA as Cu
import Preferences as Prefs
import PythonCall as Py

const jax = Py.pyimport("jax")
const jnp = Py.pyimport("jax.numpy")

const jax = Py.pyimport("jax")
const jnp = Py.pyimport("jax.numpy")

Rx.set_default_backend("gpu")

batchsize = 128
input_dim = 2
hidden_dim = 128
output_dim = 2
# build a simple model in Flux
model_fx = Fx.Chain(
  Fx.Dense(input_dim, hidden_dim, Fx.relu),
  Fx.Dense(hidden_dim, hidden_dim, Fx.relu),
  Fx.Dense(hidden_dim, hidden_dim, Fx.relu),
  Fx.Dense(hidden_dim, hidden_dim, Fx.relu),
  Fx.Dense(hidden_dim, output_dim)
) |> Fx.gpu
x_fx = Cu.randn(Float32, input_dim, batchsize)

# convert the model to Reactant, do not compile just yet
model_rx = Rx.to_rarray(model_fx)
x_rx = Rx.to_rarray(x_fx)

result_rx_only = Rx.@jit model_rx(x_rx) # compiling only the julia model works
result_jax_only = Rx.@jit jnp.sum(x_rx) # compiling JAX code with a julia input works
result_rx_jax_mixed = Rx.@jit jnp.sum(model_rx(x_rx)) # compiling both jointly throws `ERROR: conversion to pointer not defined for Reactant.ConcretePJRTArray{Float32, 2, 1}`
```

Stacktrace:

```julia


ERROR: conversion to pointer not defined for Reactant.ConcretePJRTArray{Float32, 2, 1}
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] unsafe_convert(::Type{Ptr{Float32}}, a::Reactant.ConcretePJRTArray{Float32, 2, 1})
    @ Base ./pointer.jl:66
  [3] gemm!(transA::Char, transB::Char, alpha::Float32, A::Reactant.ConcretePJRTArray{…}, B::Reactant.ConcretePJRTArray{…}, beta::Float32, C::Reactant.ConcretePJRTArray{…})
[...]

PR #1955, opened 2025-12-10
Clarify behavior of `@jit foo(bar(x))`

lassepe:
xref #1954

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -2594,6 +2594,20 @@ end
 
 Compile the function `f` with arguments `args` and return the compiled function.
 
+## Note
+
+Note that `@compile foo(bar(x))` is equivalent to
+```julia
+y = bar(x)  # first compute the output of `bar(x)`, say `y`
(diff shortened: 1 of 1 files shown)


src/Compiler.jl
```diff
@@ -2617,6 +2617,20 @@ end
...
+Note that `@jit foo(bar(x))` is equivalent to
+```julia
+y = bar(x)  # first compute the output of `bar(x)`, say `y`
+@jit foo(y) # then compile `foo` for `y` and execute it
+```
+That is, like `@compile`, `@jit` only applies to the outermost function call; it does *not* compile the composed function `foo(bar(x))` jointly.
+Hence, if you want to compile the composed function `foo(bar(x))` jointly, you need to introduce an intermediate function, i.e.,
+```julia
+baz(x) = foo(bar(x))
+@jit baz(x)
```
avik-pal: we should do this for `@compile` as well

lassepe: done

wsmoses (approved):

PR #1967, opened 2025-12-12
Comment out Metal's `__init__`

giordano:
The init function is run unconditionally, and this was causing the pluging to be always downloaded, even if it's not going to be used.

### DIFF
--- a/src/accelerators/Metal.jl
+++ b/src/accelerators/Metal.jl
@@ -7,11 +7,11 @@ using p7zip_jll: p7zip
 
 const metal_pjrt_plugin_dir = Ref{Union{Nothing,String}}(nothing)
 
-function __init__()
-    @static if Sys.isapple()
-        Reactant.precompiling() || setup_metal_pjrt_plugin!()
-    end
-end
+# function __init__()
+#     @static if Sys.isapple()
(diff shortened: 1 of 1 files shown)


src/accelerators/Metal.jl
```diff
@@ -7,11 +7,11 @@ using p7zip_jll: p7zip
...
 const metal_pjrt_plugin_dir = Ref{Union{Nothing,String}}(nothing)
 
-function __init__()
-    @static if Sys.isapple()
-        Reactant.precompiling() || setup_metal_pjrt_plugin!()
-    end
-end
+# function __init__()
+#     @static if Sys.isapple()
+#         Reactant.precompiling() || setup_metal_pjrt_plugin!()
```
wsmoses, line 12: do we ever actually set it up (that said I'm not sure its usable rn anyways)

giordano: No (Avik posted this on Slack): https://github.com/EnzymeAD/Reactant.jl/blob/d7fbda54b4350f731eb48c0aa277d6954d157625/src/xla/XLA.jl#L283-L296

wsmoses: ack, yeah can we keep a todo issue for commenting these both back in when apple releases a newer pjrt plugin

PR #1973, opened 2025-12-14
Add example for AMD GPUs in documentation

wsmoses:


docs/src/index.md
```diff
@@ -76,6 +76,11 @@ using Reactant
 Reactant.set_default_backend("gpu")
 ```
 
+```julia [AMD GPUs]
```
giordano, line 82: How about putting this together with Nvidia GPUs for the time being? The code to execute is just the same.

wsmoses: yeah so one thing I'm debating doing here, is we could have a commented out line set_default_backend("cuda") vs "rocm" which will enforce which one

giordano: I like being explicit, it reduces confusion.

Maybe should keep "gpu" as an alias of "cuda" for the rest of v0.2 though?

wsmoses: So the current setup at least has gpu essentially as an Alia's for whatever GPU was found on the system, including when metal earlier worked.

I think it's nice to have both so you can either just say get me a GPU or alternatively make sure you get this specifically

wsmoses: @giordano what do you think of this?

giordano: Looks better (although the code commented out is easy to ignore/miss)

PR #1985, opened 2025-12-16
fix: enable batching except slice to batch passes

avik-pal:


avik-pal:
tested upstream that this doesn't regress GB25

PR #1986, opened 2025-12-16
Add `tessera` attribute

jumerckx:
```jl
@trace tessera function reciprocal(x)
	return 1/x
end
```
This will add an attribute with name `tessera_name` and value a stringattr with the function name (`"reciprocal"`) to the generated MLIR function. (note that you need `@code_hlo optimize=false` to see this attribute, otherwise it is inlined.)
```
@code_hlo optimize=false reciprocal.(x)
```
generates code containing:
```mlir
  func.func private @"####reciprocalinternal#267#287"(%arg0: tensor<f64> {enzymexla.memory_effects = []}) -> (tensor<f64>, tensor<f64>) attributes {enzymexla.memory_effects = [], tessera_name = "reciprocal"} {
    %0 = stablehlo.convert %c : (tensor<i64>) -> tensor<f64>
    %1 = stablehlo.divide %0, %arg0 : tensor<f64>
    return %1, %arg0 : tensor<f64>, tensor<f64>
  }
```

(note the `tessera_name` attribute)

---

It might be nicer to have a dedicated macro, i.e.:
```jl
@tessera function reciprocal(x)
	1/x
end
```
I just did it this way because it was quicker to reuse some stuff.


jessicacotturone21:
```
@tessera_op "reciprocal" function reciprocal(x)
	return 1/x
end
```

Now a custom `@tessera_op` macro can also be used with a string like "reciprocal" to add a tessera_op attribute to a function declaration.

PR #1994, opened 2025-12-19
ci: use dependabot instead of compat helper

avik-pal:


.github/dependabot.yml
```diff
@@ -1,7 +1,18 @@
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 version: 2
+enable-beta-ecosystems: true
```
giordano: This shouldn't be necessary anymore.

.github/dependabot.yml
```diff
@@ -1,7 +1,18 @@
...
 updates:
   - package-ecosystem: "github-actions"
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "julia"
+    directories:
+      - "/"
```
giordano, line 11: If you use workspaces for docs, tests, benchmarks, like in https://github.com/NumericalEarth/Breeze.jl/pull/208, you can keep only root and lib/reactantcore, the benefit being that you have a single PR for all workspaces.

giordano (approved):

PR #1896, opened 2025-11-21
feat: better lowering for Base.Slices

avik-pal:
fixes #1892

PR #2001, opened 2025-12-21
feat: new jll + new passes

avik-pal:


avik-pal:
same failure as https://github.com/EnzymeAD/Reactant.jl/pull/1896 and only on the TPU machines. Very strange given nothing is being changed here atleast on the julia end

PR #2011, opened 2025-12-23
Update MPI support

romanlee:
Previously, MPI commands were lowered directly to LLVM inside Reactant. In this PR, we instead first lower to a set of MPI Ops defined in Enzyme-JAX by calling the generated MLIR bindings, then invoke the lowering passes (also defined in Enzyme-JAX) to lower from the EnzymeXLA Ops to LLVM.

Currently supported MPI commands are 

* MPI_Comm_rank

* MPI_Comm_size
* MPI_Barrier

* MPI_Send
* MPI_Recv
* MPI_Isend
* MPI_Irecv
* MPI_Wait
* MPI_Allreduce

Note: Currently only MPI_COMM_WORLD is the only communicator supported


ext/ReactantMPIExt/Ops.jl
```diff
@@ -8,500 +8,133 @@ using Reactant.Ops: mlir_stacktrace, mlir_type
...
-            location,
-            output_operand_aliases,
-        ),
-    )
-    return TracedRNumber{Cint}((), res)
+@noinline function comm_rank(;
+    location=mlir_stacktrace("mpi.comm_rank", @__FILE__, @__LINE__)
+)
+    rank = Reactant.Ops.constant(Int32(-1))
+    ret = enzymexla.mpi_comm_rank(; rank=mlir_type(rank), location)
```
wsmoses: Can we avoid constructing the constant and just use mlir_type?

romanlee: How about that? da5d8d97bdbbb9c3c58c1df50ef42ad7a7c44b36

ext/ReactantMPIExt/Ops.jl
```diff
@@ -8,647 +8,276 @@ using Reactant.Ops: mlir_stacktrace, mlir_type
...
-        IR.Value[buf.mlir_data, count.mlir_data, dest.mlir_data, tag.mlir_data];
-        fn=sym_attr,
-        result_0=IR.Type[],
-        output_operand_aliases=IR.Attribute(IR.Attribute[]),
+    enzymexla.mpi_send(
+        buf.mlir_data,
+        count.mlir_data,
+        dest.mlir_data,
+        tag.mlir_data;
+        datatype=MLIR.API.enzymexlaMPIDatatypeAttrGet(IR.context(), mpi_datatype),
```
romanlee, line 58: Is this an ok way to pass in the datatype?

PR #2036, opened 2025-12-27
feat: builtin profiler in the terminal

avik-pal:
```julia
julia> Reactant.Profiler.@profile linear(x, W, b)
```

image


wsmoses (approved):
amazing, we should add something about to the profile docs tho

src/Profiler.jl
```diff
@@ -865,6 +866,405 @@ macro time(args...)
...
+
+            $(_print_summary_header)("SUMMARY")
+            println()
+            println(profile_result.profiling_result)
+            println()
+
+            profile_result.val
+        end,
+    )
+end
```
wsmoses, line 1281: later on we should also add a utility to parse and generate a report from an existing xplane profile, for sake of generating plots from existing profiles

wsmoses: cc @JBlaschke

avik-pal: There's also a tool for getting the roofline model data. I remember discussing about that during GB

wsmoses (approved):

docs/src/tutorials/profiling.md
```diff
@@ -31,6 +32,14 @@ Reactant.@timed nrepeat=100 linear(x, W, b)
 Note that the information returned depends on the backend. Specifically CUDA and TPU
 backends provide more detailed information regarding memory usage and allocation.
 
+Additionally for GPUs and TPUs, we can use the [`Reactant.@profile`](@ref) macro to profile
+the function and get information regarding each of the kernels executed.
+
+```@example profiling
+Reactant.@profile linear(x, W, b)
+nothing # hide
+```
```
wsmoses, line 67: does this show the nice output [like in the screenshot]. If not we should show that somewhere in the docs. The output is so useful we should show it off!

wsmoses: ideally we should also write about exporting and importing profiles too

avik-pal:
https://enzymead.github.io/Reactant.jl/previews/PR2036/tutorials/profiling

PR #2053, opened 2026-01-01
feat: add a option for structured tensor optimization passes

avik-pal:


wsmoses (approved):
not urgent, but in the future we should consider renaming the pass patterns to be positive, e.g. disable_X -> X or enable_X for clarity

Issue #594, opened 2025-01-21
Precompile fails "LoadError: UndefVarError: `libReactantExtra` not defined in `Reactant_jll`"


macquarrielucas:
I'm new to Julia so forgive me if this is a user issue, but I don't believe it is.

Reactant can't be precompiled due to libReactantExtra not defined in Reactant_jll. Below is the stack trace and my version. I don't have another computer to test this on, can anyone reproduce it?
Version info:
```
julia> versioninfo()
Julia Version 1.11.2
Commit 5e9a32e7af (2024-12-01 20:02 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 16 × 12th Gen Intel(R) Core(TM) i5-12500H
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, alderlake)
Threads: 1 default, 0 interactive, 1 GC (on 16 virtual cores)
```

Stack trace: 
```
(LuxTutOne) pkg> activate --temp
  Activating new project at `C:\Users\macqu\AppData\Local\Temp\jl_l5msT0`

(jl_l5msT0) pkg> add Reactant
   Resolving package versions...
    Updating `C:\Users\macqu\AppData\Local\Temp\jl_l5msT0\Project.toml`
  [3c362404] + Reactant v0.2.21
    Updating `C:\Users\macqu\AppData\Local\Temp\jl_l5msT0\Manifest.toml`
  [79e6a3ab] + Adapt v4.1.1
  [fa961155] + CEnum v0.5.0
  [34da2185] + Compat v4.16.0
  [187b0558] + ConstructionBase v1.5.8
  [7da242da] + Enzyme v0.13.28
  [f151be2c] + EnzymeCore v0.8.8
  [e2ba6199] + ExprTools v0.1.10
  [21656369] + ExpressionExplorer v1.1.1
  [d9f16b24] + Functors v0.5.2
  [46192b85] + GPUArraysCore v0.2.0
  [61eb1bfa] + GPUCompiler v1.1.0
  [692b3bcd] + JLLWrappers v1.7.0
  [929cbde3] + LLVM v9.1.3
  [1914dd2f] + MacroTools v0.5.15
  [d8793406] + ObjectFile v0.4.3
  [bac558e1] + OrderedCollections v1.7.0
  [aea7be01] + PrecompileTools v1.2.1
[...]

PR #2063, opened 2026-01-04
ProbProg: MCMC + Pure StableHLO Migration

sbrantq:


sbrantq:
Took back the probprog benchmark CI from this PR since there is still a few minor stuff to sort out (will do that in a separate PR)

Project.toml
```diff
@@ -20,6 +20,7 @@ LLVM = "929cbde3-209d-540e-8aea-75f648917ca0"
 LLVMOpenMP_jll = "1d63c593-3942-5779-bab2-d838dc0a180e"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MCMCDiagnosticTools = "be115224-59cd-429b-ad48-344e309966f0"
```
wsmoses: What is this dep? Can it be avoided or an extension

sbrantq: I use it to compute ESS and R-hat (which is kind of complex to write our own). Yeah i just made it an extension

src/probprog/FFI.jl
```diff
@@ -0,0 +1,830 @@
...
+    if N <= 1
+        return copy(unsafe_wrap(Array, ptr, shape))
+    end
+    reversed_shape = reverse(shape)
+    transposed = unsafe_wrap(Array, ptr, reversed_shape)
+    return permutedims(transposed, N:-1:1)
+end
+
+function to_row_major!(
+    ptr::Ptr{T}, src::AbstractArray{T,N}, shape::NTuple{N,<:Integer}
```
wsmoses: @gbaraldi can you take a look at the ffi related bits here


@sbrantq can you explain the design and potential safety or not and whats happening.

Also of relevance, what happens if the device is _not_ a cpu

avik-pal:
@sbrantq the JLL side changes are in https://github.com/EnzymeAD/Reactant.jl/pull/2284, though it is currently blocked by windows failures

sbrantq:
> @sbrantq the JLL side changes are in #2284, though it is currently blocked by windows failures

I see, thanks!

src/probprog/FFI.jl
sbrantq: Yeah, basically this piece of code constructs C function pointers that enable the compiled thunk to call back into Julia land through `enzymexla::JITCallOp` to read/write Julia data structures (mostly Julia dicts in our case) by passing tensor and Julia pointers using XLA buffers (hence the row/column-major conversions).

Regarding safety we have `_keepalive!` to prevent GC on `ProbProgTrace`s (adding more assertion shortly). And there is type assertions and `try` statements to do runtime checks. (I will add more coverage shortly)

So far it assumes CPU-accessible memory; the use of a GPU backend is unsupported and will cause compile time error in Enzyme-XLA (e.g., guarded by https://github.com/EnzymeAD/Enzyme-JAX/blob/a234f7345ce7559099c2818265cc934e8432c0d1/src/enzyme_ad/jax/Passes/LowerEnzymeProbProg.cpp#L296). To enable FFI.jl on GPU we will implement proper device-to-host memcpy in Enzyme-JAX lowering or switch to pure StableHLO trace representation.

sbrantq:
Thanks @avik-pal ! (I'm doing a big refactor to migrate to a pure stablehlo trace implementation which will eliminate FFI.jl here)

sbrantq:
My apologies, due to review capacity atm poking more folks to review the high level stuff (probprog tests are failing as they depend on https://github.com/EnzymeAD/Enzyme/pull/2692 https://github.com/EnzymeAD/Enzyme-JAX/pull/2030)

ext/ReactantMCMCDiagnosticToolsExt.jl
```diff
@@ -0,0 +1,26 @@
+module ReactantMCMCDiagnosticToolsExt
+
+using Reactant.ProbProg: ProbProg
```
wsmoses: I thought we got rid of the need for this? maybe not?

wsmoses:
```
Run actions/upload-artifact@v6
With the provided path, there will be 32 files uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

ext/ReactantMCMCDiagnosticToolsExt.jl
sbrantq: Yeah it's still needed if a user wants advanced mcmc diagnostics, though it's completely optional

src/probprog/FFI.jl
```diff
@@ -0,0 +1,130 @@
...
+function __init__()
+    dump_ptr = @cfunction(
+        dump,
+        Cvoid,
+        (Ptr{Any}, Ptr{UInt8}, Ptr{UInt64}, Ptr{UInt64}, Ptr{UInt64}, Ptr{UInt64})
+    )
+    @ccall MLIR.API.mlir_c.EnzymeJaXMapSymbol(
+        :enzyme_probprog_dump::Cstring, dump_ptr::Ptr{Cvoid}
+    )::Cvoid
+
```
giordano, line 128: This break loading Reactant on platforms where the artifact isn't available.  This has been a problem before, and it broke testing downstream packages for platforms where Reactant can't (legitimately) be loaded.

giordano: Similar past problems: #594.

The package must be loadable even if the artifact isn't available.

sbrantq: I think we can just get rid of it entirely (it's neither required nor effective atm) given that

giordano: That'd be good (or whatever fixes the problem, really)

PR #2064, opened 2026-01-04
chore: bump jll and new passes

avik-pal:


avik-pal:
the only remaining slowdowns for syrk and covariance https://github.com/EnzymeAD/Enzyme-JAX/issues/1877

PR #2089, opened 2026-01-08
feat: new passes

avik-pal:


wsmoses (commented):
weren't there also a bunch of cse and licm ones as well?

PR #2093, opened 2026-01-08
redesign MLIR.IR

### LINKED ISSUE #319: Allocations and time on `compile` are ever-increasing
Recently I found this strange behavior

```julia
julia> @btime Reactant.with_debug() do
           @jit Ops.abs(x)
       end
  37.419 ms (147556 allocations: 6.07 MiB)
2-element ConcreteRArray{Int64, 1}:
 1
 1

julia> @btime @jit Ops.abs(x)
  58.990 ms (297529 allocations: 12.08 MiB)
2-element ConcreteRArray{Int64, 1}:
 1
 1

julia> @btime @jit Ops.abs(x)
  74.636 ms (411819 allocations: 16.63 MiB)
2-element ConcreteRArray{Int64, 1}:
 1
 1

julia> @btime Reactant.with_debug() do
           @jit Ops.abs(x)
       end
  86.295 ms (505732 allocations: 20.37 MiB)
2-element ConcreteRArray{Int64, 1}:
 1
 1

julia> @btime Reactant.with_debug() do
           @jit Ops.abs(x)
       end
  97.753 ms (586540 allocations: 23.59 MiB)
2-element ConcreteRArray{Int64, 1}:
 1
 1

julia> @btime Reactant.with_debug() do
           @compile Ops.abs(x)
       end
  100.258 ms (652550 allocations: 26.15 MiB)

ulia> @btime @compile Ops.abs(x)
  111.832 ms (721709 allocations: 28.93 MiB)
Reactant.Compiler.Thunk{Symbol("##abs_reactant#83526")}()
```

Why are allocations (and time, but this looks like a consequence) ever increasing on each call to `compile`? For sure we are not freeing the previously compiled functions who live in the `Thunk`s but that shouldn't affect?

### LINKED ISSUE #2088: Redesign MLIR IR bindings
On #319, we noticed that there are MLIR objects getting leaked, and while trying to fix them in #2048, some data races on garbage collection surfaced.

It seems clear now that we need to change the design of the `MLIR.IR` module to account for the ownership friction between the C++ side and the Julia/GC side.

## Proposal

In the new design, ~there should be only two objects that are fully owned by Julia (i.e. lifetimes are managed by Julia's GC):~ nothing is truly fully owned by Julia (i.e. the GC finalizers never call C-API free functions).

- ~`Context` who takes care of immutable variables.~
- ~`Module` who takes care of operations.~

As the order in which the GC schedules the finalizers is non-deterministic (which produces use-after-free errors), relying on GC for resource management is problematic, leading to the problems seen in https://github.com/EnzymeAD/Reactant.jl/pull/2048#issuecomment-3700929728 and https://github.com/EnzymeAD/Reactant.jl/pull/2048#issuecomment-3711056630. Furthermore, context managers in Julia (i.e. do-block syntax) have performance problems (see https://github.com/JuliaLLVM/LLVM.jl/pull/309).

After experimenting with some API designs, I think the best solution is too try to use `dispose!` as LLVM.jl and go for a similar API design.

### Changes

- Just like LLVM.jl, we should manually `dispose!` objects.
  - MLIR.IR types are no longer mutable.
  - We could import the `@dispose` macro from LLVM.jl as it come handy.
- ~`Context` and `Module` should be fully owned by Julia. The rest of objects should avoid Julia-ownership since they are refs to objects that will be deallocated by a `Context` (i.e. attributes and types) or a `Module` (i.e. operations, blocks and regions).~
- ~An `Operation` should no longer be owned, and thus, our bindings must make sure that an `Operation` is always created as part of a `Block`.~
- ~There should be a scoped `Block` stack (just like there is a `Context` stack), with the latest one being the target of newly created operations. First element of the stack should be the body of the top `Module`.~
- We should try that by default, `create_operation_common` creates the operation in the current `Block`-scope. This way we avoid dangling ops (and encourage to avoid creating dangling operations!).

### Questions

- ~Modules are just ops in MLIR, and a module inside a module doesn't need to be owned on the Julia side... so prob we could have a non-owned `Module` type and a `TopModule` owned type.~

mofeing:
closes #2088
fixes #319 

## changes

- [x] fix type instability on global state functions (i.e. `activate!`, `deactivate!`, `mmodule`, `_has_context`...)
- [x] remove mutability and move to a manual `dispose!` based design
- [x] replace `mlirIsNull` assert on type constructors for `LLVM.@checked` and `refcheck`
- [x] rename `context!`, `block!`, `mmodule!` to `with_context`, `with_block`, `with_module` and deprecate them for `@scope` macro
  - [ ] look for a better name for `@scope`
- [x] rename `block()` to `current_block`, `mmodule()` to `current_module`
- [x] give `Region` a global stack
- [x] rename `type!`, `attr!`, `attr` to `settype!`, `setattr!`, `getattr`
- [x] remove `BlockIterator`, `RegionIterator`, `OperationIterator` and move code directly to `Region`, `Operation` and `Block`
- [x] update code
- [x] add debugging capability to disposals like in LLVM.jl (i.e. `memcheck`)
  - [ ] add sth similar to `lose_ownership!` but for memcheck
- [x] minor formatting
- [x] fix segfault on `show(::Operation)`
- [x] make one `Context` persistent all along the Julia session
- [x] fix CUDA extension
  - @wsmoses this required to make `GPUCompiler.compile` to return a MLIR `Module` (or in string repr) and move the `LinkInModule` outside of it... if not, the next time the CUDA kernel is launched, the LLVM IR is not linked to current module

### DIFF
--- a/src/ConcreteRArray.jl
+++ b/src/ConcreteRArray.jl
@@ -281,7 +281,12 @@ function Base.show(io::IO, X::Union{ConcretePJRTScalar,ConcreteIFRTScalar})
         return nothing
     end
     print(io, "$(typeof(X))(")
-    show(io, to_number(X))
+    MLIR.IR.@dispose ctx = MLIR.IR.Context(registry[]) begin
+        MLIR.IR.register_enzymexla_dialects(ctx)
+        MLIR.IR.@scope ctx begin
+            show(io, to_number(X))
+        end
+    end
     print(io, ")")
     return nothing
 end
@@ -291,7 +296,12 @@ function Base.print_array(io::IO, X::Union{AnyConcretePJRTArray,AnyConcreteIFRTA
         print(io, "<Empty Buffer eltype $(eltype(X)) of size $(size(X))>")
         return nothing
     end
-    return Base.print_array(io, convert(Array, X))
+    MLIR.IR.@dispose ctx = MLIR.IR.Context(registry[]) begin
+        MLIR.IR.register_enzymexla_dialects(ctx)
+        MLIR.IR.@scope ctx begin
+            Base.print_array(io, convert(Array, X))
+        end
+    end
 end
 
 function Base.showarg(


mofeing:
ultimately I had to set a persistent MLIR Context in order for `Module` and `Operation` to survive between `compile` calls. `@code_hlo` wouldn't work otherwise. luckily writing MLIR is now both safer and easier, as in general you don't need to create and dispose the MLIR `Context`. there is still a stack; it's just that it doesn't start empty.

there is a 15% reduction in allocations here, most prob due to some type-inference fixes on `activate!`, `deactivate!`, `_has_*`, ...

I've imported `@dispose` from LLVM, but I've left activation/deactivation of global state explicit. for that, there is now a `@scope` macro that replaces `context!`/`block!`/`mmodule!` (which used closures, check JuliaLLVM/LLVM.jl#309). another name would be preferred but I'm out of inspiration.

### before

```julia
julia> for _ in 1:20
           @btime @compile Reactant.Ops.abs(x)
       end
  42.816 ms (2039 allocations: 1.09 MiB)
  44.508 ms (2039 allocations: 1.09 MiB)
  43.203 ms (2039 allocations: 1.09 MiB)
  48.257 ms (2039 allocations: 1.09 MiB)
  47.820 ms (2039 allocations: 1.09 MiB)
  48.241 ms (2039 allocations: 1.09 MiB)
  53.433 ms (2039 allocations: 1.09 MiB)
  51.344 ms (2039 allocations: 1.09 MiB)
  52.365 ms (2039 allocations: 1.09 MiB)
  51.383 ms (2039 allocations: 1.09 MiB)
  52.029 ms (2039 allocations: 1.09 MiB)
  52.395 ms (2039 allocations: 1.09 MiB)
  53.260 ms (2039 allocations: 1.09 MiB)
  52.293 ms (2039 allocations: 1.09 MiB)
  62.027 ms (2039 allocations: 1.09 MiB)
  61.427 ms (2039 allocations: 1.09 MiB)
  66.278 ms (2039 allocations: 1.09 MiB)
  68.043 ms (2039 allocations: 1.09 MiB)
  68.822 ms (2039 allocations: 1.09 MiB)
  69.772 ms (2039 allocations: 1.09 MiB)
```

### after

```julia
julia> for _ in 1:20
           @btime @compile Reactant.Ops.abs(x)
       end
  28.362 ms (1715 allocations: 753.44 KiB)
  31.502 ms (1715 allocations: 753.44 KiB)
  28.297 ms (1715 allocations: 753.44 KiB)
  28.210 ms (1715 allocations: 753.44 KiB)
  28.790 ms (1715 allocations: 753.44 KiB)
  27.518 ms (1715 allocations: 753.44 KiB)
  28.637 ms (1715 allocations: 753.44 KiB)
  28.576 ms (1715 allocations: 753.44 KiB)
  28.750 ms (1715 allocations: 753.44 KiB)
  31.689 ms (1715 allocations: 753.44 KiB)
  29.254 ms (1715 allocations: 753.44 KiB)
  32.592 ms (1715 allocations: 753.44 KiB)
  32.450 ms (1715 allocations: 753.44 KiB)
  32.218 ms (1715 allocations: 753.44 KiB)
  28.699 ms (1715 allocations: 753.44 KiB)
  29.658 ms (1715 allocations: 753.44 KiB)
  28.815 ms (1715 allocations: 753.44 KiB)
  28.250 ms (1715 allocations: 753.44 KiB)
  29.735 ms (1715 allocations: 753.44 KiB)
  28.597 ms (1715 allocations: 753.44 KiB)
```

mofeing:
now that #2122 has been merged, this PR should be ready to merge if tests pass

edit: funk

mofeing:
this PR will be split up into smaller PRs

Issue #2078, opened 2026-01-06
fft missing methods


ptiede:
Reactant is missing a few use cases where an FFT is required, but the current extension is missing methods. Here are a few cases I have encountered.

### Plan is defined based on cpu arguments and is essentially constant

```julia
using Reactant, FFTW

plan = plan_fft!(copy(x))

plan * x #works in place

xr = Reactant.to_rarray(x)

@compile p * xr
# Gives execution error because it is trying to use fftw methods
# Missing Base.:*(plan, x::RArray) method
```

### Inplace ffts are not defined

```julia
using Reactant, FFTW

x = rand(ComplexF64, 64, 64)
@compile fft!(x)
# Missing fft!(x::RArray) 
```

### ReactantFFT plans doesn't support dims

```julia
x = rand(ComplexF64, 64, 64, 4)
xr = Reactant.to_rarray(xr)
@compile plan_fft(xr, 1:2)
# ERROR: MethodError: no method matching plan_fft(::Reactant.TracedRArray{ComplexF64, 2}, ::UnitRange{Int64})
```

I noticed this while reviewing the new code from #1926. This should be easy enough to fix by changing the struct to capture the dims argument, or otherwise make it `1:ndims(x)`

The second and third ones are easy to fix. For the first one, my naive guess would be to define the overload and use Reactant to compute the fft.

wsmoses:
in place FFT's are also easy, we just should define it as copyto!(x, fft(x))

want to give these a go?

ptiede:
Yep!

PR #2095, opened 2026-01-09
Add additional FFT and FFTW support 

ptiede:
This PR aims to add support for the issues in #2078. I also added the ability to do `plan_irfft` since this is covered in AbstractFFTs but wasn't added in #1926. 

My approach to adding support for multiplying FFT plans by RArrays, is not "smart" so people should review it and make sure the approach I took isn't too dumb. I thought about using `make_tracer` but that seemed complicated, and is rather annoying since AbstractFFTs and FFTW live in separate extensions.

### DIFF
--- a/ext/ReactantAbstractFFTsExt.jl
+++ b/ext/ReactantAbstractFFTsExt.jl
@@ -179,21 +179,11 @@ function normbfft(::Type{T}, size, dims) where {T}
     return inv(AbstractFFTs.normalization(real(T), size, dims))
 end
 
-function AbstractFFTs.bfft(x::AnyTracedRArray{T}, dims) where {T}
-    y = AbstractFFTs.ifft(x, dims)
-    y .*= normbfft(real(T), size(x), dims)
-    return y
-end
-
+# Because we override the plan_bfft and plan_brfft functions we actually do not need to define
+# AbstractFFTs.bfft functions since they come for free via the plan mechanism.
 function AbstractFFTs.plan_bfft(x::Reactant.TracedRArray{T}, dims=1:ndims(x)) where {T}
     pl = AbstractFFTs.plan_ifft(x, dims)
     return normbfft(real(T), size(x), dims) * pl # ScaledPlan
-end
-
-function AbstractFFTs.bfft!(x::AnyTracedRArray{T}, dims) where {T}
-    AbstractFFTs.ifft!(x, dims)
-    x .*= normbfft(real(T), size(x), dims)
-    return x
 end
 
 function AbstractFFTs.plan_bfft!(x::Reactant.TracedRArray{T}, dims=1:ndims(x)) where {T}
@@ -204,12 +194,6 @@ end
 # This must be implemented for bfft
 function reallength end
 
-function AbstractFFTs.brfft(x::AnyTracedRArray{T}, dims) where {T}
-    y = AbstractFFTs.$(iop)(x, dims)
-    y .*= normbfft(real(T), size(y), dims)
-    return y
-end
-
 function AbstractFFTs.plan_brfft(
     x::Reactant.TracedRArray{T}, length::Integer, dims=1:ndims(x)
 ) where {T}
@@ -218,14 +202,14 @@ function AbstractFFTs.plan_brfft(
     return normbfft(real(T), sz, dims) * y # ScaledPlan
 end
 
-function LinearAlgebra.mul!(
-    y::Reactant.TracedRArray,
-    p::AbstractFFTs.ScaledPlan{<:AbstractReactantFFTPlan},
-    x::Reactant.TracedRArray,
-)
-    mul!(y, p.p, x)
-    y .*= p.scale
-    return y
-end
+# function LinearAlgebra.mul!(
+#     y::Reactant.TracedRArray,
+#     p::AbstractFFTs.ScaledPlan{<:AbstractReactantFFTPlan},
+#     x::Reactant.TracedRArray,
+# )
+#     mul!(y, p.p, x)
+#     y .*= p.scale
+#     return y
+# end
 
 end
--- a/test/integration/fft.jl
+++ b/test/integration/fft.jl
@@ -20,6 +20,7 @@ using FFTW, Reactant, Test, LinearAlgebra
 
     y_ra = @jit(fft(x_ra))
     @test @jit(ifft(y_ra)) ≈ x
+    @test @jit(bfft(y_ra)) ≈ x * length(x)
 
     shifted_fft = @jit(fftshift(y_ra))
     @test shifted_fft ≈ fftshift(Array(y_ra))
@@ -32,6 +33,8 @@ using FFTW, Reactant, Test, LinearAlgebra
         @test @jit(fft(x_ra)) ≈ fft(x)
         @test @jit(fft(x_ra, (1, 2))) ≈ fft(x, (1, 2))
         @test @jit(fft(x_ra, (1, 2, 3))) ≈ fft(x, (1, 2, 3))
+        @test @jit(ifft(x_ra)) ≈ ifft(x)
+        @test @jit(bfft(x_ra)) ≈ bfft(x)
     end
 end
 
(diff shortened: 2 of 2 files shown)


ext/ReactantAbstractFFTsExt.jl
```diff
@@ -24,7 +24,7 @@ for op in (:rfft, :fft, :ifft)
         if __is_valid_stablehlo_fft_dims(dims, ndims(x))
             return @opcall fft(
                 TracedUtils.materialize_traced_array(x);
-                type=$(uppercase(string(op))),
+                type=($(uppercase(string(op)))),
```
giordano: Just trying to understand what's going on: the extra parentheses don't do anything, right?

ext/ReactantAbstractFFTsExt.jl
```diff
@@ -33,31 +33,62 @@ for op in (:rfft, :fft, :ifft)
...
+
     # Out-of-place plan
     plan_name = Symbol("Reactant", uppercase(string(op)), "Plan")
     plan_f = Symbol("plan_", op)
-    @eval struct $(plan_name){T} <: AbstractFFTs.Plan{T} end
-    @eval AbstractFFTs.$(plan_f)(::Reactant.TracedRArray{T}) where {T} = $(plan_name){T}()
-    @eval Base.:*(::$(plan_name){T}, x::Reactant.TracedRArray{T}) where {T} =
-        AbstractFFTs.$(op)(x)
+    @eval struct $(plan_name){T,D} <: AbstractFFTs.Plan{T}
+        dims::D
```
giordano, line 75: Is it necessary to use a type parameter for the dimension?  Can't we just make this `Int`?

ext/ReactantFFTWExt.jl
```diff
@@ -0,0 +1,100 @@
+module ReactantFFTWExt
+
+using FFTW
+using FFTW: AbstractFFTs, LinearAlgebra
```
giordano: It's dodgy to load packages from inside other packages.  `AbstractFFTs` is even already a trigger for this extension anyway
```suggestion
using AbstractFFTs
using LinearAlgebra
```

ext/ReactantAbstractFFTsExt.jl
ptiede: I don't think so? Tbh I don't even remember doing that. I think an LLM may have gotten me.

ptiede: Or maybe the formatter?

ext/ReactantAbstractFFTsExt.jl
ptiede: I was worried about getting a `TracedRNumber{<:Integer}` but if that should never happen I can change that.

ext/ReactantFFTWExt.jl
ptiede: Ya I was being lazy. Let me fix that.

ext/ReactantAbstractFFTsExt.jl
ptiede: Ok ya `JuliaFormatter` did this. I have no idea why.

giordano: Ok, I don't understand why JuliaFormatter does it and I'd classify it as an upstream bug, but I'm pretty sure they don't do anything.

ext/ReactantAbstractFFTsExt.jl
ptiede: Also, should I be worried about Int32 v.s. Int64?

ext/ReactantAbstractFFTsExt.jl
ptiede: :'(

ext/ReactantAbstractFFTsExt.jl
ptiede: Tests seem to pass when I get rid of the type parameter so I'll remove it.

ext/ReactantAbstractFFTsExt.jl
avik-pal: this is one of the v2 nonsense. pin juliaformatter to v1

ext/ReactantFFTWExt.jl
```diff
@@ -0,0 +1,101 @@
...
+using AbstractFFTs
+using LinearAlgebra
+using Reactant
+
+const FORWARD = -1
+const BACKWARD = 1
+
+#################################
+# c2c plans
+function Base.:*(plan::FFTW.cFFTWPlan{T,FORWARD,true}, x::Reactant.TracedRArray) where {T}
```
avik-pal: how are we ending up with a non-traced plan?

ptiede: So this happens in some of my code where I have some setup like

```julia
cache = prepare_cache(...) #creates a bunch of stuff including a FFTW plan
x = # some data that I want to trace and will be `Active` in Enzyme language
@compile computation(cache, Reactant.to_rarray(x))
```

In this case the FFTW plan is created outside of Reactant and isn't automatically converted during tracing since the `AbstractFFTs` extension only fixes the use case where the plan is created while tracing.

ext/ReactantAbstractFFTsExt.jl
ptiede: Ahh ok that's good to know.

ext/ReactantFFTWExt.jl
avik-pal: I think we should be able to do better than hardcode for FFTW. Specifically when tracing (see Tracing.jl) we can convert Abstract Plans https://github.com/JuliaMath/AbstractFFTs.jl/blob/04a14f5a3491697d9b5a6e17ffa994cf82ce2ab0/src/definitions.jl#L9 to our ReactantPlan (we might want to warn the user in that case). That way we are agnostic to whether the plan was created from FFTW or FastTransform.

ptiede:
I think all tests are passing, and the only missing lines in the code coverage are the function definitions. I am guessing those missing lines are an upstream Julia code coverage issue when working with Reactant?

ext/ReactantFFTWExt.jl
ptiede: The problem is that `Plan` is not enough to tell me what kind of FFT it is. I still need to know whether it is forward/backward FFT and whether it is R2C or C2R and `AbstractFFTs` doesn't give the required information in its abstract type to tell me that.

ptiede: I had thought about `make_tracer` to convert it, but I'd still need to hard code `FFTW` dispatches to create the correct plans. Now I could make this a lot more generic by defining some abstract `make_tracer` and a specialized dispatch method like `convert_plan` and then just define those in extensions but then because of Julia's extension system I think I'd need to define this `convert_plan` inside Reactant proper so that each extension could define something.

avik-pal: > The problem is that Plan is not enough to tell me what kind of FFT it is.

I see. Yeah that's fair.

> Julia's extension system I think I'd need to define this convert_plan inside Reactant proper so that each extension could define something.

I think it does make sense to convert the FFTW plan to reactant plan and forward those dispatches, that reduces some redundancy atleast

avik-pal: I dont have strong opinions here either way. We can merge and polish this in a follow-up

ptiede: Let me see if I can make this a bit more generic though :)

ptiede:
Ok, I tweaked how the plan conversion is done. The big question I have is whether I am allowed to access a `struct` defined in one extension in another. Since `FFTW` depends on `AbstractFFTs`, I guessed this might be ok, but it depends on some Julia internals I don't know, so someone who does should chime in. 

On the bright side, this implementation is much simpler, and if someone did want to add dispatches to other FFT libraries, they only need to define a single method now.

ext/ReactantFFTWExt.jl
ptiede: @avik-pal does the new version look better to you?

avik-pal: Yes this looks much better!

ext/ReactantAbstractFFTsExt.jl
```diff
@@ -72,7 +126,7 @@ for op in (:irfft,)
         if __is_valid_stablehlo_fft_dims(dims, ndims(x))
             return @opcall fft(
                 TracedUtils.materialize_traced_array(x);
-                type=$(uppercase(string(op))),
+                type=($(uppercase(string(op)))),
```
avik-pal: nit: remove this change

ext/ReactantAbstractFFTsExt.jl
```diff
@@ -82,13 +136,38 @@ for op in (:irfft,)
             @opcall(
                 fft(
                     TracedUtils.materialize_traced_array(permutedims(x, perm));
-                    type=$(uppercase(string(op))),
+                    type=($(uppercase(string(op)))),
```
avik-pal: same here

avik-pal (approved):

ext/ReactantAbstractFFTsExt.jl
ptiede: Ugh that silly formatter. Will do

ptiede: Ugh that silly formatter. Will do

ptiede:
Ok formatting fixed!

ext/ReactantAbstractFFTsExt.jl
giordano: > Tests seem to pass when I get rid of the type parameter so I'll remove it.

You didn't remove it 😅

ptiede: Ok this was some confusion on my end. I thought you were referring to the length for the `rfft` which I did remove. The type parameter `D` is for the `dims` field rather than the length. That can be essentially any iterable (see the FFTW definition here https://github.com/JuliaMath/FFTW.jl/blob/566f3c613ef3d3ed0150e87daf63900715bc8eb4/src/fft.jl#L254C1-L273C1)

so I don't think I can explicitly type that.

ptiede:
I'm not sure if the errors are from me.

wsmoses:
er cc @avik-pal since those seem to dump about a bunch of protobuf stuff you were working on earlier

ext/ReactantAbstractFFTsExt.jl
ptiede: @giordano, is this explanation alright, or do you want me to see if I can always collect the dims into a common type?

giordano: Ok, makes sense.

ptiede:
I made some tweaks that should fix a small bug and add support for the other plan AbstractFFTs supports, `bfft`.

ptiede:
The two errors don't appear to be from me? Is there anything else I should work on for this PR?

PR #2100, opened 2026-01-10
feat: auto-generate protobuf

avik-pal:


avik-pal:
This would allow us to not have to update the C functions for every new flag we want to add. Haven't updated the ccall's yet, will do that in a follow up PR

wsmoses:
Can you separate the autogenerated code into a separate PR? And this be the code in Julia or c++ which produces or consumes

giordano:
Yes, I don't know how to review a 16k-line PR, GitHub UI wouldn't even let me see more than one file at a time, I have no clue of what to look at.

avik-pal:
done. added a stub Proto.jl for now

deps/ReactantExtra/make-proto-bindings.jl
```diff
@@ -0,0 +1,272 @@
...
+"""
+    write_main_module(output_dir::String, generated::Vector{String})
+
+Create a main Proto module file that exports all generated types.
+"""
+function write_main_module(output_dir::String, generated::Vector{String})
+    # Find all generated Julia files
+    jl_files = String[]
+    for file in readdir(output_dir)
+        if endswith(file, ".jl") && file != "Proto.jl" && file != ".gitkeep"
```
giordano: `file != ".gitkeep"` is redundant, right?  `.gitkeep` wouldn't match `endswith(file, ".jl")` anyway.  Also, `jl_files` can be defined in one line with
```julia
jl_files = [file for file in readdir(output_dir) if endswith(file, ".jl") && file != "Proto.jl"]
```

PR #2103, opened 2026-01-10
feat: use generated proto for XLA.compile

avik-pal:


avik-pal:
The proto defaults are different from the xla defaults probably

src/xla/IFRT/LoadedExecutable.jl
```diff
@@ -73,35 +73,22 @@ end
...
+    num_outputs::Int64,
+    is_sharded::Bool,
     num_replicas::Int64,
     num_partitions::Int64,
-    use_shardy_partitioner::Bool,
 )
-    device_id = is_sharded ? Int64(-1) : Int64(XLA.device_ordinal(device))
-    GC.@preserve client mod begin
+    compile_options_bytes = Reactant.ProtoUtils.proto_to_bytes(compile_options)
+    GC.@preserve client mod compile_options_bytes begin
```
wsmoses, line 85: what happened to the device ids?

avik-pal: it is set in `XLA.make_compile_options`

PR #2106, opened 2026-01-10
feat[cpp]: use protobuf to pass options to compile

avik-pal:


avik-pal:
ifrt exports are added via `_ifrt*`

PR #2119, opened 2026-01-13
feat[cpp]: extract default options from xla

avik-pal:


deps/ReactantExtra/API.cpp
```diff
@@ -3679,3 +3680,21 @@ REACTANT_ABI int XSpaceToToolsData(
 
   return 0;
 }
+
+REACTANT_ABI void *ReactantGetDebugOptions(size_t *size) {
```
wsmoses, line 3684: we have a free already vendored right?

avik-pal: yes

PR #2120, opened 2026-01-14
Add checkpoints option to @trace macro for periodic checkpointing

michel2323:
Adds support for specifying the number of checkpoints when using periodic checkpointing with @trace loops. Integrates with https://github.com/EnzymeAD/Enzyme-JAX/pull/1910.

```julia
# Static bounds: defaults to isqrt(num_iters)
@trace checkpointing=true for i in 1:100
    # uses checkpoints = 10
end

# Explicit checkpoints
@trace checkpointing=true checkpoints=20 for i in 1:100
    # uses checkpoints = 20
end
```

- For for loops with static bounds, defaults to isqrt(num_iters)
- For while loops or for loops with dynamic bounds, checkpoints must be specified explicitly
- Sets enzymexla.checkpoints MLIR attribute on while operations

### DIFF
--- a/test/core/control_flow.jl
+++ b/test/core/control_flow.jl
@@ -710,9 +710,9 @@ end
 
     ir = @code_hlo optimize = "enzyme-batch" for_no_track_numbers(x_ra, n_ra)
     @test @filecheck begin
-        @check "enzyme.disable_mincut"
-        @check "enzymexla.enable_checkpointing"
-        @check "enzymexla.checkpoints = 3"
+        @check_dag "enzyme.disable_mincut"
+        @check_dag "enzymexla.enable_checkpointing"
+        @check_dag "enzymexla.checkpoints = 3"
         ir
     end
 end
@@ -741,8 +741,9 @@ end
         show, @code_hlo optimize = "enzyme-batch" for_explicit_checkpoints(x_ra, n_ra)
     )
     @test @filecheck begin
-        @check "enzymexla.enable_checkpointing"
-        @check "enzymexla.checkpoints = 5"
+        @check_dag "enzymexla.enable_checkpointing"
+        @check_dag "enzymexla.checkpoints = 5"
+        ir
     end
 end
 
@@ -766,8 +767,8 @@ end
 
     ir = sprint(show, @code_hlo while_explicit_checkpoints(x_ra, n_ra))
     @test @filecheck begin
-        @check "enzymexla.enable_checkpointing"
-        @check "enzymexla.checkpoints = 5"
+        @check_dag "enzymexla.enable_checkpointing"
+        @check_dag "enzymexla.checkpoints = 5"
         ir
     end
 end


wsmoses:
instead of adding a new kwarg, could we pass a different value to checkpointing, maybe Periodic(10) or something?

sriharikrishna:
> instead of adding a new kwarg, could we pass a different value to checkpointing, maybe Periodic(10) or something?

Is this a purely a Julia frontend change or would it also require a change of the MLIR attributes that are checked in https://github.com/EnzymeAD/Enzyme-JAX/pull/1910.

michel2323:
> instead of adding a new kwarg, could we pass a different value to checkpointing, maybe Periodic(10) or something?

Fixed

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -3,7 +3,30 @@ module ReactantCore
...
+"""
+    Periodic(n::Int)
+
+Checkpointing strategy for traced loops that specifies periodic checkpointing with `n` checkpoints.
+
+# Examples
+
+```julia
+# Explicit periodic checkpointing with 10 checkpoints
+@trace checkpointing=Periodic(10) for i in 1:100
```
Pangoraw: can you put another value like 4, since 10 is already the isqrt

michel2323: Fixed and rebased.

test/core/control_flow.jl
```diff
@@ -2,6 +2,7 @@ using Reactant, Test
 using LinearAlgebra
 using Reactant.ReactantCore
 using Reactant: MLIR
+using Reactant.ReactantCore: Periodic
```
Pangoraw: nit: can you import from Reactant
```suggestion
using Reactant: Periodic
```

Pangoraw (approved):

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -299,6 +322,16 @@ function trace_while(expr; track_numbers, mincut, checkpointing, first_arg=nothi
 
     error_if_any_control_flow(body)
 
+    # Validate checkpointing usage - for raw while loops (not from trace_for),
+    # Periodic(n) must be used since we can't compute default checkpoints
+    if checkpointing === true && first_arg === nothing
```
wsmoses, line 327: can you give a test for this error message, I'm a bit confused when it would occur?

Pangoraw: IIUC, this is for when using `@trace while` and not `@trace for` but this is not really possible to compile end-to-end for now has we hit the following error:

```
 error: WhileOp does not have known iteration count for cache removal
```

since our cache removal for while currently relies on the while op having a proper identifiable for-loop style.

PR #2122, opened 2026-01-15
set GPUCompiler cache key type on CUDA ext to MLIR Module

mofeing:
this should fix the issue in #2093 about not linking properly the CUDA LLVM IR on different `@compile` invocations.

the problem i see with this implementations is that it will (redundantly) rerun CUDA compilation inside a nested submodule, but i find that an edge case.

an alternative would be to define sth like a `Reactant.CompilationSession` that we can certify that there's one truly assigned during a `@compile` invocation, and use that as key type.

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -4110,7 +4110,7 @@ function llvm_compiler_cache(_mod::MLIR.IR.Module)
     cache = get(_llvm_compiler_caches, _mod, nothing)
     if cache === nothing
         cache = Dict{Any,LLVMFunc}()
-        _compiler_caches[_mod] = cache
+        _llvm_compiler_caches[_mod] = cache
     end
     return cache
 end


mofeing:
do you mean to remove it from here? https://github.com/EnzymeAD/Reactant.jl/blob/36f9a1677cd743b10db0dc973efd7ade36d7199a/ext/ReactantCUDAExt.jl#L1304-L1305

no, we don't. it looks a lil bit dangerous to free it when the module finishes... imo that fits better with a `CompileSession` alike.

wsmoses:
it's safe to clear the cache key for sure at the end of compile as this list doesn't maintain ownership, just avoiding recompilation of the gpu kernel. We will never recompile a gpu kernel into the same module once we finish @ compile

mofeing:
I've moved `LLVMFunc` and `_compiler_caches` to `Reactant.Compiler` because adding hooks for pkg extensions is a headache and it can also serve us for AMD/ROCm kernels in the future.

PR #2123, opened 2026-01-15
add self dependency for tests

mofeing:
required for Julia 1.12

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -79,4 +79,4 @@ Zygote = "0.7"
 Reactant_jll = "0192cb87-2b54-54ad-80e0-3be72ad8a3c0"
 
 [sources]
-ReactantCore = {path = ".."}
+Reactant = {path = ".."}


wsmoses:
@giordano you know this better than I if you can take a look

giordano:
> required for Julia 1.12

This is fine (and coupled with the `[sources]` section it makes it easy to activate the right environment with the right Reactant), but not sure it's _required_ :thinking:

mofeing:
> ... but not sure it's required 🤔

I get this error locally with Julia 1.12.4

```
     Testing Running tests...
ERROR: LoadError: ArgumentError: Package Reactant not found in current path.
- Run `import Pkg; Pkg.add("Reactant")` to install the Reactant package.
```

giordano:
Uhm, maybe that's because there's a workspace https://github.com/EnzymeAD/Reactant.jl/blob/79ce7f7f3ae6912b8b15e67e7d7a24abd10d4cd3/Project.toml#L136-L137?  But not sure.  Anyway, it's a good chance regardless of whether it's required or not.

PR #2133, opened 2026-01-16
fix: various errors caught by JETLS

avik-pal:
see https://github.com/aviatesk/JETLS.jl/issues/478

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -90,7 +90,7 @@ end
     return Base.setproperty!(obj, field, val)
 end
 
-@inline traced_setfield!(@nospecialize(obj), field, val, path) =
+@inline traced_setfield!(@nospecialize(obj), field, val, _path) =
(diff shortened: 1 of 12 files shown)


avik-pal:
somehow I managed to cause a seg fault, investigating

src/Compiler.jl
```diff
@@ -90,7 +90,7 @@ end
     return Base.setproperty!(obj, field, val)
 end
 
-@inline traced_setfield!(@nospecialize(obj), field, val, path) =
+@inline traced_setfield!(@nospecialize(obj), field, val, _) =
```
wsmoses: Even if unused I think keeping the names here is nice

avik-pal: keeping the names, but marking with a prefix `_` to make the warnings go away

wsmoses (approved):

PR #2146, opened 2026-01-19
fix: skip integer inputs for finite difference

avik-pal:


src/TestUtils.jl
```diff
@@ -110,7 +110,14 @@ function finite_difference_gradient(
             # we will naively insert the args here, cse will take care of the rest
             new_arguments = TracedRArray[]
 
-            epsilon = default_epslion(method, Reactant.unwrapped_eltype(arg))
+            elT = Reactant.unwrapped_eltype(arg)
```
wsmoses, line 114: we can potentially do better, how about  Enzyme.Compiler.guaranteed_const(elT)

src/TestUtils.jl
```diff
@@ -110,7 +110,14 @@ function finite_difference_gradient(
...
+            if elT <: Integer
```
avik-pal:
```suggestion
            if Enzyme.Compiler.guaranteed_const(elT)
```

PR #2161, opened 2026-01-20
Early error for invalid complex of int

wsmoses:
realedly, do we know where this might be generated from.

In any case erring julia side is better than erring within a pass


jumerckx (approved):
I saw this being generated here for example https://github.com/EHTJulia/VLBISkyModels.jl/pull/114

Issue #2164, opened 2026-01-20
copyto! with reshaped array scalar indexing warning


ptiede:
MWE

```julia
foo(out, g, I) = copyto!(out, reshape(@view(g[I]), :))

outr = Reactant.to_rarray(zeros(6, 6))
gr = Reactant.to_rarray(ones(6, 6))
Ir = Reactant.to_rarray(eachindex(1:length(outr)))

@compile foo(outr, gr, Ir)
ERROR: Scalar indexing is disallowed.
Invocation of getindex(::TracedRArray, ::Vararg{Int, N}) resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] errorscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:151
  [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:124
  [4] assertscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:112
  [5] getindex(a::Reactant.TracedRArray{Int64, 1}, index::Int64)
    @ Reactant.TracedIndexing ~/.julia/dev/Reactant/src/Indexing.jl:40
  [6] iterate
    @ ./abstractarray.jl:1216 [inlined]
  [7] iterate
    @ ./abstractarray.jl:1214 [inlined]
  [8] checkindex(::Type{Bool}, inds::Base.OneTo{Int64}, I::Reactant.TracedRArray{Int64, 1})
    @ Base ./abstractarray.jl:774
  [9] checkbounds
    @ ./abstractarray.jl:687 [inlined]
 [10] checkbounds
    @ ./abstractarray.jl:702 [inlined]
 [11] view
    @ ./subarray.jl:184 [inlined]
 [12] foo
    @ ./REPL[53]:1
[...]

Issue #2154, opened 2026-01-20
Broadcasting assignment error with slicing an array with a TracedRArray{Int}


ptiede:
Needed for some NFFT code

```julia
function foo(outr, arr, dindx)
    @allowscalar @inbounds outr[dindx] .= arr
    return nothing
end

f = rand(ComplexF64, 64, 64)
p = rand(ComplexF64, 64 * 64)
fr = Reactant.to_rarray(f)
pr = Reactant.to_rarray(p)
dindxr = Reactant.to_rarray(collect(1:length(fr)))

foo(f, p, 1:4096)
@compile foo(fr, pr, dindxr)

ERROR: MethodError: no method matching CartesianIndex(::Tuple{Reactant.TracedRNumber{Int64}})

Closest candidates are:
  CartesianIndex(::Tuple{Vararg{Integer, N}}) where N
   @ Base multidimensional.jl:78
  CartesianIndex(::Tuple{Vararg{Union{Integer, CartesianIndex}}})
   @ Base multidimensional.jl:90
  CartesianIndex(::Integer...)
   @ Base multidimensional.jl:79
  ...

Stacktrace:
  [1] (::Reactant.TracedUtils.var"#3#4"{Tuple{Reactant.TracedRArray{Int64, 1}}})(idx::CartesianIndex{1})
    @ Reactant.TracedUtils ~/.julia/dev/Reactant/src/TracedUtils.jl:201
  [2] iterate
    @ ./generator.jl:47 [inlined]
  [3] _collect
    @ ./array.jl:854 [inlined]
  [4] collect_similar
    @ ./array.jl:763 [inlined]
  [5] map
    @ ./abstractarray.jl:3286 [inlined]
  [6] _get_ancestor_and_indices_linear
    @ ~/.julia/dev/Reactant/src/TracedUtils.jl:200 [inlined]
  [7] get_ancestor_and_indices_inner
    @ ~/.julia/dev/Reactant/src/TracedUtils.jl:193 [inlined]
  [8] get_ancestor_and_indices
    @ ~/.julia/dev/Reactant/src/TracedUtils.jl:172 [inlined]
  [9] set_mlir_data!(x::SubArray{…}, data::Reactant.MLIR.IR.Value)
    @ Reactant.TracedUtils ~/.julia/dev/Reactant/src/TracedUtils.jl:155
 [10] _copyto!
    @ ~/.julia/dev/Reactant/src/TracedRArray.jl:538 [inlined]
[...]

PR #2171, opened 2026-01-21
fix: setindex of subarray with traced

avik-pal:
This kind of works, but it produces a gigantic concatenate (which is later folded but that means a very long compile time). Long term we would want to implement our own reindex to bypass this issue.

fixes https://github.com/EnzymeAD/Reactant.jl/issues/2154
fixes https://github.com/EnzymeAD/Reactant.jl/issues/2164

### DIFF
--- a/test/indexing.jl
+++ b/test/indexing.jl
@@ -476,8 +476,8 @@ function bcast_setindex!(outr, arr, dindx)
 end
 
 @testset "bcast setindex!" begin
-    f = rand(ComplexF64, 4, 4)
-    p = rand(ComplexF64, 4 * 4)
+    f = rand(ComplexF32, 4, 4)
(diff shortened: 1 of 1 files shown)


wsmoses (approved):
I'm guessing from context you saw that our initial rewrite into the broadcast hit the issue in the copyro

avik-pal:
```
checkindex(::Type{Bool}, inds::Base.OneTo{Int64}, i::Reactant.TracedRNumber{Int64})
```

Should we always skip bounds checking (with a warning)? CI will run with check bounds so inbounds specification wont work

wsmoses:
I think that's reasonable in context here.

Longer term if we want we can register a custom call for error handling and actually have an if statement to check it properly

PR #2174, opened 2026-01-21
feat: patch ProtoBuf to generate mutable structs

avik-pal:


gbaraldi:
So I took a much longer look and this won't help. The functions that are generated are just super unwieldy for the big structs like DebugOptions. They generate several mb of IR each. An actual fix would be to not use a static structure but move to some more dynamic handling of those protobufs by making them for example Dicts. I doubt we need the extreme runtime performance that specializing on a message gives you

gbaraldi:
I did an experiment of turning photo into a separate sub package like ReactantCore. It takes 20 seconds to compile by itself 🥶

avik-pal:
We can do https://github.com/EnzymeAD/Reactant.jl/pull/2175 + make the functionality available via another package (which must be manually loaded).

avik-pal:
Converting to a dict is also probably pretty simple. We convert the structs to dicts and then forward the getproperty to the dict

giordano:
Bump.  Reactant precompile time is still atrocious, which also badly affects CI here
```
  492950.5 ms  ✓ Reactant
```

avik-pal:
follow up in https://github.com/EnzymeAD/Reactant.jl/pull/2522

PR #2175, opened 2026-01-21
fix: only use protobuf path for unconventional options

avik-pal:


avik-pal:
Dont merge before https://github.com/EnzymeAD/Reactant.jl/pull/2174

PR #2176, opened 2026-01-21
feat: convert protos to be dicts [skip ci]

avik-pal:
@gbaraldi would something like this work?

### DIFF
--- a/src/proto/google/protobuf/any_pb.jl
+++ b/src/proto/google/protobuf/any_pb.jl
@@ -5,7 +5,7 @@ using ProtoBuf.EnumX: @enumx
 export var"#Any"
 
 
-struct var"#Any"
+mutable struct var"#Any"
     type_url::String
     value::Vector{UInt8}
 end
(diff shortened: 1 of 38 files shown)


gbaraldi:
This doesn't precompile atm but I suspect this would work

src/proto/google/protobuf/any_pb.jl
```diff
@@ -5,7 +5,7 @@ using ProtoBuf.EnumX: @enumx
 export var"#Any"
 
 
-struct var"#Any"
+mutable struct var"#Any"
     type_url::String
     value::Vector{UInt8}
```
giordano, line 10: For all the immutable structs that have been changed to mutable, fields should be marked as constant, otherwise they lose their property. Also, I'm not sure this is necessary for small structs like this one with 1-2 fields (these aren't even parametric)

gbaraldi: It’s unclear where the line is. Though that one looks too small

PR #2185, opened 2026-01-22
More don't rewrite in interp

wsmoses:


wsmoses:
okay these failures are unrelated, but @avik-pal related to your recent copy pr

PR #2196, opened 2026-01-23
feat: support size(x, i) where i <: Traced

avik-pal:
fixes the remaining failures in https://github.com/LuxDL/Lux.jl/pull/1631

fixes https://github.com/LuxDL/Lux.jl/issues/1637

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -50,11 +50,7 @@ Base.size(x::TracedRArray) = x.shape
 Base.size(x::TracedRArray, i::Integer) = x.shape[i]
 
 function Base.size(x::TracedRArray, i::TracedRNumber{<:Integer})
-    branches = []
-    for idx in 1:ndims(x)
-        push!(branches, () -> TracedRNumber{Int32}(size(x, idx)))
-    end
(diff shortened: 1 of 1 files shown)


src/TracedRArray.jl
```diff
@@ -47,6 +47,15 @@ TracedRArray{T,N}(x::AbstractArray) where {T,N} = convert(TracedRArray{T,N}, x)
 Base.Tuple(x::TracedRArray) = ntuple(Base.Fix1(getindex, x), length(x))
 
 Base.size(x::TracedRArray) = x.shape
+Base.size(x::TracedRArray, i::Integer) = x.shape[i]
+
+function Base.size(x::TracedRArray, i::TracedRNumber{<:Integer})
+    branches = []
+    for idx in 1:ndims(x)
```
wsmoses: the case op for sure is useful (and we should keep the op), but is there any reason not to just dynamic slice into convert(TracedRArray, size(x)) or something?

or alternatively to do cases of stablehlo.get_dimension_size [which will future proof us for dynamic dimensions], since its going to be a traced return anyways

avik-pal: Ah that is definitely smarter :sweat_smile:

PR #2199, opened 2026-01-23
fix: finite differences for complex inputs

avik-pal:


src/TestUtils.jl
```diff
@@ -36,13 +36,20 @@ function default_epslion(::Val{fdtype}, ::Type{T}) where {fdtype,T}
...
-        LinearAlgebra.Diagonal(fill(epsilon, length(x)));
+        TracedRArray{real(elT),2}, LinearAlgebra.Diagonal(fill(epsilon, length(x)))
     )
-    return permutedims(
+    perturbation = permutedims(
         reshape(onehot_matrix, size(x)..., length(x)), (ndims(x) + 1, 1:(ndims(x))...)
     )
+    # For complex numbers, we need to perturb real and imaginary parts separately
+    if elT <: Complex
+        real_perturbation = complex.(perturbation, zero(perturbation))
```
wsmoses, line 48: so we need to be careful here because forward and reverse mode should have different conventions here. see : https://enzymead.github.io/Enzyme.jl/stable/faq/#Complex-numbers

I think this is equivalent to forward mode

avik-pal: should we document this is only valid for holomorphic functions?

wsmoses: yeah, or at least we should document the convention [which ofc for holomorphic is equivalent across all conventions]

wsmoses: actually I don't think that's the case, are you passing in the equivalent of 1+1im ? [except onehot]

avik-pal: `1 + 0i` and `0 + 1i`

wsmoses: ah, I guess I'm not understanding how we combine them then for the final answer?

avik-pal: similar to how FiniteDiff does it https://github.com/JuliaDiff/FiniteDiff.jl/blob/f37d23dd0e3eba050309fd1d0e88be0bb53867e3/src/gradients.jl#L327-L337

PR #2208, opened 2026-01-24
windows dump final module jit debug

wsmoses:


wsmoses:
https://github.com/llvm/llvm-project/blob/485eb62afd7edb1ecb13c975bf62da03ff166b60/lld/COFF/MinGW.cpp#L232 ?

wsmoses:
@avik-pal I'm willing to disable this for now, open an issue, and come back to later. Seems specific to windows arm in a way I don't fully understand

avik-pal:
Yeah let's do that

PR #2228, opened 2026-01-26
feat: support for rounding modes

avik-pal:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/2220

### DIFF
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -899,6 +899,11 @@ end
     @testset "$T" for (convfn, T) in [
         (identity, Float64), (x -> Float32.(x), Float32), (x -> floor.(Int32, x), Int32)
     ]
+        if RunningOnTPU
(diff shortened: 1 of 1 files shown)


test/basic.jl
```diff
@@ -871,6 +871,42 @@ end
...
+        end
+    end
+end
+
+@testset "mod2pi" begin
+    a = [-1.1, 7.7, -3.3, 9.9, -5.5]
+
+    @testset "$T" for (convfn, T) in [
+        (identity, Float64), (x -> Float32.(x), Float32), (x -> floor.(Int32, x), Int32)
+    ]
```
avik-pal, line 901:
```suggestion
    ]
        if RunningOnTPU
            @warn "Skipping rem2pi test on TPU. F64 bitcast not supported on TPU"
            break
        end

```

PR #2272, opened 2026-01-27
A few MLIR.IR fixes

Pangoraw:
Following up on #2181.

### DIFF
--- a/src/mlir/IR/Utils.jl
+++ b/src/mlir/IR/Utils.jl
@@ -34,14 +34,10 @@ Base.String(str::API.MlirIdentifier) = String(API.mlirIdentifierStr(str))
 
 function visit(f, op)
     all_ok = true
-    for region in RegionIterator(op)
-        for block in BlockIterator(region)
-            for op in OperationIterator(block)
-                all_ok &= visit(f, op)
-                r = f(op)
-                if r isa Bool
-                    all_ok &= r
-                end
+    for region in op
+        for block in region
+            for op in block
+                all_ok &= f(op)
             end
         end
     end


src/mlir/IR/Operation.jl
```diff
@@ -291,7 +291,7 @@ The other operation must belong to a block.
 In any case, the ownership is transferred to the block of the other operation.
 """
 function move_before!(op::Operation, other::Operation)
-    lose_ownership!(op)
+    op.owned && lose_ownership!(op)
```
mofeing, line 294: right, although `owned` field is gonna disappear when I rebase #2093

src/mlir/IR/Utils.jl
```diff
@@ -34,10 +34,14 @@ Base.String(str::API.MlirIdentifier) = String(API.mlirIdentifierStr(str))
...
-            for op in block
-                all_ok &= f(op)
+    for region in RegionIterator(op)
+        for block in BlockIterator(region)
+            for op in OperationIterator(block)
+                all_ok &= visit(f, op)
+                r = f(op)
+                if r isa Bool
+                    all_ok &= r
+                end
```
mofeing: right, I was planning to integrate `*Iterator`s directly into their wrapping types (i.e. `iterate(op)` returns a `Region`, `iterate(region)` returns a `Block` and `iterate(block)` returns a `Operation`).

what do you think? this is actually the logic in MLIR (an Operation contains Regions, ...)

src/mlir/IR/Pass.jl
```diff
@@ -168,22 +177,23 @@ end
...
-    DUMP_MLIR_ALWAYS[] && dump_mlir(mod, pm, isempty(key) ? "pre_pm" : "pre_$(key)_pm")
+    DUMP_MLIR_ALWAYS[] &&
+        dump_mlir(operation, pm, isempty(key) ? "pre_pm" : "pre_$(key)_pm")
     status = LogicalResult(@static if isdefined(API, :mlirPassManagerRunOnOp)
-        API.mlirPassManagerRunOnOp(pm, Operation(mod))
+        API.mlirPassManagerRunOnOp(pm, operation)
     else
-        API.mlirPassManagerRun(pm, mod)
+        API.mlirPassManagerRun(pm, Module(operation))
     end)
```
mofeing: it's crashing here because here `operation` is actually a `Module`.

I'm ignoring old MLIR versions, so prob we can just remove the `@static if`?

```suggestion
    status = LogicalResult(API.mlirPassManagerRunOnOp(pm, Operation(operation)))
```

and add the following method

```julia
run!(pm::PassManager, mod::Module, key::String="") = run!(pm, Operation(mod), key)
```

src/mlir/IR/Pass.jl
```diff
@@ -89,10 +89,19 @@ const MLIR_DUMP_COUNTER = Threads.Atomic{Int}(0)
...
+    mod::Module, pm::Union{Nothing,PassManager}=nothing, mode::String=""; failed::Bool=false
+)
+    return dump_mlir(Operation(mod), pm, mode; failed)
+end
+
 # Utilities for dumping to a file the module of a failed compilation, useful for
 # debugging purposes.
 function dump_mlir(
-    mod::Module, pm::Union{Nothing,PassManager}=nothing, mode::String=""; failed::Bool=false
+    mod::Operation,
```
mofeing: maybe you can just rename this `module_op` or `op`?

src/mlir/IR/Utils.jl
Pangoraw: would be also good to, just needed it rn

mofeing: yeah, let's fix it and will try opening the other PR today
sorry for this, I thought I hadn't modified this

mofeing: now that #2275 has been merged, this change can be removed

Pangoraw: Would like to keep the new recursive functionality, wdyt?

mofeing: i have no hard opinion on that, but `verifyall(::Operation)` uses it, and it requires to also call `verify` on the intermediate ops... so how about a `flatten` kwarg instead?

PR #2277, opened 2026-01-27
Support push!/pop!/append!/pushfirst!/popfirst! for TracedRArray{T,1}

### LINKED ISSUE #2148: BoundsError when push! grows a vector inside mutable struct during compilation
## Summary

When using `@compile` on a function that uses `push!` to grow a vector inside a mutable struct, the compiled function fails with a `BoundsError` during result extraction.

## MWE

```julia
using Reactant
using Reactant: @compile

mutable struct Integrator{T}
    u::T
    history::Vector{T}
end

function simulate!(integ)
    for i in 1:3
        integ.u = integ.u .* 0.5
        push!(integ.history, copy(integ.u))  # This grows the vector
    end
    return integ.u
end

# Works without Reactant
x = [1.0, 2.0, 3.0]
integ_normal = Integrator(copy(x), [copy(x)])
simulate!(integ_normal)  # history has 4 elements

# Fails with Reactant
x_reactant = Reactant.to_rarray([1.0, 2.0, 3.0])
integ_reactant = Integrator(x_reactant, [x_reactant])  # history has 1 element
compiled = @compile simulate!(integ_reactant)
compiled(integ_reactant)  # BoundsError
```

## Error

```
BoundsError: attempt to access 1-element Vector{ConcretePJRTArray{Float64, 1, 1}} at index [2]
Stacktrace:
 [1] throw_boundserror
 [2] getindex
 [3] traced_getfield
 [4] traced_setfield_buffer!
 ...
```

## Analysis

The issue is that `push!` grows the vector from 1 to 4 elements during tracing, but Reactant's result extraction (`traced_setfield_buffer!`) expects the vector to have the same length as when the function was called. It tries to access indices 2, 3, 4 of a 1-element vector.

## Impact

This pattern is very common in iterative algorithms like ODE solvers where solution history is accumulated via `push!` during the solve loop. This blocks using Reactant with OrdinaryDiffEq.jl.

## Workaround

Pre-allocating the vector with the correct size works:

```julia
function simulate_preallocated!(integ, n_steps)
    for i in 1:n_steps
        integ.u = integ.u .* 0.5
        integ.history[i+1] = copy(integ.u)  # Use pre-allocated slot
    end
    return integ.u
end

x_reactant = Reactant.to_rarray([1.0, 2.0, 3.0])
history_preallocated = [copy(x_reactant) for _ in 1:4]
integ_preallocated = Integrator(x_reactant, history_preallocated)
compiled = @compile simulate_preallocated!(integ_preallocated, 3)
compiled(integ_preallocated, 3)  # Works!
```

However, this workaround isn't always feasible because:
1. The number of iterations may not be known ahead of time (adaptive algorithms)
2. It requires significant refactoring of existing code

## Environment

- Reactant v0.2.199
- Julia 1.11

ChrisRackauckas-Claude:
## Summary

- Implements `push!`, `pop!`, `pushfirst!`, `popfirst!`, and `append!` for `TracedRArray{T,1}` by using `Ops.concatenate` and `Ops.slice` to produce new MLIR data, then updating `mlir_data` and `shape` in-place on the array (per wsmoses's suggested approach in #2148)
- Fixes a compiler bug where XLA `tf.aliasing_output` was set between input/output buffers of different shapes, causing `INTERNAL: Expected aliased input ... to have the same size` errors. Added a type equality check before setting the alias attribute.

## Test plan

- [x] Added 6 tests in `test/basic.jl` covering `push!`, `push!` with multiple items, `pushfirst!`, `pop!`, `popfirst!`, and `append!`
- [x] All new tests pass locally
- [x] CI passes

Closes #2148

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -1364,4 +1364,6 @@ function Base.append!(a::TracedRArray{T,1}, b::TracedRArray{T,1}) where {T}
     return a
 end
 
+Base.extrema(A::TracedRArray) = minimum(A), maximum(A)
+
 end
--- a/ext/ReactantStatisticsExt.jl
+++ b/ext/ReactantStatisticsExt.jl
@@ -1,7 +1,10 @@
 module ReactantStatisticsExt
 
-using Reactant: AnyTracedRArray
-using Statistics: Statistics
+using Reactant: AnyTracedRArray, AnyTracedRMatrix, AnyTracedRVector, TracedRNumber
+using ReactantCore: materialize_traced_array
+using GPUArraysCore: @allowscalar
+using LinearAlgebra: diag, tril!, Diagonal
+using Statistics: Statistics, middle
 
 function Statistics._mean(f::F, A::AnyTracedRArray{T,N}, dims) where {F,T,N}
     denom = dims isa Colon ? length(A) : prod(Base.Fix1(size, A), Tuple(dims))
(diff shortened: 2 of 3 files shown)


src/TracedRArray.jl
```diff
@@ -1312,4 +1312,55 @@ function Base.permutedims!(dest::TracedRArray, src::AnyTracedRArray, perm)
     return dest
 end
 
+function Base.push!(a::TracedRArray{T,1}, items...) where {T}
```
wsmoses, line 1326: @avik-pal @mofeing does julia push/pop/etc take multiple args like that?

mofeing: apparently, [it does](https://docs.julialang.org/en/v1/base/collections/#Base.push!)

but I would probably make one array with `items` and call `concatenate` just once... if that's doable

src/Compiler.jl
```diff
@@ -2505,7 +2505,7 @@ function compile_mlir!(
                 donated_args_mask[i] = true
 
                 residx = findfirst(Base.Fix1(===, arg), linear_results2)
-                if residx !== nothing
+                if residx !== nothing && in_tys[i] == out_tys2[residx]
```
mofeing, line 2703: what is the reasoning behind this change?

avik-pal: With this PR, the size of the tensors can change. Previously we never needed to check that.

PR #2281, opened 2026-01-28
ci: workflow to update XPROF commit

avik-pal:


.github/workflows/update-xprof.yml
```diff
@@ -0,0 +1,28 @@
+name: "Open PR to update Xprof commit"
+
+on:
+  schedule:
+    - cron: '19 16 * * *'
```
giordano: I recommend opening the PR at a different time, to avoid opening multiple PRs all together, overloading the CI system, more than it already is
```suggestion
    - cron: '19 04 * * *'
```

PR #2284, opened 2026-01-28
feat: add new passes + update jll

avik-pal:


wsmoses:
we should fix windows, otherwise lgtm

Project.toml
```diff
@@ -122,7 +122,7 @@ PythonCall = "0.9.25"
 Random = "1.10"
 Random123 = "1.7"
 ReactantCore = "0.1.16"
-Reactant_jll = "0.0.310"
+Reactant_jll = "0.0.314"
```
avik-pal:
```suggestion
Reactant_jll = "0.0.315"
```

PR #2291, opened 2026-01-28
fix `traced_type_inner` on `UnionAll` of concrete types

mofeing:
cc @maximilian-gelbrecht

### DIFF
--- a/test/core/tracing.jl
+++ b/test/core/tracing.jl
@@ -120,6 +120,7 @@ end
                 (Array{Float64}, Array{Float64}, Array{TracedRNumber{Float64}}),
                 (Matrix, Matrix, Matrix),
                 (Array{<:AbstractFloat}, Array{<:AbstractFloat}, Array{<:AbstractFloat}),
+                (Array{>:Float32}, Array{>:Float32}, Array{>:Float32}),
                 (ConcreteRNumber, TracedRNumber, TracedRNumber),
                 (ConcreteRArray, TracedRArray, TracedRArray),
                 (ConcreteRArray{Float64}, TracedRArray{Float64}, TracedRArray{Float64}),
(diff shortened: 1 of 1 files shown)


mofeing:
@wsmoses is there a reason why `traced_type(ConcreteRArray)` should return `TracedRArray{T,N}` instead of `TracedRArray`?

wsmoses:
fix test breakage, otherwise lgtm

mofeing:
tests are fixed but the behavior on `UnionAll` still feels a lil bit arbitrary as
- `ConcreteRArray` maps to `TracedRArray{T,N} where {T,N}` and not `TracedRArray`
- but `Array` maps to `Array{Any}`

wsmoses:
TracedRArray{T,N} where {T,N} === TracedRArray ?

test/tracing.jl
```diff
@@ -115,6 +115,14 @@ end
                     Union{Nothing,TracedRArray{Float64,1}},
                 ),
 
+                # UnionAll types
+                (
+                    ConcreteRArray,
+                    TracedRArray{T,N} where {T,N},
+                    TracedRArray{T,N} where {T,N},
+                ),
+                (Array, Array{Any}, Array{Any}),
```
wsmoses: this is wrong, and shouldn't happen

mofeing:
you're right: `TracedRArray{T,N} where {T,N} === TracedRArray`

but the tests are failing because what `traced_type` returns is not `TracedRArray` but `TracedRArray.body.body` (or equally, `Base.unwrap_unionall(TracedRArray)`) which is no longer a `UnionAll` but a `DataType` and gets printed as `TracedRArray{T,N}` without the `where`.

```
ConcreteIFRTArray: Test Failed at /home/runner/work/Reactant.jl/Reactant.jl/test/tracing.jl:236
  Expression: tracedty == targetty
   Evaluated: Reactant.TracedRArray{T, N} == Reactant.TracedRArray
```

my understanding is that this is a bug and that we shouldn't return the unwrapped UnionAll.

test/tracing.jl
mofeing: that's fixed now

PR #2309, opened 2026-01-30
fix: convert AbstractRange to concrete types (#2276)

### LINKED ISSUE #2276: AbstractRange not converted to Concrete Types
## Background

In my workflow, I encountered an inconsistency where using `reshape(range(...))` produces an array of type `Base.ReshapedArray{Float32, 2, StepRangeLen{Float32, Float64, Float64, Int64}, Tuple{}}`, while the test cases typically use `ConcretePJRTArray{Float32, 2, 1}`. Example code:

```julia
y_data = reshape(range(1f0, 10f0, step=k_STEP), 1, EVAL_POINTS)
y_data = y_data |> xdev
```

This results in a `Base.ReshapedArray{Float32, 2, StepRangeLen{Float32, Float64, Float64, Int64}, Tuple{}}`. My code previously compiled (albeit slowly), but now it seems to hit a problem that the test cases miss due to the use of concrete arrays.

From https://discourse.julialang.org/t/slow-reactant-compilation-for-modified-deeponet/135133/8

## Suggestion

Add logic to convert `AbstractRange` types into a concrete type that can then be safely passed downstream and converted to the respective traced type. This will help support more general array usage, consistent with regular Julia workflows, and ensure the test cases match real-world usage.

KookiesNKareem:
Fixes #2276.

- Adds support for converting `AbstractRange` types (e.g., `range()`, `reshape(range(...))`) to `ConcreteRArray` when using `to_rarray` or `Adapt.adapt`
- Added `AbstractArray` constructors for `ConcretePJRTArray` and `ConcreteIFRTArray`
- Added `to_rarray_internal` methods for `AbstractArray` types
- Added `Adapt.adapt_structure` overrides for range types
- Added tests for AbstractRange conversion

### DIFF
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -1,8 +1,39 @@
 using Reactant, Test, Enzyme, Statistics, InteractiveUtils
+using Adapt: adapt
 
 const RunningOnTPU = contains(string(Reactant.devices()[1]), "TPU")
 
 fastmax(x::AbstractArray{T}) where {T} = reduce(max, x; dims=1, init=float(T)(-Inf))
+
+@testset "AbstractRange conversion to ConcreteRArray" begin
+    # Test basic range conversion (issue #2276)
+    r = range(1.0f0, 10.0f0; length=10)
+    a = Reactant.to_rarray(r)
+    @test a isa ConcreteRArray{Float32,1}
+    @test Array(a) ≈ collect(r)
+
+    # Test StepRangeLen conversion
+    r_step = range(1.0f0, 10.0f0; step=0.5f0)
+    a_step = Reactant.to_rarray(r_step)
+    @test a_step isa ConcreteRArray{Float32,1}
+    @test Array(a_step) ≈ collect(r_step)
+
+    # Test reshaped range conversion (the original issue case)
+    r_reshaped = reshape(range(1.0f0, 10.0f0; length=10), 2, 5)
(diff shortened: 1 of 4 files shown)


test/basic.jl
```diff
@@ -1,9 +1,40 @@
...
 const RunningOnTPU = contains(string(Reactant.devices()[1]), "TPU")
 
 fastmax(x::AbstractArray{T}) where {T} = reduce(max, x; dims=1, init=float(T)(-Inf))
 
+@testset "AbstractRange conversion to ConcreteRArray" begin
+    # Test basic range conversion (issue #2276)
+    r = range(1.0f0, 10.0f0; length=10)
+    a = Reactant.to_rarray(r)
+    @test a isa ConcreteRArray{Float32,1}
+    @test Array(a) ≈ collect(r)
```
avik-pal, line 13: This is not the right solution. Eagerly materializing any structured data structure will hurt perf really bad

KookiesNKareem: Ahh, good point. Would limiting to AbstractRange only also cause performance degradation or no?

wsmoses: What is the context within which we need the support (eg the root cause of the issue)?

Eg are we materializing a large constant?

KookiesNKareem: I updated to limit only to AbstractRange if this support is needed at all.

PR #2310, opened 2026-01-30
test: use ParallelTestRunner

avik-pal:
- [x] migrate to ParallelTestRunner
- [x] limit memory on TPU/CUDA runners?


giordano:
There are some improvements:

* I strongly recommend doing https://github.com/EnzymeAD/Enzyme.jl/blob/2366bed57374ebb8b9adb3910ddf183d8df65dcc/.github/workflows/CI.yml#L147-L157 otherwise tests run serially on macOS (also, making output more verbose is more informative to see what tests are being run)
* all the filtering of the tests should be guarded by filter_tests!, see https://github.com/EnzymeAD/Enzyme.jl/blob/2366bed57374ebb8b9adb3910ddf183d8df65dcc/test/runtests.jl#L19 as example, otherwise you can't choose the tests with the script arguments. This is also clearly documented: https://juliatesting.github.io/ParallelTestRunner.jl/stable/advanced/#Filtering-Test-Files

PR #2328, opened 2026-02-01
Add make_tracer for BitArray (Fix #2327)

### LINKED ISSUE #2327: BitArrays
Hi,

Currently I get the following when trying to include `BitArray`s in tracing:
```
using Reactant
x = [0, 1, 2]
mask = x .< 2
# Reactant.to_rarray(Float32.(mask))  # Current workaround (works)
Reactant.to_rarray(mask) # fails
```
This is on  the latest Reactant v0.2.210
Stacktrace:
```
NoFieldMatchError: 
Cannot convert type BitVector, best attempt BitVector failed.
This could be because the type does not capture the fieldtypes that should be converted in its type parameters.
name=chunks idx=1 Derived: ConcretePJRTArray{UInt64, 1, 1} Existing: Vector{UInt64} Best Attempt: Vector{UInt64}
name=len idx=2 Derived: Int64 Existing: Int64 Best Attempt: Int64
name=dims idx=3 Derived: Tuple{Int64} Existing: Tuple{Int64} Best Attempt: Tuple{Int64}


Stacktrace:
 [1] traced_type_inner(T::Type, seen::Dict{Type, Type}, mode::Reactant.TraceMode, track_numbers::Type, ndevices::Any, runtime::Any)
   @ Reactant ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:793
 [2] traced_type(T::Type, ::Val{Reactant.ArrayToConcrete}, track_numbers::Type, sharding::Reactant.Sharding.NoSharding, runtime::Val{:PJRT})
   @ Reactant ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:1001
 [3] make_tracer_unknown(seen::Reactant.OrderedIdDict{Any, Any}, prev::Any, path::Any, mode::Reactant.TraceMode; track_numbers::Type, sharding::Any, runtime::Any, kwargs::@Kwargs{device::Nothing, client::Nothing})
   @ Reactant ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:1163
 [4] make_tracer_unknown
   @ ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:1140 [inlined]
 [5] #make_tracer#128
   @ ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:1321 [inlined]
 [6] to_rarray_internal(x::Any, track_numbers::Type, sharding::Any, runtime::Any, device::Any, client::Any)
   @ Reactant ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:2081
 [7] #to_rarray#153
   @ ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:2070 [inlined]
 [8] to_rarray(x::Any)
   @ Reactant ~/.julia/packages/Reactant/EhaVb/src/Tracing.jl:2060
 [9] top-level scope
```

niklasschmitz:
Fixes #2327 .

I'm not familiar yet with the Reactant.jl internals, but @wsmoses ' suggestion in https://github.com/EnzymeAD/Reactant.jl/issues/2327#issuecomment-3830994088  seemed concrete enough to try to pattern-match the existing rules with an AI assistant.

I can confirm it does fix my issue in #2327, but if looks not right or there is a better way, feel free to just close  :+1:

### DIFF
--- a/test/core/tracing.jl
+++ b/test/core/tracing.jl
@@ -122,6 +122,10 @@ end
                 (AbstractVector, AbstractVector, AbstractVector),
                 (AbstractMatrix, AbstractMatrix, AbstractMatrix),
 
+                # BitArray types
+                (BitArray, BitArray, BitArray),
+                (BitArray{1}, BitArray{1}, BitArray{1}),
+
                 # Union types
                 (Union{Nothing,Int}, Union{Nothing,Int}, Union{Nothing,TracedRNumber{Int}}),
                 (


src/Tracing.jl
```diff
@@ -512,6 +512,30 @@ Base.@nospecializeinfer function traced_type_inner(
...
 
+Base.@nospecializeinfer function traced_type_inner(
+    @nospecialize(A::Type{<:BitArray}),
+    seen,
+    mode::TraceMode,
+    @nospecialize(track_numbers::Type),
+    @nospecialize(ndevices),
+    @nospecialize(runtime)
+)
+    N = ndims(A)
```
avik-pal: for simplicity, can we forward the call to traced_type_inner with Array{Bool,N}

src/Tracing.jl
```diff
@@ -1823,6 +1847,44 @@ Base.@nospecializeinfer function make_tracer(
...
+    @nospecialize(path),
+    mode;
+    @nospecialize(track_numbers::Type = Union{}),
+    @nospecialize(sharding = Sharding.NoSharding()),
+    @nospecialize(runtime = nothing),
+    @nospecialize(device = nothing),
+    @nospecialize(client = nothing),
+    kwargs...,
+)
+    if mode != NoStopTracedTrack && haskey(seen, prev)
```
avik-pal: same here

niklasschmitz:
Thanks a lot @avik-pal, I was also unsure about the previous code duplication, your change looks much better indeed!

src/Tracing.jl
```diff
@@ -512,6 +512,19 @@ Base.@nospecializeinfer function traced_type_inner(
...
+    @nospecialize(A::Type{<:BitArray}),
+    seen,
+    mode::TraceMode,
+    @nospecialize(track_numbers::Type),
+    @nospecialize(ndevices),
+    @nospecialize(runtime)
+)
+    return traced_type_inner(
+        Array{Bool,ndims(A)}, seen, mode, track_numbers, ndevices, runtime
+    )
```
mofeing: this is an edge case, but I just had to fix some stuff related to UnionAlls so do you mind adding the case in which `ndims` is not defined? (i.e. the type is `BitArray` and not `BitArray{2}` for example)


```suggestion
    return traced_type_inner(
        A isa UnionAll ? Array{Bool} : Array{Bool,ndims(A)}, seen, mode, track_numbers, ndevices, runtime
    )
```

src/Tracing.jl
```diff
@@ -1823,6 +1836,12 @@ Base.@nospecializeinfer function make_tracer(
     return newa
 end
 
+Base.@nospecializeinfer function make_tracer(
+    seen, @nospecialize(prev::BitArray), @nospecialize(path), mode; kwargs...
+)
+    return make_tracer(seen, Array{Bool,ndims(prev)}(prev), path, mode; kwargs...)
```
mofeing: the problem with calling `Array(prev)` here is that you will also allocate an `Array` on cases where you don't need to. you should only call `Array(prev)` depending on the `mode`.


```suggestion
    if mode == ArrayToConcrete
        return make_tracer(seen, Array(prev), path, mode; kwargs...)
    else
        return prev
    end
```

src/Tracing.jl
mofeing: this breaks compilation of `BitArray` when considered as "constant" (i.e. not "traced") for partial evaluation.


```suggestion
    if mode == ArrayToConcrete
        A´ = A isa UnionAll ? Array{Bool} : Array{Bool,ndims(A)}
        return traced_type_inner(
            A´, seen, mode, track_numbers, ndevices, runtime
        )
    else
        return A
    end
```

test/core/tracing.jl
```diff
@@ -377,6 +377,28 @@ end
...
+
+        struct BitVecStorage{B}
+            v::B
+        end
+
+        mask_struct = BitVecStorage(mask)
+        mask_struct_ra = Reactant.to_rarray(mask_struct)
+        @test mask_struct_ra isa BitVecStorage{<:ConcreteRArray{Bool,1}}
+        @test mask_struct_ra.v isa ConcreteRArray{Bool,1}
+        @test Array(mask_struct_ra.v) == Bool[true, true, false]
```
mofeing, line 403: mmm why this?

test/core/tracing.jl
```diff
@@ -377,6 +377,28 @@ end
         @test @inferred Reactant.to_rarray(x_ra) isa ConcreteRNumber
     end
 
+    @testset "BitArray" begin
+        mask = BitVector([true, true, false])
+        mask_ra = Reactant.to_rarray(mask)
+        @test mask_ra isa ConcreteRArray{Bool,1}
+        @test Array(mask_ra) == Bool[true, true, false]
```
mofeing, line 388: I would add this tests to the "specialized dispatches" test set just above

mofeing (changes requested):
The way this is coded right now actually breaks `BitArray` on the case that it's treated as a "constant" (i.e. not traced).

For example, on main the behavior is this:

```julia
julia> x = [0, 1, 2]
3-element Vector{Int64}:
 0
 1
 2

julia> mask = x .< 2
3-element BitVector:
 1
 1
 0

julia> f(::BitArray) = "a bitarray"

julia> f(x) = "not a bitarray"
f (generic function with 2 methods)

julia> @jit f(mask)
"a bitarray"
```

while on your PR, the behavior is the following

```julia
julia> @jit f(mask)
"not a bitarray"
```

Furthermore, `make_tracer` is called in multiple occasions and calling `Array(prev)` always will provoke to allocate multiple times an `Array` that won't be used except once.

What you should do instead is to add a conditional checking if the mode is `ArrayToConcrete`. Otherwise, return itself.

test/core/tracing.jl
avik-pal: otherwise the trace_type_inner wont get triggered

niklasschmitz:
Many thanks @avik-pal and @mofeing !

PR #2334, opened 2026-02-02
test(TPU): only use single worker

avik-pal:


test/runtests.jl
```diff
@@ -60,6 +60,10 @@ if !NUMPYRO_INSTALLED[]
     delete!(testsuite, "integration/numpyro")
 end
 
+if Reactant.Accelerators.TPU.has_tpu() || BACKEND == "tpu"
+    @set! parsed_args.jobs = 1
```
wsmoses: so one other issue to be cognisant of is that i think the test is run in a different process. That's a problem since then the runtests process may have a tpu and the job itself does not

test/Project.toml
```diff
@@ -32,6 +32,7 @@ PythonCall = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Random123 = "74087812-796a-5b5d-8853-05524746bad3"
 Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
+Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
```
giordano, line 35: Compat bound

test/runtests.jl
avik-pal: I made sure to not trigger any of the XLA APIs in this process

wsmoses: sg

wsmoses (approved):

avik-pal:
:confused: 

```
Error During Test at /__w/Reactant.jl/Reactant.jl/test/core/qa.jl:61
  Got exception outside of a @test
  ArgumentError: Malformed UUID string: "AssertionError(\"The TPU is already in use by process with pid 19095. Not attempting to load libtpu.so in this process.\")"
  Stacktrace:
    [1] (::Base.var"#throw_malformed_uuid#throw_malformed_uuid##0")(s::String)
      @ Base ./uuid.jl:82
    [2] parse
      @ ./uuid.jl:85 [inlined]
    [3] UUID
      @ ./uuid.jl:89 [inlined]
    [4] iterate
      @ ./generator.jl:48 [inlined]
    [5] grow_to!(dest::Vector{Base.UUID}, itr::Base.Generator{Base.EachLine{IOBuffer}, Type{Base.UUID}}, st::Nothing)
      @ Base ./array.jl:894
    [6] grow_to!(dest::Vector{Base.UUID}, itr::Base.Generator{Base.EachLine{IOBuffer}, Type{Base.UUID}})
      @ Base ./array.jl:867
    [7] collect
      @ ./array.jl:787 [inlined]
```

giordano:
Doesn't the initialisation of the backend happen upon loading the package?

avik-pal:
It should happen on triggering any kind of XLA api

avik-pal:
So `Aqua` loads the package in a separate process which causes this mess. I will pull out the qa tests to run after the main tests

PR #2335, opened 2026-02-02
Define new methods for FFT plans with keyword arguments

giordano:
This lets us support methods defined in `FFTW.jl` which have keyword arguments.  Ref: https://github.com/CliMA/Oceananigans.jl/issues/5244.

CC @ptiede @dkytezab

### DIFF
--- a/ext/ReactantAbstractFFTsExt.jl
+++ b/ext/ReactantAbstractFFTsExt.jl
@@ -155,7 +155,9 @@ for op in (:rfft, :fft, :ifft)
     end
     @eval $(plan_name){T}(dims) where {T} = $(plan_name){T,typeof(dims)}(dims)
 
-    @eval function AbstractFFTs.$(plan_f)(x::AnyTracedRArray{T}, dims=1:ndims(x)) where {T}
+    @eval function AbstractFFTs.$(plan_f)(
+        x::AnyTracedRArray{T}, dims=1:ndims(x); _kwargs...
+    ) where {T}
         return $(plan_name){T,typeof(dims)}(dims)
     end
 


ext/ReactantAbstractFFTsExt.jl
```diff
@@ -179,6 +183,15 @@ for op in (:rfft, :fft, :ifft)
...
             return $(plan_name!){T,typeof(dims)}(dims)
         end
+        # This method with `AnyStridedTracedRArray` is needed to extend methods defined in
+        # `FFTW.jl` for `StridedArray`s which have keyword arguments (which we'll ignore
+        # here).  The body of the method is the same as the above one, but we can't use a
+        # `Union` because it'd still be less specific than the FFTW's methods.
+        @eval function AbstractFFTs.$(plan_f!)(
+            x::AnyStridedTracedRArray{T}, dims=1:ndims(x); _kwargs...
+        ) where {T}
+            return $(plan_name!){T,typeof(dims)}(dims)
```
avik-pal, line 195: Should we warn here that kwargs are being ignored (if anything apart from default was passed)?

giordano: I'm a bit on the fence about this: I understand why this may be useful, but then when you run distributed applications you get hundreds and hundreds of warnings you can't do anything about which only clutter the log files.

wsmoses: I'm okay not warning here, time limit and numthreads aren't semantically meaningful for us anyways

giordano:
Tests should pass now:
```
Running 1 tests in parallel. If this is too many, specify the `--jobs=N` argument to the tests, or set the `JULIA_CPU_THREADS` environment variable.
                      │          │ ──────────────── CPU ──────────────── │
Test         (Worker) │ Time (s) │ GC (s) │ GC % │ Alloc (MB) │ RSS (MB) │
integration/fft   (1) │   159.90 │   2.81 │  1.8 │   17127.45 │  1653.03 │

Test Summary: | Pass  Total     Time
  Overall     |  179    179  2m45.6s
    SUCCESS
Test Summary:  | Total     Time
Reactant Tests |     0  2m47.1s
     Testing Reactant tests passed
```

wsmoses:
with the great github lack of ci for now, I'll trust you as things come back alive

PR #2340, opened 2026-02-03
add `traced_type_inner` case for `Memory`

### LINKED ISSUE #2301: `Memory{T}` causes TypeError in `@compile` / `make_tracer_unknown`
I am currently going through our model types in SpeedyWeather to find problems with Reactant. This one is actually used / a part of `ProgessMeter` from ProgressMeter.jl that we use as well as part of our model types 

MWE 

```Julia 
using Reactant 

struct Foo 
    bar::Memory{UInt8}
end 

foo = Foo(Memory{UInt8}(undef, 1))

do_nothing(foo::Foo) = nothing 

@compile do_nothing(foo)
```

```
ERROR: TypeError: in new, expected DataType, got Type{Memory{UInt8}}
Stacktrace:
  [1] make_tracer_unknown(seen::Reactant.OrderedIdDict{…}, prev::Any, path::Any, mode::Reactant.TraceMode; track_numbers::Type, sharding::Any, runtime::Any, kwargs::@Kwargs{…})
    @ Reactant ~/.julia/dev/Reactant/src/Tracing.jl:1159
  [2] make_tracer_unknown
    @ ~/.julia/dev/Reactant/src/Tracing.jl:1121 [inlined]
  [3] #make_tracer#128
    @ ~/.julia/dev/Reactant/src/Tracing.jl:1303 [inlined]
  [4] make_tracer_unknown(seen::Reactant.OrderedIdDict{…}, prev::Any, path::Any, mode::Reactant.TraceMode; track_numbers::Type, sharding::Any, runtime::Any, kwargs::@Kwargs{…})
    @ Reactant ~/.julia/dev/Reactant/src/Tracing.jl:1203
  [5] make_tracer_unknown
    @ ~/.julia/dev/Reactant/src/Tracing.jl:1121 [inlined]
  [6] #make_tracer#128
    @ ~/.julia/dev/Reactant/src/Tracing.jl:1303 [inlined]
  [7] make_tracer
    @ ~/.julia/dev/Reactant/src/Tracing.jl:1293 [inlined]
  [8] prepare_mlir_fn_args(args::Tuple{…}, name::String, concretein::Bool, toscalar::Bool, argprefix::Symbol, runtime::Val{…}, optimize_then_pad::Bool, do_transpose::Bool, input_shardings::Nothing, verify_arg_names::Nothing)
    @ Reactant.TracedUtils ~/.julia/dev/Reactant/src/TracedUtils.jl:460
  [9] make_mlir_fn(f::typeof(do_nothing), args::Tuple{…}, kwargs::@NamedTuple{}, name::String, concretein::Bool; toscalar::Bool, return_dialect::Symbol, args_in_result::Symbol, construct_function_without_args::Bool, do_transpose::Bool, within_autodiff::Bool, input_shardings::Nothing, output_shardings::Nothing, runtime::Val{…}, verify_arg_names::Nothing, argprefix::Symbol, resprefix::Symbol, resargprefix::Symbol, num_replicas::Int64, optimize_then_pad::Bool)
    @ Reactant.TracedUtils ~/.julia/dev/Reactant/src/TracedUtils.jl:331
 [10] make_mlir_fn
    @ ~/.julia/dev/Reactant/src/TracedUtils.jl:284 [inlined]
 [11] compile_mlir!(mod::Reactant.MLIR.IR.Module, f::typeof(do_nothing), args::Tuple{…}, compile_options::CompileOptions, callcache::Dict{…}, sdycache::Dict{…}, sdygroupidcache::Tuple{…}; fn_kwargs::@NamedTuple{}, backend::String, runtime::Val{…}, legalize_stablehlo_to_mhlo::Bool, client::Reactant.XLA.PJRT.Client, kwargs::@Kwargs{})
    @ Reactant.Compiler ~/.julia/dev/Reactant/src/Compiler.jl:1759
 [12] compile_mlir!
    @ ~/.julia/dev/Reactant/src/Compiler.jl:1721 [inlined]
 [13] compile_xla(f::Function, args::Tuple{…}; before_xla_optimizations::Bool, client::Nothing, serializable::Bool, kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/.julia/dev/Reactant/src/Compiler.jl:3727
 [14] compile_xla
    @ ~/.julia/dev/Reactant/src/Compiler.jl:3700 [inlined]
 [15] compile(f::Function, args::Tuple{…}; kwargs::@Kwargs{…})
    @ Reactant.Compiler ~/.julia/dev/Reactant/src/Compiler.jl:3815
 [16] top-level scope
    @ ~/.julia/dev/Reactant/src/Compiler.jl:2870
```

mofeing:
fixes #2301

### DIFF
--- a/test/core/tracing.jl
+++ b/test/core/tracing.jl
@@ -188,6 +188,7 @@ testsuite = [
     (Ptr, Ptr, Ptr),
     (Core.LLVMPtr, Core.LLVMPtr, Core.LLVMPtr),
     (Base.RefValue, Base.RefValue, Base.RefValue),
+    (Ref, Ref, Ref),
 
     # Function types
     (


src/Tracing.jl
```diff
@@ -577,9 +569,7 @@ Base.@nospecializeinfer function traced_type_inner(
     @nospecialize(ndevices),
     @nospecialize(runtime)
 ) where {T,A}
-    return Core.LLVMPtr{
-        traced_type_inner(PT.parameters[1], seen, mode, track_numbers, ndevices, runtime),A
-    }
+    return Core.LLVMPtr{traced_type_inner(T, seen, mode, track_numbers, ndevices, runtime),A}
```
mofeing: @wsmoses I've taken the freedom to refactor these lines to avoid accessing internal fields. I guess there's no problem accessing `T` this way.

mofeing: actually, I think the `@nospecialize` on the type argument is doing nothing on these cases.

test/core/tracing.jl
```diff
@@ -208,6 +208,13 @@ end
                     Base.RefValue{TracedRArray{Float64,1}},
                     Base.RefValue{TracedRArray{Float64,1}},
                 ),
+                (Memory{UInt8}, Memory{UInt8}, Memory{TracedRNumber{UInt8}}),
```
mofeing:
```suggestion
                (Base.Memory{UInt8}, Base.Memory{UInt8}, Base.Memory{TracedRNumber{UInt8}}),
```

test/core/tracing.jl
```diff
@@ -208,6 +208,13 @@ end
...
+
+                # Ptr + UnionAll types
+                (Ptr, Ptr, Ptr),
+                (Core.LLVMPtr, Core.LLVMPtr, Core.LLVMPtr),
+                (Base.RefValue, Base.RefValue, Base.RefValue),
+                (Memory, Memory, Memory),
```
mofeing:
```suggestion
                (Base.Memory, Base.Memory, Base.Memory),
```

test/core/tracing.jl
```diff
@@ -208,6 +208,13 @@ end
                     Base.RefValue{TracedRArray{Float64,1}},
                     Base.RefValue{TracedRArray{Float64,1}},
                 ),
+                (Base.Memory{UInt8}, Base.Memory{UInt8}, Base.Memory{TracedRNumber{UInt8}}),
```
wsmoses: I'm skeptical of this, we presumably don't want a memory of a tracedrnumber, but a concreterarray ?

mofeing: yeah... I have doubts on what are the kind of semantics we want here.

I think the conservative move here is to either error or do nothing if `T <: track_numbers` for `Memory{T}`, because `Memory` is thought to be wrapped by an array type, and that array type is what we should be really converting.

the issue in #2301 is because `ProgressMeter` uses it but we are not going to trace over a `ProgressMeter` (I hope), so I guess we don't need support for `track_numbers` here.

src/Tracing.jl
```diff
@@ -598,6 +598,19 @@ Base.@nospecializeinfer function traced_type_inner(
     }
 end
 
+@static if VERSION >= v"1.11"
```
mofeing:
```suggestion
@static if isdefined(Core, :Memory)
```

mofeing:
`traced_type_inner` behaviour of `Memory` is under discussion and haven't reached consensus yet. also, the issue this PR tries to solve doesn't require a `traced_type_inner` rule.

PR #2346, opened 2026-02-03
Don't free buffers after clients are destroyed

wsmoses:


src/xla/XLA.jl
```diff
@@ -105,10 +105,14 @@ else
 end
 const global_state = State()
 
+const is_live = Ref(true)
```
giordano: Should this be an atomic?

wsmoses: how does one do that properly here

giordano:
```julia-repl
julia> const is_live = Threads.Atomic{Bool}(true)
Base.Threads.Atomic{Bool}(true)

julia> is_live[]
true

julia> is_live[] = false
false

julia> is_live[]
false
```

giordano: Basically only change
```suggestion
const is_live = Threads.Atomic{Bool}(true)
```

wsmoses:
conda issues appear unrelated and confirmed gb fix, going to merge and release

PR #2362, opened 2026-02-06
Adding an Oceananigans benchmark

jlk9:
This adds an Oceananigans model run to the Reactant benchmark.

### DIFF
--- a/benchmark/setup.jl
+++ b/benchmark/setup.jl
@@ -121,7 +121,8 @@ Run all benchmark subdirectories in separate processes and aggregate results.
 """
 function run_all_benchmarks(backend::String)
     base_dir = dirname(@__FILE__)
-    benchmark_dirs = find_benchmark_dirs(base_dir)
+    # benchmark_dirs = find_benchmark_dirs(base_dir)
+    benchmark_dirs = [joinpath(base_dir, "oceananigans")] # XXX: remove
 
     @info "Found $(length(benchmark_dirs)) benchmark directories: $(basename.(benchmark_dirs))"
 


jlk9:
This adds a sample Oceananigans script to the benchmark directory for CI runs. I'm not sure how to update the 'benchmark.yml' file so that this script is run. I know the benchmark directory is listed here:

https://github.com/EnzymeAD/Reactant.jl/blob/9227d57f248250440ef77a25d954d9a3b629e1a2/.github/workflows/benchmark.yml#L16C1-L23C1

but none of the subdirectories are.

Also, I attached the script's dependencies in a separate `Project.toml` file in the new `benchmark/oceananigans` subdirectory. Maybe those dependencies should just go into the toml in the benchmark folder itself - let me know if that's better.

@avik-pal @wsmoses

benchmark/oceananigans/abernathey_channel.jl
```diff
@@ -0,0 +1,449 @@
+#using Pkg
+# pkg"add Oceananigans CairoMakie"
+using Oceananigans
+ENV["GKSwstype"] = "100"
+
+pushfirst!(LOAD_PATH, @__DIR__)
```
giordano: This looks unnecessary.  In general, I've never found a good reason for using `LOAD_PATH`, [it's only a source of troubles](https://github.com/EnzymeAD/Enzyme.jl/pull/2265#discussion_r1919393652)
```suggestion
```

jlk9:
Additionally, checkpointing is set for this for loop:
https://github.com/EnzymeAD/Reactant.jl/blob/9227d57f248250440ef77a25d954d9a3b629e1a2/benchmark/oceananigans/abernathey_channel.jl#L280C1-L286C4

To enable/disable checkpointing you need to change the `checkpointing = true` line there.

wsmoses:
```
- Run `import Pkg; Pkg.add("Chairmarks")` to install the Chairmarks package.
```

wsmoses:
@jlk9 seems like a setup error?

```

ERROR: LoadError: MethodError: no method matching Duplicated(::Reactant.TracedRArray{Float64, 2}, ::Matrix{Float64})
The type `Duplicated` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  Duplicated(::T1, ::T1) where T1
   @ EnzymeCore ~/.julia/packages/EnzymeCore/RpjpI/src/EnzymeCore.jl:68
  Duplicated(::T1, ::T1, ::Bool) where T1
   @ EnzymeCore ~/.julia/packages/EnzymeCore/RpjpI/src/EnzymeCore.jl:68
  Duplicated(::T1, ::T1) where T1<:SubArray
   @ EnzymeCore ~/.julia/packages/EnzymeCore/RpjpI/src/EnzymeCore.jl:69
  ...

Stacktrace:
  [1] macro expansion
    @ /__w/Reactant.jl/Reactant.jl/src/utils.jl:0 [inlined]
  [2] call_with_reactant(::Reactant.EnsureReturnType{Union{}}, ::Type{Duplicated}, ::Reactant.TracedRArray{Float64, 2}, ::Matrix{Float64})
    @ Reactant /__w/Reactant.jl/Reactant.jl/src/utils.jl:1177
  [3] call_with_reactant(::typeof(Main.AbernatheyChannel.differentiate_tracer_error), 
```

avik-pal:
That's a screwup from me

benchmark/oceananigans/abernathey_channel.jl
```diff
@@ -0,0 +1,462 @@
...
+    return nothing
+end
+
+#####
+##### Forward simulation (not actually using the Simulation struct)
+#####
+
+function loop!(model)
+    Δt = model.clock.last_Δt
+    @trace mincut = true checkpointing = true track_numbers = false for i in 1:100
```
wsmoses, line 270: potentially for a follow up, @avik-pal we should try testing checkpointing on vs off, and also optimizations on vs off

jlk9:
@avik-pal I think the benchmark has too many timesteps in the autodiff run. Locally having 10 steps in `loop!` passes but 100 (what it is in this PR now) fails. I'm testing if 10 passes on the reduced benchmark.

PR #2290, opened 2026-01-28
refactor ownership in MLIR.IR

mofeing:
refactor from #2093

## to do

- [x] refactor `lose_ownership!` to something like `mark_transfer` _alas_ LLVM.jl
  - right now it does nothing, but will properly add debug msgs on subsequent PR
- [x] add `dispose` calls or rely on scopes
- [x] refactor `with_*` functions to use macro-based scopes
- [x] add `@activate` macro
- [x] add `ctx::MLIR.IR.Context` arg to `compile` and `compile_*` functions
- [x] stop returning MLIR Module in `compile_xla` (it was getting leaked)
- [x] refactor macros to construct and dispose `Context`

PR #2365, opened 2026-02-06
refactor MLIR.IR ownership to manual disposal

mofeing:
#2290 turned quite big again... so I splitted up this to ease reviewing. this PR leaves leaks unresolved, but prepares the code for fixing them.

### to MLIR.jl

- refactor MLIR.IR types to use `LLVM.@checked` and `LLVM.refcheck`
- remove `owned` field from `Context`, `Operation`, `Block`, `Module` and refactor to use `LLVM.dispose` on them
  - it also reuses `mark_alloc`, `mark_dealloc` and `mark_use` to detect leaks on debugging sessions with memcheck enabled
  - `lose_ownership!` has been renamed to `mark_donate` which currently is a noop but in the future I would like to add some specific behaviour to the memcheck
- refactor to use `LLVM.activate`, `LLVM.deactivate`, `LLVM.dispose` and `LLVM.@dispose`

### to Compiler.jl

- fix MLIR.IR.Module leak from `@code_hlo` and `@code_mhlo`
- introduce `code_hlo`, `code_mhlo`, `code_xla` for functional version
  - required to return Module within a Context without leaking
- refactor macros (`@code_hlo`, `@compile`, `@jit`, ...) to construct and dispose a Context without leaking

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -2842,7 +2842,7 @@ $(COMMON_COMPILE_OPTIONS_DOCS)
 See also [`@code_xla`](@ref), [`@code_mhlo`](@ref).
 """
 macro code_hlo(args...)
-    (; f, args, kwargs, options) = parse_macro_code_call(
+    (; f, args, kwargs, options) = parse_call_expr(
         merge(get_common_compile_options(), Dict{Symbol,Any}(:shardy_passes => :(:none))),
         args...,
     )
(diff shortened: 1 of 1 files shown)


wsmoses:
@gbaraldi can you take a look?

mofeing:
#2374 fixes the failing tests

mofeing:
@wsmoses unfortunately #2374 doesn't work due to data-races on disposal... it's the chicken & egg problem: either #2374 fixes the leak on macros but runs onto the finalization data-races due to the GC, or this PR errors due to the the leaks on macros. there's no way to split them up so I added the fixes here.

about the `with_*` functions, I've replaced the ones I introduced for `try` blocks or `@dispose`. I plan to deprecate these functions and replace their callers all over our codebase on a subsequent PR.

wsmoses (approved):

PR #2368, opened 2026-02-07
Add new patterns

wsmoses:


wsmoses:
@giordano @avik-pal did something change in the runner setup?

giordano:
https://github.com/EnzymeAD/Reactant.jl/actions/runs/21777224352 it says that the job was cancelled. Might be a github problem, although https://www.githubstatus.com/ is green, yesterday I also had a job cancelled seemingly for no reason.

giordano:
Ah, it also says "The hosted runner lost communication with the server". That definitely sounds like a github problem

giordano:
Uhm, but also

> Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

🤔

PR #2374, opened 2026-02-08
fix Module leak in macros

mofeing:
required for #2365 tests to pass, but didn't want to make that PR bigger so i'm splitting that here.

the main change is that now `compile`, `compile_mlir` and `compile_xla` require a MLIR Context as its first argument (but i keep the methods without explicit context passing for backward compat). this is required for the cases in which they return a MLIR object and the Context lifetime must match the one of the MLIR operation / module.

such is the case of the `@code_hlo` and `@code_mhlo` macros, but i had to change the code of `compile_call_expr` to make all compatible.

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -1558,7 +1558,9 @@ function compile_mlir(f, args; kwargs...)
     end
 end
 
-function compile_mlir(ctx, f, args; client=nothing, drop_unsupported_attributes=false, kwargs...)
+function compile_mlir(
+    ctx, f, args; client=nothing, drop_unsupported_attributes=false, kwargs...
+)
     client = client !== nothing ? client : XLA.default_backend()
     backend = XLA.platform_name(client)
 


src/Compiler.jl
```diff
@@ -1552,7 +1552,15 @@ function __get_compile_options_and_kwargs(;
     )
 end
 
-function compile_mlir(f, args; client=nothing, drop_unsupported_attributes=false, kwargs...)
+function compile_mlir(f, args; kwargs...)
+    MLIR.IR.with_context() do ctx
```
wsmoses, line 1556: because these are hot functions, we should avoid the closure here, can you use explicit activate/deactivate/finally here?

mofeing: sure. the idea whatsoever is to deprecate or directly remove the `with_*` functions like I did #2290, but I left that change for the successor PR of #2365 to keep it "reviewable"

wsmoses: yeah but we already have versions that don't use a closure already right?

mofeing: if you mean about the `@dispose`, I'm introducing it in #2365. unfortunately this PR doesn't work without the changes in #2365 😞 so I'm moving these changes to the other PR.

Issue #2375, opened 2026-02-08
Tracing through bitwise binary operations


dominic-chang:
I get the following bug if I try to trace through some bitwise binary operations

```julia
using Reactant

function test(x)
    return xor(x, 1)
end

x = Reactant.to_rarray([1, 2, 3])

@compile test.(x)
```

```julia
ERROR: TypeError: in RNumber, in T, expected T<:Union{Complex{Int16}, Complex{Int32}, Complex{Int64}, Complex{Int8}, Complex{UInt16}, Complex{UInt32}, Complex{UInt64}, Complex{UInt8}, Bool, Float16, Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8, Reactant.F8E4M3B11FNUZ, Reactant.F8E4M3FN, Reactant.F8E4M3FNUZ, Reactant.F8E5M2, Reactant.F8E5M2FNUZ, ComplexF64, ComplexF32}, got Type{Integer}
Stacktrace:
  [1] xor
    @ ~/.julia/dev/Reactant/src/TracedRNumber.jl:634
  [2] call_with_reactant(::Reactant.EnsureReturnType{Union{}}, ::typeof(xor), ::Reactant.TracedRNumber{Int64}, ::Int64)
    @ Reactant ~/.julia/dev/Reactant/src/utils.jl:1090
  [3] test
    @ ~/Software/JacobiElliptic.jl/test/reactant_test.jl:5
  [4] call_with_reactant(::Reactant.EnsureReturnType{Union{}}, ::typeof(test), ::Reactant.TracedRNumber{Int64})
    @ Reactant ~/.julia/dev/Reactant/src/utils.jl:1090
  [5] copy
    @ ~/.julia/dev/Reactant/src/TracedRArray.jl:283
  [6] call_with_reactant(::Reactant.EnsureReturnType{…}, ::typeof(copy), ::Base.Broadcast.Broadcasted{…})
    @ Reactant ~/.julia/dev/Reactant/src/utils.jl:1090
  [7] materialize
    @ ./broadcast.jl:903 [inlined]
  [8] BroadcastFunction
    @ ./broadcast.jl:1381 [inlined]
  [9] #apply#66
    @ ~/.julia/dev/Reactant/src/utils.jl:107
[...]

PR #2376, opened 2026-02-08
Fix incorrect parameterized type on bitwise binary operators

dominic-chang:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/2375


src/TracedRNumber.jl
```diff
@@ -585,64 +585,72 @@ function Base.:*(x::TwicePrecision, v::TracedRNumber)
     return result
 end
 
-for (T1, T2) in zip((Bool, Integer), (Bool, Integer))
-    T = promote_type(T1, T2)
+for T in (Bool, Integer)
```
wsmoses: we still need the int/bool versions [where binary has one vs the other type]

also add a test?

cc @avik-pal

src/TracedRNumber.jl
```diff
@@ -585,61 +585,69 @@ function Base.:*(x::TwicePrecision, v::TracedRNumber)
...
+        function Base.xor(x::A, y::TracedRNumber{B}) where {A<:$(T1),B<:$(T2)}  
+            C = promote_type(A, B)
             return @opcall xor(
-                Reactant.promote_to(TracedRNumber{$(T)}, x),
-                Reactant.promote_to(TracedRNumber{$(T)}, y),
+                Reactant.promote_to(TracedRNumber{C}, x),
+                Reactant.promote_to(TracedRNumber{C}, y),
             )
         end
         Base.:!(x::TracedRNumber{<:$(T1)}) = @opcall not(x)
```
wsmoses: this is accidentally called twice for each T1

wsmoses: might as well just have this be outside the for and have this as

Base.:!(x::TracedRNumber{<:Union{Bool, Integer}}) = @opcall not(x)

src/TracedRNumber.jl
```diff
@@ -585,61 +585,69 @@ function Base.:*(x::TwicePrecision, v::TracedRNumber)
     return result
 end
 
-for (T1, T2) in zip((Bool, Integer), (Bool, Integer))
-    T = promote_type(T1, T2)
+for (T1,T2) in ((Bool, Integer), (Integer, Bool), (Bool, Bool), (Integer, Integer))
```
wsmoses: zip from ealier is a bit simpler to retain

dominic-chang: The `zip` before didn't seem to generate all permutations. What about `(T1,T2) in ((i,j) for i in (Bool, Integer) for j in (Bool, Integer))`?

wsmoses: oh interesting, then I suppose indeed it didn't handle the mixed case earlier.

I'll defer reviewing this to @avik-pal who I believe did this part originally

src/TracedRNumber.jl
dominic-chang: Addressed in https://github.com/EnzymeAD/Reactant.jl/pull/2376/changes/6e6859e7527677c080d77f0f9fd409dca504dc5e

src/TracedRNumber.jl
avik-pal: We should do a `Iterators.product`

src/TracedRNumber.jl
dominic-chang: Which file is an appropriate place for the tests? Should I add it to [math_ops](https://github.com/EnzymeAD/Reactant.jl/blob/main/test/core/math_ops.jl) ?

avik-pal: Yeah

PR #2385, opened 2026-02-09
Try to work around nvtx injection path issue

wsmoses:


giordano:
Instead of doing this globally, how about wrapping `with_profile` with
```julia
withenv("NVTX_INJECTION64_PATH" => "") do
    # ....
end
```
?  Environment variables are terrible to handle nicely.

wsmoses:
> Instead of doing this globally, how about wrapping `with_profile` with
> 
> ```julia
> withenv("NVTX_INJECTION64_PATH" => "") do
>     # ....
> end
> ```
> 
> ? Environment variables are terrible to handle nicely.

I tried that, it failed. It needs to be set like this earlier [and can not be scoped at within profile]

PR #2386, opened 2026-02-09
Automatically log the julia functions within the profile activity

wsmoses:
@mofeing @avik-pal @jumerckx in principle we should be able to extend this idea to automatically also capture the backtrace and emit into mlir debug info without the expensive backtrace call

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -661,12 +661,12 @@ function traced_mi(mi)
 end
 
 function push_debug_stack!(fname::String, file::String, line::Int64)
-    push!(Reactant.Compiler.debugcache(), (; f_name=fname, file=file, line=line))
+    push!(Compiler.debugcache(), (; f_name=fname, file=file, line=line))
     return nothing
 end
 
 function pop_debug_stack!()
-    pop!(Reactant.Compiler.debugcache())
+    pop!(Compiler.debugcache())
     return nothing
 end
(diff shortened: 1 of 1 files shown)


src/Ops.jl
```diff
@@ -122,7 +122,16 @@ end
...
     # calling `stacktrace` can add a lot of time overhead, so let's avoid adding debug info if not used
     if !DEBUG_MODE[]
-        return MLIR.IR.Location(name, MLIR.IR.Location(file, line, 0))
+        init = MLIR.IR.Location(name, MLIR.IR.Location(file, line, 0))
+	res = init
+	if Reactant.Compiler._has_debugcache()
+		for stackframe in Base.reverse(Reactant.Compiler.debugcache())
+			res = MLIR.IR.Location(res, MLIR.IR.Location(
+			    string(stackframe.f_name), MLIR.IR.Location(stackframe.file, stackframe.line, 0)
+			    ))
```
mofeing: do we have a MLIR dump? I fear that this might create a huge amount of identical location annotations.

in such case we could create cache dict of `stackframe` to `Location`?

wsmoses: I'm manually inspecting the dumps. The locations are unique'd so there should be no issue

mofeing: great

PR #2391, opened 2026-02-09
refactor `hloSharding(To/From)TensorShardingAttr` to use MLIR-C types on the interface

mofeing:
while debugging for other PR, i noticed that `hloShardingToTensorShardingAttr` and `hloShardingFromTensorShardingAttr` functions in ReactantExtra use MLIR C++ types on an C-interface.

although no problems have appeared, this is risky.

this PR addresses that by correctly using MLIR-C types `MlirContext` and `MlirAttribute`.


wsmoses:
@avik-pal take a quick look, but lgtm otherwise

PR #2415, opened 2026-02-11
remove `with_context`, `with_module`, `with_block` MLIR resource managers

mofeing:
closures are problematic... we should use `MLIR.IR.@scope` instead activation during a certain scope.

(_open to change name_)


Pangoraw:
> closures are problematic..

could you elaborate why?

wsmoses:
so closures capture the state of all variables in scope, which can lead to some very unfortunate bugs when doing memory/gc management, they also add nontrivial overhead and extra harder to debug stackframes.

ext/ReactantCUDAExt.jl
```diff
@@ -1106,13 +1106,15 @@ Reactant.@reactant_overlay @noinline function (func::LLVMFunc{F,tt})(
...
+    wrapfunc = MLIR.IR.@scope MLIR.IR.body(mod) begin
+        MLIR.Dialects.llvm.func(;
+            sym_name,
+            sym_visibility=MLIR.IR.Attribute("private"),
+            function_type=wrapftype,
+            body=MLIR.IR.Region(),
+            CConv,
+        )
+    end
     push!(MLIR.IR.body(mod), wrapfunc)
```
Pangoraw: this line becomes redundant iiuc

src/mlir/IR/Module.jl
```diff
@@ -93,11 +93,7 @@ function current_module(; throw_error::Core.Bool=true)
...
 
-function with_module(f, blk::Module)
-    activate(blk)
-    try
-        f()
-    finally
-        deactivate(blk)
-    end
+function with_module(f, mod::Module)
+    depwarn("`with_module` is deprecated, use `@scope` instead.", :with_module)
```
wsmoses: Considering this is likely package internal, I think I'd be okay [and semver compliant] to remove without dep, if desired

mofeing:
@Pangoraw you can find some more context here https://github.com/JuliaLLVM/LLVM.jl/pull/309

ext/ReactantCUDAExt.jl
mofeing: right, I was trying to not use `@scope` here because it's only one line, but I changed my mind and forgot to remove it.

mofeing:
@wsmoses this is weird. I can confirm that...

1. sharding and optimize_comms tests only fail on PJRT, and
2. no PJRT code was modified here or in #2365

could it be that some bug was unveiled now?

wsmoses:
Uhhh that's weird, yeah idk what's up

PR #2398, opened 2026-02-10
Migrate deprecated JSON3 -> JSON

giordano:
[`JSON3.jl`](https://github.com/quinnj/JSON3.jl) has been deprecated for quite some time now.

PR #2419, opened 2026-02-11
perf: fix aggregation

avik-pal:


giordano:
At least this one shouldn't conflict with #2398

PR #2425, opened 2026-02-12
Testing reduced Oceananigans script. DO NOT MERGE

jlk9:
DO NOT MERGE. This is checking if reductions to the Oceananigans script in #2362 allow it to run with CI (particularly TPUs).


wsmoses:
okay @jlk9 / @avik-pal I've fixed this. The tests still take forever so we should consider reducing in scope perhaps [with a flag?] also perhaps adding options for turning checkpointing/optimizations for AD/etc off. but nevertheless now is hopefully unstuck

PR #2427, opened 2026-02-12
add `make_tracer` rule for `Memory`

### LINKED ISSUE #2301: `Memory{T}` causes TypeError in `@compile` / `make_tracer_unknown`

mofeing:
fixes #2301

@wsmoses i have doubts about the behavior it should have on different modes:

- on `ArrayToConcrete`, it maps to `ConcreteRArray{T,1}`.
- on `ConcreteToTraced` and other modes, i think it should map to itself.

also, unless we return itself always, I think we need a `traced_type_inner` rule.

### DIFF
--- a/src/Tracing.jl
+++ b/src/Tracing.jl
@@ -1952,8 +1952,36 @@ Base.@nospecializeinfer function make_tracer(
         end
         return nothing
     end
-
-    return seen[prev] = prev
+    TT = traced_type(eltype(RT), Val(mode), track_numbers, sharding, runtime)
+    newa = Memory{TT}(undef, length(prev))
+    seen[prev] = newa
+    same = true
+    for I in eachindex(prev)
+        if isassigned(prev, I)
+            pv = prev[I]
+            nv = make_tracer(
+                seen,
+                pv,
+                append_path(path, I),
+                mode;
+                track_numbers,
+                sharding=Base.getproperty(sharding, I),
+                runtime,
+                device,
+                client,
+                kwargs...,
+            )
+            if pv !== nv
+                same = false
+            end
+            @inbounds newa[I] = nv
+        end
+    end
+    if same
+        seen[prev] = prev
+        return prev
+    end
+    return newa
 end
 
 Base.@nospecializeinfer function make_tracer(


wsmoses:
yeah we'll also need traced_type_inner, but the specific error from the issue was the missing make_tracer

wsmoses:
concretetotraced we should do the same thing as array [aka if its a memory{T} we will recur on memory(make_tracer(T)), to handle things like memory{concreterarray} but leave alone memory{uint}

test/core/tracing.jl
```diff
@@ -272,6 +273,14 @@ end
         ),
         (Wrapper, Wrapper, Wrapper),
     ]
+
+    if isdefined(Core, :Memory)
+        push!(testsuite, (Memory{UInt8}, Memory{UInt8}, Memory{TracedRNumber{UInt8}}))
```
mofeing, line 278: @wsmoses in the other PR, I think you said that converting here to `TracedRNumber` needs to be discussed because it seems a too edgy case. If so, I agree, but right now it feels awkward that we treat Memory the same as Array except for this point.

PR #2429, opened 2026-02-12
chore: remove legacy testset nesting in tracing tests

mofeing:
while working in some PRs, this was bugging me because we no longer need a `@testset "Tracing"`.


mofeing:
there seems to be some connection problem with pkg.julialang.org. all tests are failing due to it.

wsmoses:
its not just this PR either

PR #2433, opened 2026-02-12
perf: run the benchmarks in parallel

avik-pal:


avik-pal:
The tflops update failure doesn't happen on master. we can investigate that separately but this PR is ready to go

wsmoses (commented):
how exactly are the benchmarks being run in parallel?

For example for CPU that seems like a bad thing since all threads will be consumed by a single benchmark execution, making runtime unreliable. For GPU/TPU we may be able to use the equivalent of CUDA_VISIBLE_DEVICES to partition amongst devices

avik-pal:
In completely independent runners. Previously it would hog up a single runner for an hour or so, now we use multiple runners if available and free those up much quicker.

wsmoses:
ah got it, yeah that's a lot better

PR #2436, opened 2026-02-13
Further debug info stripping

wsmoses:
@avik-pal we should turn on strip=:all during benchmarking, unless we want more info.

relatedly @mofeing @jumerckx @giordano @glwagner @ftynse we now get actionable profiling debug info, will make gb25 pr once this [and jll] lands

Screenshot 2026-02-13 at 12 06 24 AM
Screenshot 2026-02-13 at 12 07 38 AM

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Reactant"
 uuid = "3c362404-f566-11ee-1572-e11a4b42c853"
-version = "0.2.218"
+version = "0.2.219"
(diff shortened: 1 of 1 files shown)


wsmoses:
and also [indicating a copy of a field is made for fill north/south halo, @jumerckx / @ftynse we should inspect to see if the buffer stuff we separately discussed can resolve]:

Screenshot 2026-02-13 at 12 16 57 AM

mofeing:
something I don't understand from the code is what's the default behavior.

_to strip, or not to strip, that is my question_

wsmoses:
Default atm is to not strip

PR #2445, opened 2026-02-14
Add MPI Waitall support

romanlee:
Main change:

* Adds support for `MPI_Waitall`

Related minor fixes/changes:

* Use `TracedRNumber` instead of `TracedRequest` to represent `MPI_Request` objects. `TracedRequest` or similar will only be necessary if we want to allow requests to cross the compile boundary, which for now we don't.
* FIX: For ext/ReactantMPIExt/Ops.jl level and below, use `TracedRNumber{Int32}` instead of `TracedRNumber{Int64}` for`MPI_Requests`. Necessary for MPI ABI compatibility.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -98,18 +98,17 @@ total_jobs = min(
         end
     end
 
-    # TODO: reenable once fixed
-    # if (
-    #     isempty(parsed_args.positionals) ||
-    #     "integration" ∈ parsed_args.positionals ||
-    #     "integration/mpi" ∈ parsed_args.positionals
-    # )
-    #     @testset "MPI" begin
-    #         using MPI
-    #         nranks = 2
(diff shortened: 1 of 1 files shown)


mofeing:
mmm I wanted to make a comment on this yesterday before merging but got too busy.

> - Use TracedRNumber instead of TracedRequest to represent MPI_Request objects. TracedRequest or similar will only be necessary if we want to allow requests to cross the compile boundary, which for now we don't.

what was the reasoning behind removing `TracedRequest`?

romanlee:
I removed it because it added no functionality over tracedrnumber, ie was effectively just a duplicate. If we want to allow requests to cross the compile boundary, or need additional functionality for any other reason, we can put it back in. But until then it seems better to remove it

PR #2453, opened 2026-02-14
Add multirotate_spmd to OptimizeCommunicationOptions

jumerckx:


wsmoses:
```
error: 'enzymexla.rotate' op result shardings don't match number of values: 4 shardings vs 1 values        
```

running on the gb code.

implying it ran! but failed in the pass

PR #2458, opened 2026-02-15
Add profiler_backends dep for ROCm build

jariji:
For https://github.com/EnzymeAD/Reactant.jl/issues/2455


wsmoses:
does this fix it if you build locally?

jariji:
Turns out it doesn't, afaict.

PR #2459, opened 2026-02-15
Fix accidental type piracy

ptiede:


avik-pal:
Just merged the fix on master

PR #2465, opened 2026-02-16
Remove comment about needing to add @skip_rewrite* in the module for precompilation

jumerckx:
As linked by Sergio on slack, adding the macro in `__init__` should suffice to be picked up by precompilation: https://discourse.julialang.org/t/precompilation-init-and-eval/70188/2

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -215,9 +215,7 @@ will need a `@reactant_overlay` method.
     signature.
 
 !!! warning
-    The macro call should be inside the `__init__` function. If you want to
-    mark it for precompilation, you must add the macro call in the global scope 
-    too.
+    The macro call should be inside the `__init__` function.
 
 See also: [`@skip_rewrite_type`](@ref)
 """
@@ -264,9 +262,7 @@ If you want to mark the set of constructors over it's type parameters or over it
 abstract type, you should use then the `Type{<:MyStruct}` syntax.
 
 !!! warning
-    The macro call should be inside the `__init__` function. If you want to
-    mark it for precompilation, you must add the macro call in the global scope 
-    too.
+    The macro call should be inside the `__init__` function.
 """
 macro skip_rewrite_type(typ)
     typ = if Base.isexpr(typ, :curly) && typ.args[1] === :Type


mofeing:
I have to eat my own words. I did a quick test an apparently we **do need to add it in the top module too for precompilation**.

```julia
module InitTest

function __init__()
    Core.println("YELLOOOO")
end

Core.println("precomp")

end
```

The print in `__init__` only runs after precompilation 😢 

```julia
(InitTest) pkg> precompile
     Project No packages added to or removed from `~/Developer/InitTest.jl/Project.toml`
    Manifest No packages added to or removed from `~/Developer/InitTest.jl/Manifest.toml`
Precompiling packages finished.
  1 dependency successfully precompiled in 1 seconds
  1 dependency had output during precompilation:
┌ InitTest
│  precomp
└

julia> using InitTest
YELLOOOO
```

Maybe we can use this PR to be more concise on the docs. `@skip_rewrite_func` and `@skip_rewrite_type` do sth similar to the following code:

```julia
@reactant_overlay function THE_FUNC(args...)
    call_with_native(THE_FUNC, args...)
end
```

but `@skip_rewrite_*` should be faster as it avoids IR rewriting and thus, avoids compilation of the code above under Reactant custom pipeline. In the end, is a mechanism to try make Reactant's Julia compilation slightly faster.

As such, you need it in the top-module **to make precompilation faster**, and you need it in the `__init__` **to make compilation on runtime faster**.

jumerckx:
> In the end, is a mechanism to try make Reactant's Julia compilation slightly faster.

In the structarrays pr I used for more than that, compilation fails when Reactant rewrites a particular function, so the macro is required for correctness...

mofeing:
> > In the end, is a mechanism to try make Reactant's Julia compilation slightly faster.
> 
> In the structarrays pr I used for more than that, compilation fails when Reactant rewrites a particular function, so the macro is required for correctness...

right. should we close this or make some clarifications in the docstring?

jumerckx:
I think, in the end, the current docstrings are actually quite good 🙃
Feel free to close or let me know if you want me to add other stuff.

PR #2466, opened 2026-02-16
Fix `Base.reduce_empty` call to `BottomRF`

maximilian-gelbrecht:
To be honest, I am not even sure what exactly this does, but this just seems like a straight bug, `BottomRF` isn't in scope. I just got a  ```ERROR: UndefVarError: `BottomRF` not defined in `Reactant```  while working on our model


wsmoses (approved):

mofeing:
my fault. these has to do with #2227

Issue #2467, opened 2026-02-16
Broadcasting Error


dominic-chang:
I'm not sure how to get a minimal working example for this, but I get an error when broadcasting over the following code

```julia
function Base.floatmin(::Type{Reactant.TracedRNumber{T}}) where {T}
	return floatmin(T)
end

function Base.floatmax(::Type{Reactant.TracedRNumber{T}}) where {T}
	return floatmax(T)
end

function DRF_whilebody_ifbody(XN, YN, ZN)

    XNROOT = sqrt(XN)
    YNROOT = sqrt(YN)
    ZNROOT = sqrt(ZN)
    YNROOTZNROOT = YNROOT * ZNROOT
    LAMDA = muladd(XNROOT, (YNROOT + ZNROOT), YNROOTZNROOT)
    return (XN + LAMDA) / 4, (YN + LAMDA) / 4, (ZN + LAMDA) / 4
end

function DRF_whilebody(XN, YN, ZN, ERRTOL) 

    MU = (XN + YN + ZN) / 3
    ninvMU = -1 / MU
    XNDEV = muladd(ninvMU, (MU + XN), 2)
    YNDEV = muladd(ninvMU, (MU + YN), 2)
    ZNDEV = muladd(ninvMU, (MU + ZN), 2)
    EPSLON = max(abs(XNDEV), abs(YNDEV), abs(ZNDEV))
    XN, YN, ZN = Base.ifelse(EPSLON >= ERRTOL, DRF_whilebody_ifbody(XN, YN, ZN), (XN, YN, ZN))

    return (XN, YN, ZN, XNDEV, YNDEV, ZNDEV, MU, (EPSLON >= ERRTOL))
end

function DRF_ifbody(X::A, Y::B, Z::C, ERRTOL::D) where {A, B, C, D}
        T = promote_type(A, B, C, D)
        C1 = T(1 / 24)
        C2 = T(3 / 44)
        C3 = T(1 / 14)

        XN = X
        YN = Y
        ZN = Z
        MU = zero(T)
        XNDEV = zero(T)
        YNDEV = zero(T)
        ZNDEV = zero(T)
        CONTINUE = true 

        @trace while CONTINUE 
            XN, YN, ZN, XNDEV, YNDEV, ZNDEV, MU, CONTINUE = DRF_whilebody(XN, YN, ZN, ERRTOL)
        end
        XNDEVYNDEV = XNDEV * YNDEV
[...]

PR #2470, opened 2026-02-17
fix: broadcast error when type inference did not produce concrete eltype

avik-pal:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/2467

This probably warrants a deeper investigation, but atleast unblocks @dominic-chang's issue. The inferred type in base probably doesn't use the overloads in our interpreter.


wsmoses:
so stemming from an earlier version of his original case, https://github.com/EnzymeAD/Reactant.jl/pull/2394 does actually force it to use our interpreter for the various eltype methods

Issue #2503, opened 2026-02-22
Enzyme.Const arguments to Enzyme.autodiff frozen at trace-time values inside Reactant.compile


ChrisRackauckas-Claude:
## Summary

When `Reactant.compile` traces a kernel containing `Enzyme.autodiff`, `Enzyme.Const` arguments are frozen at their tracing-time values. At runtime, even when different values are passed to the compiled kernel, `Enzyme.autodiff` still uses the original tracing values.

This affects any function where a `Const` argument is used in a nonlinear operation (e.g. `exp(t)`, `sin(t)`). Linear usage or unused `Const` arguments are not affected since the frozen value doesn't change the derivative structure.

Found during SciMLSensitivity.jl ReactantVJP development (SciML/SciMLSensitivity.jl#1361).

## MWE

```julia
using Reactant, Enzyme

# f!(du, u, t): du[1] = exp(t) * u[1]
# Derivative w.r.t. u[1] is exp(t)
function f!(du, u, t)
    du[1] = exp(t) * u[1]
    return nothing
end

function vjp_kernel(dy_buf, u, t, λ)
    du_shadow = zero(u)
    dλ_seed = copy(λ)
    Enzyme.autodiff(Enzyme.Reverse, Enzyme.Const(f!), Enzyme.Const,
        Enzyme.Duplicated(dy_buf, dλ_seed),
        Enzyme.Duplicated(u, du_shadow),
        Enzyme.Const(t))          # <-- t is Const
    return dy_buf, du_shadow
end

# Compile with tracing value t = 0.0
dy = Reactant.ConcreteRArray(zeros(1))
u = Reactant.ConcreteRArray(zeros(1))
t_trace = Reactant.to_rarray(0.0)   # tracing value
λ = Reactant.ConcreteRArray(ones(1))

compiled = Reactant.@allowscalar Reactant.compile(vjp_kernel, (dy, u, t_trace, λ))

# Run at different t values
for t_val in [0.0, 1.0, 5.0]
    # Plain Enzyme (correct)
    du_ref = zeros(1)
    Enzyme.autodiff(Enzyme.Reverse, Enzyme.Const(f!), Enzyme.Const,
        Enzyme.Duplicated(zeros(1), [1.0]),
        Enzyme.Duplicated([1.0], du_ref),
        Enzyme.Const(t_val))

    # Reactant-compiled
    result = compiled(
        Reactant.ConcreteRArray(zeros(1)),
[...]

PR #2505, opened 2026-02-22
docs: document track_numbers and scalar tracing behavior

ChrisRackauckas-Claude:
## Summary

- Documents the `track_numbers` keyword argument for `Reactant.to_rarray`, which controls whether plain Julia scalars are tracked (converted to `ConcreteRNumber`) or left as compile-time constants
- Adds warnings about the common pitfall where scalar arguments (especially inside `Enzyme.Const`) are frozen at trace-time values because they weren't tracked

## Changes

- **`src/Tracing.jl`**: Add docstring to `to_rarray` explaining all keyword arguments with examples
- **`docs/src/api/api.md`**: Add `Reactant.to_rarray` to the API reference page under a new "Converting Data" section
- **`docs/src/tutorials/partial-evaluation.md`**: Add "Tracking Scalar Numbers with `track_numbers`" section explaining the kwarg, type filtering, nested struct support, and the common pitfall warning
- **`docs/src/tutorials/automatic-differentiation.md`**: Add "Scalar Arguments and `Enzyme.Const`" section warning about frozen scalars in AD contexts, with correct/incorrect usage examples
- **`docs/src/introduction/FAQs.md`**: Add FAQ entry "Scalar arguments are frozen at compile time" as first entry for discoverability
- **`docs/src/introduction/index.md`**: Add a note about `track_numbers` in the Getting Started page near `to_rarray` usage

## Motivation

Addresses #2503 — users were confused when `Enzyme.Const` arguments appeared frozen at trace-time values. The root cause is that `to_rarray` doesn't track scalars by default, but this wasn't documented anywhere.

## Test plan

- [ ] Verify docs build successfully (no broken cross-references)
- [ ] Check that `@docs Reactant.to_rarray` renders the new docstring correctly

### DIFF
--- a/src/Tracing.jl
+++ b/src/Tracing.jl
@@ -2113,7 +2113,7 @@ to `ConcreteRNumber`.
 - `sharding`: Sharding specification for the resulting array.
 - `device`: Target device for the resulting array.
 - `client`: XLA client to use.
-- `runtime`: Backend runtime to use (`:PJRT` or `:IFRT`).
+- `runtime`: Backend runtime to use (`Val(:PJRT)` or `Val(:IFRT)`).
 
 ## Examples
 


src/Tracing.jl
```diff
@@ -2094,6 +2094,47 @@ Base.@nospecializeinfer function make_tracer(
...
+  converted to `ConcreteRNumber`.
+  - `false` (default): scalars are left as-is and will be treated as compile-time constants
+    (frozen at tracing time).
+  - `true`: all scalar numbers are converted to `ConcreteRNumber`.
+  - A type (e.g. `Number`, `Float64`, `Int`): only scalars that are subtypes of the given
+    type are tracked.
+- `sharding`: Sharding specification for the resulting array.
+- `device`: Target device for the resulting array.
+- `client`: XLA client to use.
+- `runtime`: Backend runtime to use (`:PJRT` or `:IFRT`).
```
avik-pal:
```suggestion
- `runtime`: Backend runtime to use (`Val(:PJRT)` or `Val(:IFRT)`).
```

PR #2509, opened 2026-02-22
feat: header file for API.cpp and auto-generate the C bindings

avik-pal:


giordano:
Is this to avoid issues like #2508?  While this is fine in principle, my only word of caution is that the source of truth of the API/ABI of a library is its implementation, not the header file, which needs to be kept consistent with the implementation anyway, so you're simply shifting the place of possible error from the Julia hand-written `@ccall` to the hand-written header file, not concretely saving any source of human error. Automatically generating the Julia bindings is nice regardless of the possibility of avoiding human mistakes though.

wsmoses (commented):
can we instead get the definitions out of api.cpp? perhaps limiting to cases where the fn is marked as REACTANT_ABI ?

deps/ReactantExtra/API.cpp
```diff
@@ -2975,9 +3143,11 @@ REACTANT_ABI void ifrt_hlo_module_cost_analysis_properties(
 
 #pragma endregion
 
-REACTANT_ABI void dump_op(Operation *op) { llvm::errs() << *op << "\n"; }
-REACTANT_ABI void dump_mval(mlir::Value v) { llvm::errs() << v << "\n"; }
-REACTANT_ABI void dump_operation(Operation *op, const char *filename) {
+REACTANT_ABI void dump_op(void *op) {
+  llvm::errs() << *static_cast<Operation *>(op) << "\n";
+}
+REACTANT_ABI void dump_mval(MlirValue v) { llvm::errs() << unwrap(v) << "\n"; }
```
wsmoses, line 3161: these three explicitly use the non c api, and are here for debugging, not for julia c binding [but still need to be exported with reactant_abi]

wsmoses (approved):
minus the minor comment, lgtm

deps/ReactantExtra/make.jl
```diff
@@ -28,64 +28,174 @@ function rewrite!(::ExprDAG) end
...
+            content = read(file, String)
+            for m in eachmatch(r"REACTANT_ABI\s+(.*?)(?:\{|=(\s*nullptr\s*)?;|;)"s, content)
+                sig = m.match
+                sig = replace(sig, r"\s*\{\s*$" => ";")
+                sig = replace(sig, r"\s*=\s*nullptr\s*;\s*$" => ";")
+                sig = replace(sig, "REACTANT_ABI" => "MLIR_CAPI_EXPORTED")
+                # Remove namespaces for C mode compatibility
+                # E.g. tsl::ProfilerSession * -> ProfilerSession *
+                sig = replace(sig, r"\b(?:\w+::)+" => "")
+                # Also handle some C++ specific things that might remain
```
wsmoses: what exactly his this replacment? the api for std string/etc is not equivaelnt to const char*?

avik-pal: oops accidental. removed

PR #2515, opened 2026-02-23
chore: bump jll

avik-pal:


avik-pal:
definitely busted something upstream

avik-pal:
Seems like TPU specific failure. @wsmoses did we introduce any change upstream that would be TPU only?

wsmoses:
temporarily suspending other github ci, as we need one runner to register the jll, will restart after.

and @avik-pal nope, none to my knowledge, beyond xla/jax bumping. this doesn't include the tpu bump either [tho we should also do/investigate that]

wsmoses:
actually, let's do the jll bump in the wrap pr

PR #2529, opened 2026-02-23
Add a tiny Oceananigans model run as a CI test

jlk9:
This adds a very lightweight Oceananigans model run to the Reactant testing infrastructure for CI purposes with new merges.

@wsmoses


mofeing:
> imo this is way too big to make it into a unit test, unit tests would need to be more akin to kernel sized

but maybe it can be considered a system test and we run it only once instead of one-per-configuration? in such case, we need to extend the infra a lil bit.

Issue #2482, opened 2026-02-19
`XLA_FLAGS=--xla_dump_to` doesn't seem to be working anymore


giordano:
I don't have the time to debug this now, but since around https://github.com/EnzymeAD/Enzyme-JAX/commit/304042eec164f445e44cce230d5037c4fc6cee42 `XLA_FLAGS=--xla_dump_to` doesn't seem to trigger the XLA dump anymore for some reason, at least inside Reactant.jl, but [that still exists upstream](https://github.com/openxla/xla/blob/55b2af36b0422d1b0b0096fa1a610fbf89bf8cb8/xla/debug_options_flags.cc#L1369) as far as I can tell, so I'm quite missing what's the issue.  Are we messing with the XLA flags somehow?

wsmoses:
oh I think you might also need the pass regex

giordano:
No, you need to pass a directory, but that's what we do in https://github.com/EnzymeAD/Enzyme-JAX/blob/c49897e00386afcc9b4e0014fba378b7f55472d2/.github/workflows/test-gb-25.yml#L189

wsmoses:
yes I mean we need both flags

giordano:
`--xla_dump_to` works on its own by default, `--xla_dump_hlo_pass_re` (the regex one) is an extra for getting more stuff.  Or at least that used to be the case.

giordano:
https://github.com/openxla/xla/blob/3141cfc39a396faf6eb0f54d499bcd772f3b5c2f/docs/hlo_dumps.md#using-environment-variables confirms that `--xla_dump_to` should work on its own, then you can tweak what and how to dump stuff.

avik-pal:
might be worth checking what the value for that flag is in https://github.com/EnzymeAD/Reactant.jl/blob/772f800f5677fad227e23a41f0968756aeaae1b5/src/xla/CompileOptions.jl#L65

avik-pal:
It should be picking up flags https://github.com/EnzymeAD/Reactant.jl/blob/772f800f5677fad227e23a41f0968756aeaae1b5/deps/ReactantExtra/API.cpp#L3659

giordano:
```
$ XLA_FLAGS=--xla_dump_to=. julia --project -q
julia> using Reactant

julia> Reactant.XLA.get_debug_options().xla_dump_to
""
```
This looks wrong, no?

giordano:
> This looks wrong, no?

[...]

PR #2530, opened 2026-02-23
fix: accidental bake-in of options

avik-pal:


giordano:
Is this related to #2482?  I thought was fixed by something else (haven't checked)

avik-pal:
Yes, the linked PR accidentally made it seem like a fix.

PR #2541, opened 2026-02-24
TPU: forward properly xla flags

wsmoses:


src/xla/PJRT/Client.jl
```diff
@@ -192,6 +192,15 @@ function MakeTPUClient(;
...
                                                     distributed_runtime_client"
 
+    # LibTPU has its own internal copy of XLA which does not read the regular XLA flags
+    if !haskey(ENV,  "LIBTPU_INIT_ARGS")
+        xla_flags = "--xla_enable_enzyme_comms_opt=true"
+        if haskey(ENV,  "XLA_FLAGS")
+            xla_flags = xla_flags * " " * ENV["XLA_FLAGS"]
+        end
+        ENV["LIBTPU_INIT_ARGS"] = xla_flags
+    end
```
avik-pal, line 202: not sure if this will do it. we might need `setenv` or `withenv`

gbaraldi: That depends on what you want to happen. the ENV dict is special so setting it is equivalent to setenv

wsmoses: we need the ccall that follows this to have LIBTPU_INIT_ARGS set appropriately. Does this do that?

wsmoses: talkign with @gbaraldi on slack, ENV does immediatelly call setenv, so this should work as intended

PR #2543, opened 2026-02-24
Add script to regenerate mlir bindings manually

romanlee:
I use this script to manually do the actions that `.github/workflows/regenerate-mlir-bindings.yml` does. I find this useful when actively developing Enzyme-JAX and Reactant simultaneously, and thought it might be worth considering checking it in.


mofeing:
If you have docker and act installed, you can use them to call the `regenerate-mlir-bindings.yml` workflow locally.

romanlee:
Brilliant, thank you. CLosing the pr

PR #2546, opened 2026-02-25
Better StructArray & StaticArray support

ptiede:


src/TracedUtils.jl
```diff
@@ -1146,7 +1146,12 @@ function elem_apply_via_while_loop(f, args::Vararg{Any,Nargs}) where {Nargs}
     else
         typeof(res_tmp)
     end
-    result = similar(first(flat_args), T_res, L)
```
ptiede, line 1149: This should be reviewed closely. I had to change this line because now `flat_args` may have differing array types, e.g., one could be a `StructArray`. Before we decided the output entirely based on the first element, which could lead to errors, e.g., `2 .* sa` would try to write a `StructArray` to a `TracedRArray`.

ptiede:
@wsmoses @jumerckx I think this is ready for a first review. I'll add some tests tomorrow, but this seems to enable more support for StructArrays. Right now, the only thing that could cause an issue is if one of the underlying arrays isn't a ReactantPrimitive, but that should be enough for what I need.

src/TracedUtils.jl
jumerckx: Oh this is clever!

ext/ReactantStaticArraysExt.jl
```diff
@@ -0,0 +1,29 @@
...
+end
+
+function Reactant.materialize_traced_array(x::SArray)
+    as = Reactant.aos_to_soa(collect(x))
+    return Reactant.materialize_traced_array(as)
+end
+
+# We don't want to overload map on StaticArrays because it autopromote to TracedRArrays which we 
+# do not want.
+overloaded_map(f, a::StaticArray{<:Tuple,<:Reactant.TracedRNumber}) = f.(a)
```
avik-pal: If we do this, won't it unroll the loop? Or are we forcing an  unroll because StaticArrays are expected to be used only for small inputs?

ptiede: Ya I figured we would just force an unroll because they are small. The other option would be to convert it back to a StaticArray after the computation? I'm fine with that if people are concerned with compile times. The largest arrays I would ever do are 4x4 matrices or, at the absolute maximum, a 4x4x4x4 tensor.

avik-pal: I am okay with an unroll as well; static arrays aren't expected to be large

ext/ReactantStructArraysExt.jl
```diff
@@ -90,41 +131,61 @@ Base.@nospecializeinfer function Reactant.traced_type_inner(
...
-        mode;
-        track_numbers,
-        sharding,
-        runtime,
-        kwargs...,
-    )
-    T_traced = traced_type(typeof(prev), Val(mode), track_numbers, sharding, runtime)
-    return StructArray{first(T_traced.parameters)}(traced_components)
-end
+# function Reactant.make_tracer(
```
wsmoses: if this can be removed fully, remove the code?

ptiede: Done

ext/ReactantStaticArraysExt.jl
```diff
@@ -0,0 +1,29 @@
...
+    mode::Reactant.TraceMode,
+    @nospecialize(track_numbers::Type),
+    @nospecialize(ndevices),
+    @nospecialize(runtime)
+) where {S,T,N,L}
+    T_traced = Reactant.traced_type_inner(T, seen, mode, track_numbers, ndevices, runtime)
+    return SArray{S,T_traced,N,L}
+end
+
+function Reactant.materialize_traced_array(x::SArray)
```
ptiede: @wsmoses @avik-pal @jumerckx do we actually want to materialize SArrays? In one of my usecases I noticed a 3x slowdown when doing so. It looks like when we materialize a cutlass matrix multiply is constantly called with a block size of 1,1,1, which seems to add a ton of overhead.

Since `SArrays` are usually small, I think effectively unrolling any operations on them makes sense.

ptiede:
@avik-pal @Pangoraw @wsmoses I think this is ready for review. Once this is merged, most of Comrade should work.

ptiede:
Actually one question is what are the semantics of 

```julia
overloaded_mul(A, B, alpha, beta)
```
typically for the 5-argument mul in place is 
```
C = alpha A*B + beta * C
```
I'm not sure what this should be when C isn't defined.

ext/ReactantStructArraysExt.jl
```diff
@@ -53,10 +53,40 @@ function Base.copyto!(
...
+    Reactant.TracedUtils.elem_apply_via_while_loop(bc.f, args...; dest)
+    return dest
+end
+
+function Base.similar(
+    bc::Broadcasted{StructArrayStyle{S,N}}, ::Type{ElType}
+) where {S<:AbstractReactantArrayStyle,N,ElType}
+    bc′ = convert(Broadcasted{S}, bc)
+    # It is possible that we have multiple broadcasted arguments
+    alloc(::Type{T}) where {T} = (T <: Complex) ? similar(bc′, T) :
```
wsmoses: can this be defined outside the similar defn [aka not a closure] and take bc as an argument

src/TracedUtils.jl
```diff
@@ -1115,13 +1115,23 @@ function __elem_apply_loop_condition(idx_ref, fn_ref::F, res_ref, args_ref, L_re
...
+Base.broadcastable(x::RefFillVector) = x
+
 function __elem_apply_loop_body(idx_ref, fn_ref::F, res_ref, args_ref, L_ref) where {F}
     args = args_ref[]
     fn = fn_ref[]
     res = res_ref[]
     idx = idx_ref[] + 1
 
-    scalar_args = [@allowscalar(arg[idx]) for arg in args]
+    scalar_args = ntuple(length(args)) do i
```
wsmoses: don't bother with ntuple?

src/TracedUtils.jl
```diff
@@ -1130,13 +1140,25 @@ function __elem_apply_loop_body(idx_ref, fn_ref::F, res_ref, args_ref, L_ref) wh
 end
 
 function elem_apply_via_while_loop(f, args::Vararg{Any,Nargs}) where {Nargs}
-    @assert allequal(size.(args)) "All args must have the same size"
-    L = length(first(args))
+    scalar_arg(arg) = arg isa Base.RefValue || !(arg isa AbstractArray)
```
wsmoses: define this outside the loop

src/TracedUtils.jl
```diff
@@ -1130,13 +1140,25 @@ function __elem_apply_loop_body(idx_ref, fn_ref::F, res_ref, args_ref, L_ref) wh
...
+    non_ref_args = Tuple(arg for arg in args if !scalar_arg(arg))
```
wsmoses: array not tuple

src/TracedUtils.jl
```diff
@@ -1130,13 +1140,25 @@ function __elem_apply_loop_body(idx_ref, fn_ref::F, res_ref, args_ref, L_ref) wh
...
+    if !isempty(non_ref_args)
+        @assert allequal(size.(non_ref_args)) "All args must have the same size"
+    end
+    out_size = isempty(non_ref_args) ? () : size(first(non_ref_args))
+    L = isempty(non_ref_args) ? 1 : length(first(non_ref_args))
     # flattening the tensors makes the auto-batching pass work nicer
-    flat_args = [ReactantCore.materialize_traced_array(vec(arg)) for arg in args]
+    flat_args = ntuple(Val(Nargs)) do i
```
wsmoses: array not tuple

ext/ReactantStructArraysExt.jl
```diff
@@ -71,10 +107,21 @@ Base.@propagate_inbounds function Base.setindex!(
     s::StructArray{T,<:Any,<:Any,Int}, vals, I::TracedRNumber{TI}
 ) where {T,TI<:Integer}
     valsT = maybe_convert_elt(T, vals)
-    foreachfield((col, val) -> (@inbounds col[I] = val), s, valsT)
+    foreachfield((col, val) -> (@inbounds Reactant.@allowscalar col[I] = val), s, valsT)
```
wsmoses: lambda functions are going to cause some compile time issues, can we get rid of them with functions defined directly (possibly with fix1/etc as relevant)?

PR #2557, opened 2026-02-27
Add cosh and sinh

ptiede:
It looks like CHLO.jl already had specified cosh and sinh but the opcall was missing

### DIFF
--- a/src/TracedRNumber.jl
+++ b/src/TracedRNumber.jl
@@ -671,7 +671,9 @@ for (jlop, hloop) in (
     (:(Base.abs), :abs),
     (:(Base.:-), :negate),
     (:(Base.sin), :sine),
+    (:(Base.sinh), :sinh),
     (:(Base.cos), :cosine),
+    (:(Base.cosh), :cosh),
     (:(Base.tan), :tan),
     (:(Base.tanh), :tanh),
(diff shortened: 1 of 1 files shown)


src/TracedRNumber.jl
```diff
@@ -671,7 +671,9 @@ for (jlop, hloop) in (
     (:(Base.abs), :abs),
     (:(Base.:-), :negate),
     (:(Base.sin), :sine),
+    (:(Base.sinh), :sinh),
     (:(Base.cos), :cosine),
+    (:(Base.cosh), :cosh),
```
giordano, line 676: Where did you find these functions?  I don't see them in the [StableHLO specification](https://openxla.org/stablehlo/spec)

ptiede: They are in the chlo specification https://openxla.org/stablehlo/generated/chlo#chlosinh_chlosinhop

ptiede:
@wsmoses, for tests, the Ops.cosh is actually already tested, but the Base.cosh override isn't. Interestingly, this seems to be what is commonly done in `test/core/ops.jl`? Should I also add tests for the override of the other functions?

wsmoses:
oh right, that's fine as is then

PR #2572, opened 2026-02-28
Add gfx1102 target

jariji:
Radeon RX 7600 is gfx1102

https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html


jariji:
I have this gpu and I want to try using Reactant.jl on it. Is this PR the right idea?

PR #2574, opened 2026-03-01
feat: callback into julia from compiled mlir

avik-pal:


deps/ReactantExtra/xla_ffi.cpp
```diff
@@ -47,6 +47,104 @@ XLA_FFI_DEFINE_HANDLER(xlaThrowErrorHandlerHost, xlaThrowErrorHost,
...
+//                 void** outputs, int32_t* output_types,
+//                 int32_t* output_ranks, int64_t** output_dims,
+//                 int64_t  num_outputs);
+//
+// Element type encoding mirrors xla::ffi::DataType (== xla::PrimitiveType):
+//   PRED=1, S8=2, S16=3, S32=4, S64=5, U8=6, U16=7, U32=8, U64=9,
+//   F16=10, F32=11, F64=12, BF16=16, C64=15, C128=18, ...
+// ============================================================================
+
+using JuliaCallbackFn = void (*)(
```
wsmoses: should the julia function return a bool designating an error?

avik-pal: updated to return a bool

avik-pal:
Locally tested both host callback and cuda callback (via CUDA.jl) work successfully

src/Ops.jl
```diff
@@ -4234,4 +4234,204 @@ end
...
+        location = Ops.mlir_stacktrace("custom_call", @__FILE__, @__LINE__),
+    )
+
+Emit a `stablehlo.custom_call` that will call back into Julia function `f` at
+execution time.
+
+!!! warning "Limited Backend Support"
+
+    Currently this only supports CUDA and CPU backends. For CUDA support, `CUDA.jl`
+    must be loaded.
```
wsmoses, line 4343: I assume also only single device?

PR #2579, opened 2026-03-01
Fix getindex_linear when indices eltype is Traced

ptiede:
This comes up when I have a range or vector whose elements are `Traced` I wasn't sure if I should case `eltype(indices)` or just intercept all `AnyTracedRArray`.


src/Indexing.jl
```diff
@@ -514,7 +514,7 @@ function get_slice_stride(x)
 end
 
 function getindex_linear(a::TracedRArray{T,N}, indices::AbstractArray) where {T,N}
-    if !(indices isa Reactant.TracedType)
```
avik-pal, line 517: Add a test for future

ptiede:
Added a test that covers the usecase for TransformVariables.

PR #2581, opened 2026-03-01
Add run_mlir.jl: pre_xla MLIR → standalone XLA execute script

gbaraldi:
## Summary

- Adds `scripts/run_mlir.jl`, a two-phase code-generation tool for testing the XLA compilation/execution pipeline on `*_pre_xla_compile.mlir` dumps — without needing the full model (e.g. Oceananigans/GB-25) setup
- The generator parses MLIR signatures and emits a self-contained Julia script with hardcoded shapes, shardings, and compile options
- Generated scripts support `--cpu` flag to run on virtual CPU devices via `XLA_FLAGS="--xla_force_host_platform_device_count=N"`

## How it works

**Phase 1 — Generator** (no Reactant dependency):
```bash
julia run_mlir.jl first_time_step.mlir loop.mlir output.jl
```
- Regex-parses `func.func @main(...)` signatures (tensor types, shapes, `sdy.sharding` annotations)
- Extracts mesh spec, `mhlo.num_partitions`/`num_replicas` from module attributes
- Detects grid constants (args before first `tf.aliasing_output`)
- Emits a standalone script with all shapes/shardings baked in

**Phase 2 — Generated script**:
```bash
julia --project=Reactant.jl output.jl        # on GPUs
julia --project=Reactant.jl output.jl --cpu   # on N virtual CPU devices
```
- Creates mock `ConcreteRArray`/`ConcreteRNumber` with correct `NamedSharding`
- Loads MLIR, XLA-compiles with shardy partitioner
- Executes first_time_step → marshals outputs → executes loop
- Syncs all results (`synced_buffer`) for accurate timing
- Handles both PJRT and IFRT runtimes

## Test plan

- [x] Generator parses GB-25 MLIR (38/39 inputs, 26 outputs, mesh `["x"=2,"y"=2]`)
- [x] Example MLIR (single-device `stablehlo.add`) end-to-end on GPU
- [x] Example MLIR end-to-end with `--cpu`
- [x] Full GB-25 MLIR compile + execute with `--cpu` (4 virtual devices)
- [ ] Full GB-25 MLIR on 4 real GPUs

### DIFF
--- a/src/serialization/MLIRRunner.jl
+++ b/src/serialization/MLIRRunner.jl
@@ -18,9 +18,20 @@ using ..Reactant.MLIR: IR
 # ──────────────────────────────────────────────────────────────
 
 struct TensorSig
-    eltype::String
+    eltype::Type
     mlir_shape::Vector{Int}
     mlir_sharding::Vector{Symbol}  # per Julia dim: :_none = replicated, :x/:y = axis
+end
+
+"""Result of [`analyze_module`](@ref)."""
+struct ModuleInfo
+    inputs::Vector{TensorSig}
+    outputs::Vector{TensorSig}
+    alias_map::Dict{Int,Int}       # output_idx (1-based) → input_idx (1-based)
+    num_partitions::Int
+    num_replicas::Int
+    mesh_axes::Vector{Symbol}
+    mesh_sizes::Vector{Int}
 end
 
 """Get a named element from a dictionary attribute, returning `nothing` if absent."""
@@ -100,7 +111,7 @@ function analyze_module(mlir_string::String)
 
     for i in 1:n_in
         mlir_type = IR.input(ftype, i)
-        T = string(IR.julia_type(IR.eltype(mlir_type)))
+        T = IR.julia_type(IR.eltype(mlir_type))
         shape = IR.istensor(mlir_type) && IR.ndims(mlir_type) > 0 ?
             collect(Int, IR.size(mlir_type)) : Int[]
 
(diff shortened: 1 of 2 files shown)


gbaraldi:
@avik-pal Maybe this is useful for reproducing gb-25 bugs. Also might be useful for executing as well since it can obliviate a lot of the loop (well the logic)

scripts/run_mlir.jl
```diff
@@ -0,0 +1,594 @@
+#!/usr/bin/env julia
+"""
+    run_mlir.jl — Generator: parse pre_xla MLIR files → emit standalone execute script.
```
avik-pal: why do we need this to not have reactant dependency? Seems like parsing would be much simpler with that one dep

scripts/run_mlir.jl
```diff
@@ -0,0 +1,561 @@
+#!/usr/bin/env julia
+"""
+    run_mlir_v2.jl — Generator (Reactant-dependent): MLIR IR introspection → emit execute script.
+
+Usage:
+    julia --project=Reactant.jl run_mlir_v2.jl [first.mlir] [loop.mlir] [output.jl]
```
wsmoses: Can you clean up these comments

scripts/run_mlir.jl
```diff
@@ -0,0 +1,561 @@
...
+""")
+
+    return String(take!(io))
+end
+
+# ──────────────────────────────────────────────────────────────
+# Main: analyze MLIR via IR APIs, emit script
+# ──────────────────────────────────────────────────────────────
+
+function main()
```
wsmoses: can this just be added as a reactant-internal utility function in the serialization subpackage that takes a list of mlir files [in this case it would be first and loop], and an output path?

avik-pal: even nicer would be a cli app on 1.12+ that calls the serialization function

gbaraldi: The question is how do you choose how to pipe outputs into inputs etc.

wsmoses: at least for oceananigans both functions modify an existing model in place (and both take the same input type, an oceananigans model, which should be decomposed into the same arguments). So all outputs are just updated versions of certain input args. Reactant-side we have a map from a given output to the input path it needs to update, I don't remember if that information is available mlir-side? @avik-pal

if not we can embed it

PR #2589, opened 2026-03-02
feat: use the updated distributed API

avik-pal:
- [x] https://github.com/EnzymeAD/ReactantBuilder/pull/123
- [x] defaults are changed to mimic XLA defaults


avik-pal:
@wsmoses we can probably run on perlmutter with VLOG >= 3 to check the values being set on the cpp side

PR #2595, opened 2026-03-03
Update Reactant_jll version to 0.0.345

wsmoses:


wsmoses:
@avik-pal presumably the xprof bump caused some issues here, if you can take a look?

wsmoses:
okay @avik-pal I'm going to merge for now so I can do a nersc run with the better logging infra, can you investigate this post-commit?

PR #2558, opened 2026-02-27
Pretty printing in the REPL of @code_hlo, @code_mhlo etc.

Antipath1:
Makes the macros now emit pretty colors, making it a bit easier to read the generated IR. Does this by implementing a simple parser generating annotations. Uses StyledStrings for this which is in Stdlib for Julia 1.11+ and is a registered package for Julia 1.10 and below. I do not really now in which file I should put tests for this so a bit of help would be nice for that.

PR #2601, opened 2026-03-03
feat: use proper lexer from upstream MLIR

avik-pal:
follow up to https://github.com/EnzymeAD/Reactant.jl/pull/2558

image

image


Project.toml
```diff
@@ -147,9 +148,10 @@ SparseArrays = "1.10"
 SpecialFunctions = "2.4"
 StableRNGs = "1.0.4"
 Statistics = "1.10"
+StructArrays = "0.7.2"
 StructUtils = "2.6.2"
+StyledStrings = "1"
```
Antipath1:
```suggestion
StyledStrings = "1.0.3, 1.11"
```

src/mlir/Highlight.jl
```diff
@@ -20,147 +22,97 @@ const REACTANT_THEME = [
...
+        if !has_ctx
+            IR.dispose(ctx)
         end
     end
     return tokens
 end
 
 """
-    style(str::AbstractString, tokens::Vector{MLIRToken})
+    style_MLIR(str::AbstractString, tokens::Vector{MLIRToken})
```
mofeing:
```suggestion
    style(str::AbstractString, tokens::Vector{MLIRToken})
```

PR #2602, opened 2026-03-03
Handle broadcast when the type is Extruded

ptiede:
This came up in some of my code. I'll work to add a test shortly

### DIFF
--- a/test/core/bcast.jl
+++ b/test/core/bcast.jl
@@ -86,6 +86,29 @@ end
     @test @jit(custom_ln(x_ra)) ≈ custom_ln(x)
 end
 
+struct Unstable
+    x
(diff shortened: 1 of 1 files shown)


wsmoses (commented):
yeah seems reasonable, but yeah we should add a test

ptiede:
~I'm curious why this is even getting triggered. From Julia, it appears that the regular broadcast is inferred, but in Reactant land, type inference returns any.~

Nvm just a really nasty Julia quirk where a tiny change introduced a very unclear type instability...
Weirdly enough if I remove it I get a different error. I'll open an issue for that later.

wsmoses (approved):

PR #2604, opened 2026-03-03
Refactor activity handling in Enzyme.jl

wsmoses:


avik-pal:
Can confirm that this does fix the issue. Unfortunately don't have a smaller testcase

wsmoses:
okay if you're fine with the pr as is then im happy to merge if the rest of the tests pass

PR #2613, opened 2026-03-05
Extending the oceananigans benchmark

jlk9:
This expands the Oceananigans benchmark from 9 steps to 25, and profiles the primal as well as reverse-mode derivative.

### DIFF
--- a/benchmark/oceananigans/abernathey_channel.jl
+++ b/benchmark/oceananigans/abernathey_channel.jl
@@ -381,18 +381,18 @@ function run_abernathey_channel_benchmark!(
         model, Tᵢ, Sᵢ, u_wind_stress, v_wind_stress, T_flux
     )
 
-    restimate_tracer_error = @compile raise_first=true raise=true sync=true estimate_tracer_error(
-        model, Tᵢ, Sᵢ, u_wind_stress, v_wind_stress, T_flux, Δz
-    )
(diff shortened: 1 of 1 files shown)


wsmoses:
cpu times out so you may need to decrease steps there, for gpu/cpu results:


```
[
  {
    "name": "Oceananigans/SpinUpReentrantChannelModel/CUDA/Primal",
    "unit": "s",
    "value": 1.398014107
  },
  {
    "name": "Oceananigans/DifferentiateTracerError/CUDA/Reverse",
    "unit": "s",
    "value": 21.84650777
  }
]
[
  {
    "name": "Oceananigans/SpinUpReentrantChannelModel/TPU/Primal",
    "unit": "s",
    "value": 1.231319401
  },
  {
    "name": "Oceananigans/DifferentiateTracerError/TPU/Reverse",
    "unit": "s",
    "value": 1.624095445
  }
]
```

jlk9:
@wsmoses That makes sense. TBH I wasn't planning on adding a 400 step run to Reactant CI. This was proof for the paper that we have a working 400 step run on TPU.

Interesting that CUDA has a steep jump in primal vs reverse time, but TPU doesn't.

wsmoses:
yeah it is [we were seeing semi relatedly better memory optimizations on tpu at the moment as well for pure primals too, that may be relevant here]

PR #2626, opened 2026-03-08
feat: direct backsolve

avik-pal:


src/stdlibs/factorization/Factorization.jl
```diff
@@ -62,3 +62,31 @@ function _overloaded_backslash(F::BatchedAdjointFactorization, B::AbstractArray)
     ldiv!(F, BB)
     return BB[1:size(F)[2], ntuple(Returns(Colon()), ndims(B) - 1)...]
 end
+
+function Base.:(\)(
+    A::TracedRArray{<:Any,2}, B::Union{TracedRArray{<:Any,2},TracedRArray{<:Any,1}}
```
wsmoses, line 67: is there a reason for not doing AnyTracedArray here (e.g. view/etc)?

avik-pal: We want the fallbacks Lower/UpperTriangular, Diagonal, etc.

wsmoses: Got it, okay then LGTM

PR #2635, opened 2026-03-10
Add debug option to code_hlo

Pangoraw:


src/Compiler.jl
```diff
@@ -2937,22 +2951,31 @@ macro code_mhlo(args...)
...
             ),
         ),
         args...,
     )
+    debug = get(() -> Expr(:kw, :debug, false), options, something(findfirst(opt -> opt.args[1] === :debug, options), -1)).args[2]
+    options = filter(opt -> opt.args[1] !== :debug, options)
     return quote
         $MLIR.IR.@dispose ctx = $Reactant.ReactantContext() begin
+            debug = $(esc(debug))
+            old = $(Reactant.Ops.DEBUG_MODE)[]
```
wsmoses: this form of debug has been largely superceeeded by the debug info added via the new interpreter functions we generate and track backtraces on, do you need this specifically?

Pangoraw: I removed it, if one wants access to it, it can be wrapped with `Reactant.with_debug`

PR #2639, opened 2026-03-10
Bump ReactantCore compat to 0.1.17

Pangoraw:

PR #2646, opened 2026-03-11
Make Periodic not exported but public

Pangoraw:
There is an unfortunate collision with Oceananigans.Periodic.

### DIFF
--- a/src/Reactant.jl
+++ b/src/Reactant.jl
@@ -304,7 +304,7 @@ export ConcreteRArray,
     @trace,
     within_compile
 
-if VERSION ≥ v"1.11"
+@static if VERSION ≥ v"1.11"
     @eval $(Expr(:public, :Periodic))
 end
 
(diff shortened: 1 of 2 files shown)


giordano:
Also, when I precompile the package locally with v1.11.9 I get
```
┌ Reactant
│  WARNING: could not import ReactantCore.Periodic into Reactant
└  
```

Pangoraw:
> Also, when I precompile the package locally with v1.11.9 I get

Are you on ReactantCore@0.1.17 or ReactantCore@0.1.16? I think this warning can happen in the latter case.

giordano:
Yes: https://github.com/NumericalEarth/breeze-docker-images/actions/runs/22952120192/job/66619175350#step:6:1431.  Someone didn't bump a compat bound?  Incompatible combinations shouldn't be allowed.

Pangoraw:
yeah there are a few commits without the right compat bound: https://github.com/EnzymeAD/Reactant.jl/pull/2639

wsmoses:
@giordano @Pangoraw feel free to bump/release/etc

Pangoraw:
Ok let's release after fixing tests

PR #2658, opened 2026-03-13
Add device-to-device CuArray <-> ConcreteRArray conversion

jacobadenbaum:
## Summary

Adds zero-host-roundtrip conversion between CUDA.jl `CuArray`s and Reactant's `ConcretePJRTArray`s. Data stays on the GPU throughout -- no copies to host memory.

- `ConcretePJRTArray(cu::DenseCuArray)`: CuArray -> Reactant (forward direction)
- `CuArray(ra::ConcretePJRTArray)`: Reactant -> CuArray (reverse direction)

## Motivation

Interoperating between CUDA.jl and Reactant currently requires a host roundtrip (`Array(cu)` then `ConcreteRArray(host_array)`), which is slow for large arrays. This PR adds direct device-to-device transfer using PJRT's own stream infrastructure for correct synchronization.

This has been a particular pain point for me in work downstream, where I need to iterate back and forth between model training (which is accelerated substantially with Reactant) and certain simulation code I have written with CUDA, but which involves computational patterns that don't work with Reactant.  Without this change, the roundtrip through the host completely dominates the runtime.

I'll note: I am a computational macroeconomist, not really an expert on GPU programming.  So I've been getting a lot of help from Claude Code in actually figuring out what would need to be done here, and with many of the implementation details.  That being said, I've gone over this quite a lot, I've made sure there are comprehensive tests and that they all pass when I rebuild the project in my GPU's environment, and I am reasonably confident it is all correct. But it would certainly benefit from another human set of eyes who understands the underlying codebase better than I do.

### C++ layer (`deps/ReactantExtra/API.cpp`)

Two new functions:

**`BufferFromDevicePointer`** (forward direction):
1. Creates an ephemeral non-owning PJRT view of the external device pointer via `CreateViewOfDeviceBuffer`
2. D2D copies into a new PJRT-owned buffer via `CopyToMemorySpace`
3. Awaits the copy's completion via `GetReadyFuture().Await()` before returning, ensuring the caller's `GC.@preserve` on the source CuArray covers the entire read (since `CopyToMemorySpace` is async internally)
4. Returns a fully PJRT-owned buffer with correct definition events

**`AwaitBufferReady`** (reverse direction):
1. Calls `buffer->GetReadyFuture().Await()` to block until all PJRT definition events resolve
2. Returns the raw device pointer via `UnsafeBufferPointer`
3. After Await returns, GPU memory coherence guarantees the data is visible to any CUDA context

### Julia bindings (`src/mlir/libMLIR_h.jl`)

Standard `@ccall` wrappers for both C++ functions.

### High-level conversions (`ext/ReactantCUDAExt.jl`)

**`ConcretePJRTArray(cu::DenseCuArray)`**:
- Synchronizes the CUDA stream to ensure all writes are complete
- Calls `BufferFromDevicePointer` with `stream=0` (data already synchronized)
- Wraps result in `ConcretePJRTArray` with proper sharding metadata
- Non-contiguous `SubArray` views of CuArrays are also supported via a convenience method that materializes a contiguous copy first

**`CuArray(ra::ConcretePJRTArray{T,N,1})`** (single-device only):
- Calls `AwaitBufferReady` to get synchronized device pointer
- `cuMemcpyDtoD_v2` into a new CuArray (synchronous copy, no additional sync needed)

### macOS build support (`deps/ReactantExtra/BUILD`)

Added exported symbols for macOS compatibility.

## Synchronization model

All synchronization relies on documented PJRT API contracts, not implementation details.

**Forward (CuArray -> Reactant)**: `CUDA.synchronize()` ensures all CUDA.jl writes are visible before PJRT reads. `CopyToMemorySpace` schedules an async D2D copy on XLA's internal stream. `GetReadyFuture().Await()` blocks until the copy completes, ensuring the source CuArray memory is no longer being read before the function returns (and before `GC.@preserve` ends).

**Reverse (Reactant -> CuArray)**: `GetReadyFuture().Await()` ensures all PJRT operations are complete. After Await, GPU coherence guarantees the data is visible to CUDA.jl's context. `cuMemcpyDtoD_v2` (synchronous) copies into a new CuArray.

Both directions produce independent copies -- mutating the source after conversion does not affect the destination.

## Test plan

Comprehensive test suite in `test/integration/cuda_d2d.jl`:

- [x] Basic correctness: 1D, multi-dimensional, multiple element types (Float32, Float64, Float16, BFloat16, Int32, Int64)
- [x] Both directions: CuArray -> ConcreteRArray and ConcreteRArray -> CuArray
- [x] Round-trip fidelity in both directions
- [x] Stress tests (50 iterations each):
  - Convert immediately after async CUDA write (internal sync captures pending writes)
  - Convert after async CUDA broadcast chain
  - Convert output of `@jit` immediately
  - Convert output of `@jit` matmul immediately
- [x] Interleaved CUDA.jl and Reactant operations
- [x] Rapid repeated conversions (both directions)
- [x] Mutation independence (both directions)
- [x] Edge cases: empty arrays, single-element arrays, very large arrays (100MB), high-dimensional arrays (5D), non-contiguous views, boolean arrays


wsmoses:
@maleadt can I get your eyes/thoughts on this?

jacobadenbaum:
@wsmoses I see that some of the CI is failing.  I don't know the setup here, so I'm not sure if this is due to something pre-existing or if this indicates that there's an issue with what I've done.  If you want me to dig into this I'm happy to, but also I don't want to spend a lot of time on it if you already know the issue.

wsmoses:
so basically most of CI assumes the jll [aka shared library from c++ code] is the one deployed in the package manager, so it's actually expected that it fails. When we deploy this we'll need to have this separated into one PR for the C++ change, deploy the change, and then bump the Reactant_jll compat on the built version

wsmoses:
also bump @maleadt for thoughts

ext/ReactantCUDAExt.jl
```diff
@@ -1485,6 +1485,74 @@ end
...
+    cu::DenseCuArray{T,N};
+    client::Union{Nothing,Reactant.XLA.PJRT.Client}=nothing,
+    device::Union{Nothing,Reactant.XLA.PJRT.Device}=nothing,
+) where {T,N}
+    theclient = client === nothing ? Reactant.XLA.default_backend() : client
+    thedevice = device === nothing ? Reactant.XLA.default_device(theclient) : device
+
+    # Synchronize CUDA stream to ensure all writes to cu are complete before
+    # PJRT reads from this memory.
+    CUDA.synchronize()
```
maleadt, line 1533: This only synchronizes the task-local stream of the current device.

test/integration/cuda_d2d.jl
```diff
@@ -0,0 +1,342 @@
...
+
+        ra_empty = Reactant.to_rarray(Float32[])
+        cu_back = CuArray(ra_empty)
+        @test size(cu_back) == (0,)
+        @test Array(cu_back) == Float32[]
+    end
+
+    @testset "Large array (100MB)" begin
+        cu = CUDA.rand(Float32, 25_000_000)  # 100MB
+        CUDA.synchronize()
```
maleadt: Why all these synchronizations in the test? `Array(::CuArray)` is already synchronizing, and AFAIU the work here is supposed to be compatible with arrays that have outstanding kernels, so eagerly synchronizing in the tests hides issues there.

ext/ReactantCUDAExt.jl
```diff
@@ -1485,6 +1485,74 @@ end
...
+#
+# Safety: BufferFromDevicePointer creates an ephemeral PJRT view of cu's memory,
+# then does a D2D copy via CopyToMemorySpace, then awaits the copy's completion
+# (GetReadyFuture().Await()) before returning. This ensures the D2D copy finishes
+# while GC.@preserve still holds cu alive, so the source memory is never read
+# after it could be freed.
+function Reactant.ConcretePJRTArray(
+    cu::DenseCuArray{T,N};
+    client::Union{Nothing,Reactant.XLA.PJRT.Client}=nothing,
+    device::Union{Nothing,Reactant.XLA.PJRT.Device}=nothing,
```
maleadt, line 1514: Maybe add some assert that `device(cu)` matches `device`?

ext/ReactantCUDAExt.jl
```diff
@@ -1485,6 +1485,74 @@ end
...
+    return Reactant.ConcretePJRTArray(CUDA.CuArray(cu); kwargs...)
+end
+
+# ConcreteRArray -> CuArray: D2D copy using AwaitBufferReady for synchronization.
+# AwaitBufferReady calls GetReadyFuture().Await() to wait for all PJRT definition
+# events to resolve, then returns the raw device pointer. We then D2D copy into
+# a new CuArray.
+function CUDA.CuArray(ra::Reactant.ConcretePJRTArray{T,N,1}) where {T,N}
+    buf = Reactant.get_buffer(ra)
+    cu = CUDA.CuArray{T}(undef, size(ra))
```
maleadt, line 1578: This creates an array on the current device. Maybe again validate this matches the device of `buf`?

ext/ReactantCUDAExt.jl
```diff
@@ -1485,6 +1485,74 @@ end
...
+    nbytes = length(ra) * sizeof(T)
+    GC.@preserve ra cu begin
+        # AwaitBufferReady: waits for definition events, returns synchronized device ptr
+        src_ptr = MLIR.API.AwaitBufferReady(buf.buffer)
+        # cuMemcpyDtoD_v2 is synchronous (blocks until copy completes)
+        CUDA.cuMemcpyDtoD_v2(pointer(cu), CUDA.CuPtr{Nothing}(UInt(src_ptr)), nbytes)
```
maleadt: CUDA.jl has wrappers for device-to-device copies (`unsafe_copyto!`).

maleadt (commented):
I'm not familiar with most of the code here, so only some thoughts wrt. the CUDA.jl ops done here.

test/integration/cuda_d2d.jl
jacobadenbaum: Done. I've removed them.

ext/ReactantCUDAExt.jl
jacobadenbaum: fixed.  now in the forward direction `ConcretePJRTArray(cu)` should auto-select the PJRT device that matches the corresponding CUDA device, and asserts a mismatch if you pass the wrong device.  In the reverse direction `CuArray` should assert that the current CUDA device matches the PJRT buffer

ext/ReactantCUDAExt.jl
jacobadenbaum: should be fixed now

ext/ReactantCUDAExt.jl
jacobadenbaum: thanks -- this is fixed to use `unsafe_copyto!` now with a check that `n>0`

ext/ReactantCUDAExt.jl
jacobadenbaum: I'm not entirely sure what to do about this one.  I tried passing `CUDA.stream().handle` directly to `BufferFromDevicePointer` but XLA rejects it.  It only accepts streams XLA created itself or registered via dlpack.  CUDA.jl's streams aren't in that pool.  

I had claude look at what some of the other packages in this space do for this.  It reported:
- PyTorch/CuPy: Use the DLPack protocol — producer records a CUDA event on its stream, consumer's stream waits on that event. Non-blocking, but requires both sides to participate.
- CuPy fallback: When no consumer stream is provided, does a full CPU-blocking synchronize() on the current stream.
- DLPack.jl: No synchronization at all — caller's responsibility.

Since we can't use the event-based DLPack approach (XLA doesn't accept external streams), I've kept CUDA.synchronize() which syncs the current task's stream. This matches CuPy's fallback behavior. The comment documents the limitation and what would need to change.

Since CUDA.synchronize() only syncs the current task's stream. In multi-stream scenarios (e.g., CuArray written on a different task's stream), this wouldn't be sufficient.

I guess there's two options here:
1) We could use CUDA.device_synchronize() instead (syncs all streams on the device).  I suspect this would not be great for performance.  

2) We could leave it  current stream the right default with cross-stream usage documented as a caller responsibility? 

What do you think is the right thing here?

jacobadenbaum:
bump @wsmoses.  I think that this is all working now and we've addressed all of @maleadt's comments

ext/ReactantCUDAExt.jl
maleadt: Leaving it up to the caller is probably best then, albeit a potential footgun. `device_synchronize` is not an option because of performance, indeed.

PR #2669, opened 2026-03-15
Disable/make flag for self_to_conv

wsmoses:


wsmoses:
@avik-pal testing this in https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/23111321744/job/67129161790?pr=2244

comms are reduced from 342 to 334, memory from 10.586 GiB -> 9.626 GiB, time goes up though from [0] second loop: 16.460403 -> second loop: 16.892001 seconds

cc @giordano

concurrently we need to figure out the fusion problems on nvidia gpus, but at least we can toggle this flag

PR #2670, opened 2026-03-16
Add structured_hlo_call

Pangoraw:
the idea is to store the linearization paths in function arguments / results attributes so that we can linearize / de-linearize prior / after an hlo_call from this kind of IR. Makes it possible to store ir on disk or optimize it with a tip enzymexlamlir-opt and then run as is.

## Summary

- Adds `store_args_res_path` compile option that attaches `reactant.path` attributes to each MLIR function argument and result, encoding the Julia object traversal path (struct field indices, arg indices) as an MLIR array attribute
- Adds `structured_hlo_call` which uses these path attributes to automatically linearize nested-struct arguments, call `hlo_call`, and delinearize results back into a nested structure

## Test plan

- [ ] New test file `test/core/structured_hlo_call.jl` covering:
  - IR without the option has no `reactant.path` attributes
  - IR with the option has correct path attributes for nested structs
  - Round-trip via `structured_hlo_call` produces correct results
  - Mutation test with struct fields navigated by integer index

### DIFF
--- a/test/core/structured_hlo_call.jl
+++ b/test/core/structured_hlo_call.jl
@@ -37,16 +37,12 @@ end
     @testset "reactant.path present with the option" begin
         hlo = @code_hlo store_args_res_path = true forward(model, x)
 
-        # Paths are rooted at :args / :result; repr(sym) keeps the leading colon.
-        # Struct fields are navigated by integer index (from make_tracer):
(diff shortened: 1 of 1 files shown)


Pangoraw:
superseded by #2779

PR #2691, opened 2026-03-19
fix: use new xprof

avik-pal:


avik-pal:
This is baadd... Step markers have also disappeared from xprof

PR #2698, opened 2026-03-19
Mention xprof in profiling docs

Pangoraw:


docs/src/tutorials/profiling.md
```diff
@@ -175,6 +175,23 @@ end
...
+It can be installed in the following manner:
+
+```bash
+pip install xprof # or xprof-nightly
+```
+
+Launching xprof is then as simple as:
+
+```bash
+xprof --logdir=./
```
giordano, line 190: Where should `--logdir` point to?  If I point it to the directory which contains the `*.xplane.pb` file, I get this, so I'm quite stuck

Image

Pangoraw: The directory which contains the `plugins/profile/` dir

Pangoraw: i.e. the one passed to `Reactant.with_profiler(dir)` and then xprof should list runs based on their date.

Issue #2701, opened 2026-03-20
'llvm.call' op incorrect number of operands in 1.12


dkytezab:
cc @Pangoraw. Works in 1.11
```julia
using CUDA, Reactant, KernelAbstractions
using Reactant: ConcreteRNumber
using Adapt: Adapt

mutable struct Clock{TT, S}
    time :: TT
    stage :: S
end

Adapt.adapt_structure(to, clock::Clock) = (time          = clock.time,
                                           stage         = clock.stage)

@kernel function _fill!(c, args)
    i = @index(Global)
    @inbounds c[i] = c[i]
end

N = 16
c = Reactant.to_rarray(zeros(N))
clock = Clock(ConcreteRNumber(0.0), 1)

function loss(c, clock)
    backend = KernelAbstractions.get_backend(c)
    _fill!(backend)(c, clock; ndrange=size(c))
    KernelAbstractions.synchronize(backend)
    return 0.0
end

compiled = Reactant.@compile raise=true raise_first=true loss(c, clock)
```
```
error: 'llvm.call' op incorrect number of operands (3) for callee (expecting: 4)
┌ Error: Compilation failed, MLIR module written to /var/folders/km/c7cykv9x5s955_48nqv3f4cm0000gn/T/reactant_QV60I9/module_000_nxc9_post_raise_triton_custom_call_pm.mlir
└ @ Reactant.MLIR.IR ~/Aeolus2/Reactant.jl/src/mlir/IR/Pass.jl:146
```

PR #2704, opened 2026-03-20
Kernel import: handle 1.12 rooting convention

wsmoses:
fix https://github.com/EnzymeAD/Reactant.jl/issues/2701

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -1051,25 +1051,28 @@ function get_field_offset(T::Type, path)
 
     return offset
 end
-function mlir_extract_roots_from_value!(val::MLIR.IR.Value, roots_ptr::MLIR.IR.Value, @nospecialize(argty::MLIR.IR.Type), @nospecialize(njlvaluet::MLIR.IR.Type))
+function mlir_extract_roots_from_value!(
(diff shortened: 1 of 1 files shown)


Pangoraw:
Got the same on 1.11


~~Without raise, i get the following error on the #2701 reproducer:~~

```
ERROR: GPU compilation of MethodInstance for gpu__fill!(::KernelAbstractions.CompilerMetadata{…}, ::ReactantCUDAExt.CuTracedArray{…}, ::@NamedTuple{…}) failed
KernelError: passing non-bitstype argument

Argument 4 to your kernel function is of type @NamedTuple{time::Reactant.TracedRNumber{Float64}, stage::Int64}, which is not a bitstype:
  .time is of type Reactant.TracedRNumber{Float64} which is not isbits.
    .paths is of type Tuple which is not isbits.
    .mlir_data is of type Union{Nothing, Reactant.MLIR.IR.Value} which is not isbits.


Only bitstypes, which are "plain data" types that are immutable
and contain no references to other values, can be used in GPU kernels.
For more information, see the `Base.isbitstype` function.

dkytezab:
Seems to fix the MWE and the Breeze use case (https://github.com/NumericalEarth/Breeze.jl/pull/565)!

PR #2710, opened 2026-03-22
Actually throw plugin error message

wsmoses:


giordano:
Is this why I didn't get errors on OOM?

wsmoses:
no, this is just for plugins like tpu (where instead of throwing a julia error if someone else is using the tpu, rn it crashes)

PR #2718, opened 2026-03-24
Add stablehlo fixup pass 

Pangoraw:
workaround for https://github.com/openxla/stablehlo/pull/2924 used in https://github.com/PRONTOLab/GB-25/pull/258. 

not sure we want to merge :warning:

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -2685,6 +2685,8 @@ function compile_mlir!(
                 ),
                 "to_mhlo_shardings",
             )
+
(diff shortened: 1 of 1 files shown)


avik-pal:
Can we not apply it as a patch for the time being instead of this?

Pangoraw:
> Can we not apply it as a patch for the time being instead of this?

good point, this was mostly to test without having to build a new Reactant_jll

For a patch on Enzyme-JAX: https://github.com/EnzymeAD/Enzyme-JAX/pull/2318

PR #2722, opened 2026-03-24
Copy preserved argument if needed

Pangoraw:
The rationale is that we have a function

func.func (%arg0, %arg1)
  return %arg1, %arg1

we dont even call to xla since the args are removed. But if we disabled this optimization and run with xla each mlir result will have a new Buffer. Therefore, in the optimization I made sure that we copy as necessary to have a new buffer even when we do this optimization.

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -3780,7 +3780,7 @@ function codegen_unflatten!(
     )
 
     # if some argument is mutated, change them to point to the correct concrete results
-    preserved_arg_use_count = zeros(UInt32, length(linear_args))
+    orig_buffer_available = trues(length(linear_args))
     for (pi, (result, arg_idx)) in enumerate(preserved_args)
         paths = (
             (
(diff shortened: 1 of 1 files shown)


src/Compiler.jl
```diff
@@ -3812,12 +3814,34 @@ function codegen_unflatten!(
...
-            for p in argpath[3:end]
-                argres = :(traced_getfield($argres, $(Meta.quot(p))))
+            if isnothing(argpath_value)
+                # this traced array take an identity value derived from another argument but
+                # is its own traced array. As such, it needs to allocate a new buffer instead of using the arg directly.
+                needs_copy =
+                    initial_path[1] === :args &&
+                    argpath[1] === :args &&
+                    initial_path[2:end] != argpath[2:end] &&
+                    count(parg -> parg[2] == arg_idx, preserved_args) > 1
```
Pangoraw: one of these uses is allowed to not copy. I should update that

wsmoses:
@avik-pal can you give this a review?

Issue #2230, opened 2026-01-26
Implement missing sharding support in Tracing.jl


avik-pal:
## Description
Several code paths in Tracing.jl throw `error("TODO: implement sharding")` which need to be implemented.

## Checklist
- [ ] `src/Tracing.jl:1461` - Implement sharding
- [ ] `src/Tracing.jl:1472` - Implement sharding
- [ ] `src/Tracing.jl:1485` - Implement sharding
- [ ] `src/Tracing.jl:1550` - Implement sharding
- [ ] `src/Tracing.jl:1561` - Implement sharding
- [ ] `src/Tracing.jl:1574` - Implement sharding

## Labels
enhancement, sharding

PR #2723, opened 2026-03-24
Attempt to add sharding support to Tracing.jl for PJRT

dkytezab:
see https://github.com/EnzymeAD/Reactant.jl/issues/2230. cc @Pangoraw

### DIFF
--- a/src/TracedUtils.jl
+++ b/src/TracedUtils.jl
@@ -472,16 +472,8 @@ function prepare_mlir_fn_args(
     seen_args0 = OrderedIdDict()
     try
         for i in 1:N
-            arg_sharding = if args[i] isa Reactant.AbstractConcreteArray ||
-                              args[i] isa Reactant.AbstractConcreteNumber
-                sinfo = args[i].sharding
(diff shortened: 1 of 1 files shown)


src/Tracing.jl
```diff
@@ -1498,15 +1498,14 @@ Base.@nospecializeinfer function make_tracer(
...
-                res = ConcretePJRTArray{T,N,1}(
-                    (XLA.PJRT.AsyncEmptyBuffer,), size(prev), Sharding.NoShardInfo()
-                )
-            else
-                error("TODO(#2230): implement sharding")
-            end
+            res = ConcretePJRTArray{T,N,ndev}(
+                ntuple(_ -> XLA.PJRT.AsyncEmptyBuffer, ndev),
+                size(prev),
+                Sharding.NoShardInfo(),
```
wsmoses, line 1507: I don't think this is a fix?

avik-pal, line 1507: Not sure if this can be NoShardInfo

PR #2775, opened 2026-04-06
change `@error` for `@warn` on backend initialization error

mofeing:
As discussed in #2774, it's a lil confusing for users to see an error with backtrace when a backend fails to init. Specially when CPU backend is chosen but a GPU backend fails.

cc @tpapp

### DIFF
--- a/src/accelerators/Registration.jl
+++ b/src/accelerators/Registration.jl
@@ -108,7 +108,8 @@ function initialize_backends(
             state.clients[backend.platform_name] = client
             successful_initializations[i] = true
         catch err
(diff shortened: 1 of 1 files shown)


Pangoraw:
maybe just warn so the color is different? i feel like we lose a lot of information there if not in debug

avik-pal:
Note that this only happens when the installed version is a CUDA/ROCM and then during runtime the device isn't available. As another data-point, jax also shows error if there is a similar discrepancy.

mofeing:
Closed in favor of #2778

PR #2783, opened 2026-04-08
despecialize generic `create_result` method

mofeing:
this pr reduces time taken by `create_result` on first call to `@compile identity(model)` by a x4 factor: from 20% to 5% of total.

since most functions in GB-25 do not return the model, I don't expect to see a huge improvement on it but it's still nice and general.

### DIFF
--- a/src/Compiler.jl
+++ b/src/Compiler.jl
@@ -218,9 +218,9 @@ function traced_setfield_buffer!(
     return traced_setfield!(obj, field, cval, path)
 end
 
-function create_result(
-    tocopy::T,
-    path,
+Base.@nospecializeinfer function create_result(
+    @nospecialize(tocopy),
+    @nospecialize(path::Tuple),
     result_stores,
     path_to_shard_info,
     to_unreshard_results,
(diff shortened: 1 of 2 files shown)


gbaraldi:
At this point I wonder if not using a tuple at all is better. A any array is straight up better if you are doing this

mofeing:
> At this point I wonder if not using a tuple at all is better. A any array is straight up better if you are doing this

I'm exactly working on a version that replace it for a `Vector`, but the biggest improvement in this PR actually comes from `@nospecialize(tocopy)` on the generic method, not the `path`. The `@nospecialize(path::Tuple)` helps a lil, but I'm not seeing overheads wherever it's used.

src/Compiler.jl
```diff
@@ -230,11 +230,13 @@ function create_result(
     result_cache,
     var_idx,
     resultgen_code,
-) where {T}
+)
     if !isstructtype(typeof(tocopy))
         error("cannot copy $tocopy of type $(Core.Typeof(tocopy))")
     end
 
+    T = typeof(tocopy)
```
wsmoses: Core.Typeof

mofeing: done

PR #2801, opened 2026-04-10
Add native ifrt interpolate

wsmoses:


src/Types.jl
```diff
@@ -437,6 +437,161 @@ function ConcreteIFRTArray(
...
+        shard_shape = length.(slice)
+        if interpolation == InterpolationType.Nearest
+            src_idx_ranges = ntuple(N) do dim
+                I_range = slice[dim]
+                N_dim, M_dim = final_grid_size[dim], src_size[dim]
+                H = halo[dim]
+                H_eff = max(1, H)
+                b = N_dim == 1 ? 1 : (N_dim - 2*H_eff + 1)
+                
+                return [begin
```
giordano: Do you actually want to `return` from the function, or it's just the "return" of the `do` block?  If so, I'd remove it, it's confusing (and `return` it's buggy if it's not to return from a function)
```suggestion
               [begin
```

PR #2819, opened 2026-04-18
test: fix qa testing

avik-pal:

PR #2823, opened 2026-04-18
allow unanalyzable `Reactant.InterpolationType` in QA tests

mofeing:
seems like #2801 broke QA tests


avik-pal:
https://github.com/EnzymeAD/Reactant.jl/pull/2819

PR #2540, opened 2026-02-24
Dates extension

maximilian-gelbrecht:
Julia stdlib's `Dates` are currently not traceable as they are hard coded to have `Int64` fields. Unfortunately a lot of functions defined for its datatypes are also defined on these concrete types and not on abstract type. So, we have to reimplement a large part of `Dates` for this to work. This is currently not a complete implementation of really everything that `Dates` does. It's a surprisingly large package. 

My personal implementation for Speedy so far was just a drop-in replacement for `Dates`. This here works different, because it's supposed to work for users just working with regular `Dates`.

I am not that familiar with Reactant yet, so I also have a few questions, if I actually implemented the right things ;) 
Currently I extend `make_tracer` and `traced_type_inner` for all types that may be traced. 

Doing 
```Julia 
using Reactant, Dates 
dt = Dates.DateTime(1999, 12, 27)
res = @jit(dt + Dates.Minute(1))
``` 
Then, `res <: TracedRDateTime`. I guess that's expected, or can we convert this back into the untraced type for the user? Is this even something we would want? 

Aside from this issue, I have to add one or two slightly more complex unit tests and also see if this actually still works with what we need for Speedy or if I need to add more functionality. Folder structure and definitions follow those of `Dates`. 

Edit: Oops, I just `jlfmt` but I think with the wrong style...

Should also resolve #2046 (cc @langestefan)

PR #2826, opened 2026-04-20
More conversions, accessors and i/o for Dates extension

maximilian-gelbrecht:
I added some more functionality to the Dates extension that came up while working on SpeedyWeather (that's mostly running now correctly, but more on that separately). 

This is still not a full implementation of everything `Dates` does. As stated before in #2540, it is a surprisingly large package and with the way it's written, you really need to reimplement most functions unfortunately for Reactant compatibility.


maximilian-gelbrecht:
@mofeing Would you mind having a quick look if this could be merged? No hurry, but at some point I'd like to consolidate my PR over at SpeedyWeather to not include those things defined here anymore.

mofeing (approved):
lgtm

PR #2830, opened 2026-04-22
Cublas Support

dsding2:
Support for https://github.com/EnzymeAD/Enzyme-JAX/pull/2444
Adds support for cublasSgemm, including caching and runtime shape propagation.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -1,4 +1,5 @@
 #include <iostream>
+#include <vector>
 
 #include "mlir-c/IR.h"
 #include "mlir-c/Support.h"
@@ -3413,7 +3414,7 @@ struct ShapeInfo {
 };
 
 std::string
-EncodeShapeInfoStruct(const llvm::SmallDenseMap<int, ShapeInfo> &map) {
+encodeShapeInfoAsString(const llvm::SmallDenseMap<int, ShapeInfo> &map) {
   std::string out;
   bool firstEntry = true;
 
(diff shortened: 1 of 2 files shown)


deps/ReactantExtra/API.cpp
```diff
@@ -3425,28 +3483,106 @@ REACTANT_ABI void reactantXLAExec(LinkableRuntime **__restrict__ lrtP,
...
 
-  if (iter == cache.end()) {
-    MLIRContext context(lrt->registry);
-    RegisterDialects(wrap(&context));
+  mlir::OwningOpRef<mlir::ModuleOp> module(
+      mlir::ModuleOp::create(mlir::OpBuilder(&context).getUnknownLoc()));
+  mlir::OpBuilder builder(module->getContext());
+
+  ParserConfig config(&context, /*verify_after_parse*/ true);
+  if (failed(parseSourceString(modstr, module->getBody(), config))) {
```
wsmoses: this process here is expensive [especially since this is the hot loop function we want with low latency], can we hide this behind the cache?

dsding2: The problem is that the only mechanism I could figure out to encode shape info is through function arg annotations, which then requires parsing to extract. However, you need that extracted shape info to know whether the caching is valid. If there were some other way to pass that information to the JIT pass or at least mark the calls that don't need dynamic shape information to avoid regression, then we could utilize caching.

wsmoses: could we instead potentially add two arguments to reactantXLAExec, ShapeInfo* runtimeArgs, runtimeArgsSize, and have the codegen emit those?

Obviously to do so we should change llvm::SmallVector<int64_t, 2> shape; to a pointer and size to make it nice for a c-api

dsding2: Yeah, that might be possible. I'll try to look into it.

dsding2: The code has been slightly reworked to pass the relevant information as an argument to reactantXLAExec.

PR #2834, opened 2026-04-23
ProbProg: Docs

sbrantq:
Assumes #2831


docs/src/api/probprog.md
```diff
@@ -0,0 +1,353 @@
...
+)
+```
+
+Returns `(updated_rng, value)`. `symbol` becomes the trace address.
+
+### `untraced_call`
+
+Call a probabilistic function without recording its choices:
+
+```julia
```
avik-pal: These should be `@example <tag>` to ensure they are run in the CI

sbrantq: I see. Thanks Avik!

docs/src/api/probprog.md
```diff
@@ -0,0 +1,353 @@
...
+
+Exports from `Reactant.ProbProg`. Names are qualified as `ProbProg.<name>`;
+`using Reactant: ProbProg` unqualifies them. Usage shown in the
+[tutorials](@ref probprog).
+
+## Types
+
+### `Trace`
+
+[Gen-style trace](https://www.gen.dev/docs/stable/ref/core/gfi/#Traces) of one model execution:
```
wsmoses: we can link to gen but we should assume the doc is read by a newcomer who doesn't have any familiarity and explain from scratch

docs/src/api/probprog.md
```diff
@@ -0,0 +1,353 @@
...
+
+```julia
+mutable struct Trace
+    choices::Dict{Symbol, Any}
```
wsmoses: can we link to the trace doc itself rather than recopying?

docs/src/api/probprog.md
```diff
@@ -0,0 +1,353 @@
...
+Paths of symbols used to index into the trace.
+
+```julia
+ProbProg.Address(:slope)
+ProbProg.Address(:outer, :inner, :x) # nested
+```
+
+### `Selection`
+
+[Gen-style selection](https://www.gen.dev/docs/stable/ref/core/selections/) of addresses that is used to specify which random choices should be included in the inference operation. Constructed via [`select`](@ref).
```
wsmoses: same comments here & below

sbrantq:
https://enzymead.github.io/Reactant.jl/previews/PR2834/api/probprog

Thanks changes are in preview

docs/src/tutorials/probprog/index.md
penelopeysm, line 1: Hi @sbrantq many congratulations on your work here! By way of introduction, I've worked on Turing.jl for the past couple of years-ish. I hope it's not too rude of me to comment here, @wsmoses asked me to take a look.

I think at this early stage your docs are looking really thorough in terms of coverage. Docs are definitely something that can & will be iterated on over time, but I think it's most important to get a really good first impression, which is going to come via this index page! On the other pages you'll have a lot more leeway to not be so engaging.

For the examples on this page, I wonder if it would be possible to break it up into separate code blocks plus add some explanatory text to show at each stage what these things do? For example, that would let you move the mention of `generate()` closer to the point where it's actually called. Another benefit of splitting them up is that it lets a reader see what objects these functions return. For example, in this bit,

```julia
trace_tensor, _ = compiled_fn(
    rng, linreg, xs, obs_tensor, constrained_addresses,
    selection, step_size, inverse_mass_matrix,
)

selected_entries = ProbProg.filter_entries_by_selection(tt.entries, selection)
trace            = ProbProg.unflatten_trace(trace_tensor, 0.0, selected_entries, nothing)
```

I'd really like to see what things like `selection`, `trace_tensor`, and `trace` really _are_ (and I don't want to have to run it myself to find out 😄)

If you'd like to see an example of what I mean, here are the docs for a package I wrote: https://pysm.dev/FlexiChains.jl/stable/turing/. I think that this is in spirit similar to the idea of [literate programming](https://en.wikipedia.org/wiki/Literate_programming), or what we'd nowadays call a 'notebook'.

Alternatively, or in addition to that, I wonder to what extent the examples can be simplified? For example, the linreg example in Turing is much shorter:

```julia
using Turing
@model function linreg(x, y)
    slope ~ Normal(0, 2)
    intercept ~ Normal(0, 10)
    for i in eachindex(y)
        y[i] ~ Normal(slope * x[i] + intercept)
    end
end
x = ...
y = ...
sample(linreg(x, y), NUTS(), 500)
```

Now I can tell that there is a lot of irreducible complexity in the present example and it's unlikely that you'll be able to get it down to 10 lines, but I would ask whether there is a need to specify things like the initial mass matrix and step size, or whether the library could just choose or find reasonable defaults for the user. If this isn't possible yet then that's definitely something worth exploring as a future direction 🙂.

More broadly, and not for this PR, I'd suggest thinking about who the likely audience for the docs is, which will help shape the general approach to writing docs. Feel free to come up with your own ideas, but my suspicion is that you're initially going to get a bunch of people who know some other PPL (indeed most likely Turing, given that we're in Julia land) but have no clue how to use Reactant. I think you'd have to bring those people on board and hand hold them through some of the Reactant-specific stuff, for example explain what `@compile` does (in 2026 our attention spans are very short 😅 so I wouldn't assume that they're going to click to the rest of the Reactant docs to read it).

sbrantq: Hi Penelope, nice to meet you and really appreciate your valuable insights! Completely agree with your points here. I will do a few more iterations to restructure the index into smaller cells with more explanations in between (the FlexiChains page is a great reference, thanks for the pointer!), trim down the reducible complexity, and add Reactant-land explanation to better target our audience!

sbrantq:
Rewrote the index page here per Penny's comments above (waiting for CI): https://enzymead.github.io/Reactant.jl/previews/PR2834/tutorials/probprog/

docs/src/tutorials/probprog/index.md
```diff
@@ -0,0 +1,265 @@
...
+obs = ProbProg.Constraint(:ys => ys)
+```
+
+The current implementation requires a bit of boilerplate to flatten the
+`Constraint` into a tensor representation and to extract its address set
+before passing them to the `@compile`'d function below. See [Traces and
+constrained inference](@ref probprog-traces) for details. The resulting
+tensor and address set:
+
+```@setup probprog_index
```
penelopeysm: This looks really great and nice to read through! I'd only suggest showing this code (so use `@example` instead of `@setup`) to make it clear where those things are defined. If you want to suppress the output from the last line, you can add one final line in the code block which is `nothing # hide`, so something like this.

````
```@example probprog_index
code here...
nothing # hide
```
````

But apart from that very minor thing I think it really looks good

File: docs/src/tutorials/probprog/sampling.md [hunk 6]
```diff
@@ -0,0 +1,217 @@
...
+## Built-in distribution
+
+```julia
+using Reactant
+using Reactant: ProbProg, ReactantRNG
+
+seed = Reactant.to_rarray(UInt64[1, 4])
+rng  = ReactantRNG(seed)
+
+_, x = ProbProg.sample(rng, ProbProg.Normal(0.0, 1.0, (10,)); symbol=:x)
```
Red-Portal, line 15: Is there a chance that the `~` notation would be used similarly to Turing? For some reason, I find the `~` notation to be significantly more readable than any other notation used by `Pyro`, `PyMC`, and so on.

docs/src/tutorials/probprog/chains.md
```diff
@@ -0,0 +1,208 @@
...
+initial_position = Reactant.to_rarray(zeros(d))
+
+samples, state = ProbProg.run_chain(
+    rng, logpdf_fn, initial_position;
+    algorithm         = :NUTS,
+    num_warmup        = 1000,
+    num_samples       = 5000,
+    chunk_size        = 500,
+    step_size         = nothing,               # let adaptation pick
+    inverse_mass_matrix = nothing,             # identity to start
```
Red-Portal, line 41: In terms of API design, one thing I would suggest is to precondition  (inverse mass matrix stuff)  the target instead of the algorithm. In Turing-world terms, this would mean that the preconditioner becomes part of the `LogDensityProblem` object. The main upside would be that the algorithms won't need to worry about preconditioning, which can complicate the code for no particular advantage.

This is something that would have been difficult in Turing since a lot of API changes will be needed, but nice to have for a new PPL.

docs/src/tutorials/probprog/chains.md
```diff
@@ -0,0 +1,208 @@
...
+```julia
+using Reactant
+using Reactant: ProbProg, ReactantRNG
+
+seed = Reactant.to_rarray(UInt64[1, 5])
+rng  = ReactantRNG(seed)
+
+function logpdf(θ, X, y)
+    logits = X * θ
+    ll = sum(y .* logits .- max.(logits, 0.0) .- log1p.(exp.(.-abs.(logits))))
```
Red-Portal, line 187: You can fuse all this with `@.` . For some reason, using `@.` often improves performance too. (Wonder why Julia's optimizer can't do this automatically.)

Red-Portal (commented):
Hi, thanks for the amazing work being done here. Similarly, with @penelopeysm, I was involved with `Turing` mainly in developing `AdvancedVI.jl` and `SliceSampling.jl`. I have a few high-level questions/comments regarding where this project is going.

docs/src/tutorials/probprog/chains.md
wsmoses: So due to reactant here this will be equally fast

docs/src/tutorials/probprog/sampling.md
penelopeysm: I was thinking it would be not toooooo hard to write a macro to do this, in fact one could have very Turing-similar syntax but also I didn't want to overload haha 😅.

docs/src/tutorials/probprog/chains.md
Red-Portal: Aha!

marcobonici:
Hi, after chatting a bit with @wsmoses , here are my comments.
I want to emphasize that I am NOT a Turing or PPL developer. I am a physicist and am mostly interested in applications of PPL and differentiable computing (and have used Turing for a long time).
I will not make comments on the desing itself (I am not as expert as @penelopeysm or @Red-Portal ). I will just let you know, from a simple user perspective, what I saw that seems nice to me.

**What I like**

1. External samplers.
2. Conditioning syntax.
3. Resuming chains.
4. Submodels.

**What I miss**

1. `Turing` like syntax (e.g. `~` statement, I just like them a lot).

**What I have not seen (but should be easy to interface with)**

1. `Arviz` integration (in my opynion a must have for a strong bayesian workflow).
2. Integration with `Pathfinder` (in my opynion,  best chains initializer in town).
3. GPU support (but I presume it is present by default given `Reactant` support for hardware accelerators).

As I said to @wsmoses , I am converting some codes of mine in a `Reactant` compatible way (such that this can be tested!). I am very curios to test what you have been working on!
Best,
Marco

docs/src/tutorials/probprog/sampling.md
sbrantq: Yeah that would be very useful indeed! We can add `~` notation in a followup PR

wsmoses (commented):
lgtm for an initial docs for now.

can continue to improve docs, internals, syntax, etc in follow ups!

wsmoses (approved):

PR #2842, opened 2026-04-26
refactor: improve readability of Enzyme.jl with enums

avik-pal:


wsmoses (approved):
tangentially can we get this imported via libMLIR_h.jl or something in the future

PR #2843, opened 2026-04-26
ci: remove python testing on mac and cuda testing on windows

avik-pal:


test/runtests.jl
```diff
@@ -44,6 +44,12 @@ delete!(testsuite, "plugins/metal") # Currently completely non functional
 if Sys.isapple()
     delete!(testsuite, "core/custom_number_types")
     delete!(testsuite, "integration/enzymejax")
+    delete!(testsuite, "integration/python")
+end
+
+if Sys.iswindows()
```
wsmoses, line 50: this is still reasonable, compiling into the cpu backend?

avik-pal: So this was for https://github.com/EnzymeAD/Reactant.jl/actions/runs/24969085082/job/73108844431?pr=2842#step:21:2648

wsmoses: so that's itself an error we should fix [or report to cuda.jl]

wsmoses: we should be able to still do codegen without a cuda driver

wsmoses: cc @giordano and @maleadt since this seems like a CUDA 6 thing

PR #2865, opened 2026-05-06
Adding createLLVMMod and compileMhlotoLLVM into API.cpp

Antipath1:
This adds the functions from clang_compile.cc and compile_with_xla.cc from Enzyme-Jax and gets them a wrapper to be from the C++ side to be useable by Julia. Implementation in a second PR stacked on this since this needs a JLL change and I thought this might make it easier.


deps/ReactantExtra/API.cpp
```diff
@@ -3833,6 +3835,103 @@ REACTANT_ABI void *ReactantGetCompileOptions(size_t *size) {
   return data;
 }
 
+REACTANT_ABI xla::LocalExecutable *
+ReactantCompileMhloToLLVM(const char *mhlo_text, size_t mhlo_text_len,
```
wsmoses, line 3839: is there a reason for doing a string with length rather than cstring [aka null terminated]?

am fine either way

cc @avik-pal since this relates to native llvm codegen

Antipath1: AFAIK Julia already stores strings with its length so this has an admittadly probably extremly small performance benfit saving doing a strlen which is O(n). Also its a bit safer concerning potential buffer overflow although that should not happen anyway but I still prefer to be on the safer side. 

I can change it if it makes things easier but I prefer it with explicit length.

wsmoses (approved):

PR #2875, opened 2026-05-08
WIP: decouple `eltype` from `AbstractArray` for `TracedRArray`

mofeing:
> [!CAUTION]
> DO NOT MERGE! This is a very delicate change that can have huge negative effects.
> This is an experiment for a potential future design.

this should fix the GNNGraph problem in #2856 without needing to introduce `TracedRInteger`, `TracedRAbstractFloat`, `TracedRComplex`, ... types.

### DIFF
--- a/src/Overlay.jl
+++ b/src/Overlay.jl
@@ -205,6 +205,27 @@ for (cT, aT, bT) in (
     end
 end
 
+for (aT, bT) in (
+    (:(LinearAlgebra.StridedMaybeAdjOrTransMat{<:LinearAlgebra.BlasReal}), :(LinearAlgebra.StridedMaybeAdjOrTransMat{<:LinearAlgebra.BlasReal})),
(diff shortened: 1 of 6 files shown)


wsmoses (changes requested):
we should not do this. There is so much code that assumes as an invariant what the eltype is from an abstractarray

mofeing:
this is currently an experiment, so no planning to merge it. but as long as we have `TracedRArray{T} <: AbstractArray{TracedRNumber{T}`, there are gonna be problems.

I'm preparing a hackmd with the design and stuff, so we can discuss if further. but one of things that I've checked out is that as Julia interface docs say, `eltype` is part of the _Iteration_ interface and not a part of _AbstractArray_ (I'm aware that in reality this is a problem of Julia not having proper interfaces and users relying too much on assumptions not present on the docs).

PR #2886, opened 2026-05-10
Proposal for a new distributions interface for ProbProg

ptiede:
First, let me warn everyone that this was heavily coded with Claude's assistance. 

Ok, so given that Distributions.jl hard codes `Real` and `Float64` in a substantial part of its code, we essentially need to start from scratch. This is my attempt at creating what I had hoped `Distributions` would look like. This is largely based on ideas that have been floating around the Julia ecosystem for years, including Bijectors, ChangeOfVariables, and MeasureTheory.

The idea is that many distributions we actually work with are essentially transformations of a simple distribution.
Take, for example, the Normal distribution. `Distributions.jl` makes separate types for both the Normal, MvNormal, LogNormal, and LogitNormal distributions. This has always struck me as strange because we can essentially get all of those for free by first defining a `StdNormal` distribution and a couple of transformations, e.g., `Affine` or `Log`, to recover everything. The nice thing is that we can then apply these transformations to almost everything else, e.g., the T-Distribution, Laplace distribution, etc. 

This is my attempt at trying to generalize this approach. Essentially, I am defining some primitive distributions, e.g., StdNormal, StdExp, StdInverseGamma, etc., and then building everything else on top of that using transformations.  So far, it seems to work all right, but there are a few things that should be worked out. 

1. Do we cleanly split from Distributions.jl? I've followed ProgProg's choice here, and it is very reasonable. The only thing is that a ton of the Julia ecosystem kind of assumes you have some sort of `Distributions.jl` object. We could, of course, define an extension that automatically converts things at trace time, but I'm curious what people's opinions of this are
2. I've had to duplicate a ton of functionality that exists in other packages, e.g., InverseFunctions, ChangeOfVariables, because I didn't want to bring those deps into Reactant. If ProgProg gets separated, we probably want to change that. Similar for the special functions like `erf` where I defined my own internal methods using `opcall` instead of the definitions that existed in the SpecialFunctions.jl extension.
3. I split logpdf into two components, `unnormed_logpdf` and `lognorm`. The reason for this was that in non-reactant land, Distributions.jl often computed the aspects of the log-normalization for every logpdf call, which was always why Distributions.jl was consistently too slow for some of my work. With Reactant, it is likely smart enough to raise those operations out of the loop, but maybe not always?
4. I'm not sure how all of this impacts the lower-level stuff that @sbrantq is working on? Maybe this PR should wait until that work is finished

Finally, I'll not that I am currently going to merge a similar prototype into Comrade so I can more easily utilize some of these features.

Ok, now the general interface. 

### General distribution interface              

```julia
struct MyDist{...} <: AbstractDistribution{N}; ...;
     params...
     dims
end                                                   
  ```
Required methods

```julia
unnormed_logpdf(d, x)   # x-dependent part of log p(x)
lognorm(d)              # x-independent log normaliser
Base.rand(rng, d)       # one variate (scalar or array)                                                
Random.rand!(rng, d, x) # in-place fill (only needed  for N > 0)                                            
Base.size(d)            # ::Dims{N}, structural shape
insupport(d, x)         # Bool                        
```

The unnormed_logpdf / lognorm split is based what is rv dependent and what is constant.
For a normal distribution, lognorm would store things like the determinant of the covariance.

In general, the goal is that only a small number of distributions can get defined and then a wide array of others can be defined with the transformation interface.

 ### Transformation interface

  A transform is a bijection used to wrap one distribution into another (e.g., `LogNormal = LogTransform(Normal)`).                          

To define a transformation, we essentially duplicated (sadness...) the ChangeofVariables functionality.

```julia
  struct MyTransform <: AbstractTransform; ...; end
                                                        
  # Required:
  forward(t, z)           # y = f(z)                    
  inverse(t, y)           # z = f⁻¹(y)       
  logabsdetjac(t, z)      # log|det J_f(z)|, may depend on z                                                  
```

If the Jacobian doesn't depend on z (affine, identity, constant compositions), subtype           
  ConstantJacobianTransform <: AbstractTransform instead
   and additionally implement the no-z constant form:

```
  logabsdetjac(t)         # constant value, cached into lognorm                                               
```
  
TransformedDistribution(base, transform) then represents the law of `y = forward(transform, z)` where ` z ~ base`. 

This is kind of strange currently but right now depending on the jacobian we have two paths
                                             
  - `ConstantJacobianTransform` — Jacobian goes into lognorm(d) (cached across logpdf calls).
  - General `AbstractTransform` — Jacobian goes into unnormed_logpdf(d, y) (recomputed per evaluation).

So far, I've added things like `AffineTransform` `LogTransform`, `LogitTransform`. But again, I'd prefer to utilize `InverseFunctions` for this kind of stuff eventually.


sbrantq:
I think this is strictly better than the existing ad hoc distribution implementations! It would be very beneficial to merge, especially given that this helps with the workload that Paul has at hand.

In the longer term I think there would be a lot of benefits if we move towards a more "within the compiler" direction. Specifically I mean we vendor standard distributions as MLIR primitives (mirroring Paul's code that would be `impulse.std_normal`, `impulse.std_exp`, ...) and compositions/transform ops like `impulse.affine_dist` and `impulse.log_transform_dist` and lower the equivalent of the proposed `AbstractTransform` as a compiler pass. This gives several advantages:

* This eliminates the somewhat duplicated math. We have `AbstractTransform` here for the modeling side and https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/Enzyme/MLIR/Interfaces/TransformUtils.cpp for the inference side (it's currently used for constraint transform that makes the support in ℝⁿ so HMC/NUTS's leapfrog doesn't go wrong). Using a shared transform machinery seems like the convention in multiple PPLs, for example NumPyro implements a bunch of `*Transform` classes (https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/transforms.py) and use the same infrastructure for both distributions (https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/distribution.py) and inference (https://github.com/pyro-ppl/numpyro/blob/8d1511ef46aec3eb2b0c7f0e8cbc22045ebfcf01/numpyro/infer/util.py#L23)
* Adding these as compiler primitives enables cross-language reuse -- say in the future we want Python/C++ front end too we don't need to reimplement the same distribution/transform library or use a third-party distribution library
* This also substantially helps with graph-level probabilistic rewrites -- we don't need to do function-level MLIR annotations or complex IR pattern matching to let our compiler passes know what distribution a Julia function is sampling from, which likely enables a few additional rewrites.

Not a domain expert so would likely miss some caveats! Also I'm curious what @wsmoses and others think

PR #2895, opened 2026-05-12
Multi-process IFRT array for multinode

dkytezab:
Fix for sharded AD

### DIFF
--- a/src/xla/IFRT/Array.jl
+++ b/src/xla/IFRT/Array.jl
@@ -384,6 +384,34 @@ function copy_arrays_to_device_with_sharding(buffers::Vector{Array}, sharding::S
     return dst_arrays
 end
 
-Base.copy(b::Array) = Array(GC.@preserve b begin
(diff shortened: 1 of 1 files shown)


src/xla/IFRT/Array.jl
```diff
@@ -384,6 +384,36 @@ function copy_arrays_to_device_with_sharding(buffers::Vector{Array}, sharding::S
     return dst_arrays
 end
 
-Base.copy(b::Array) = Array(GC.@preserve b begin
-    MLIR.API.ifrt_copy_array(b.buffer)
-end)
+function Base.copy(b::Array)
+    sharding = XLA.sharding(b)
+    if all(XLA.is_addressable, XLA.devices(sharding))
+        return Array(GC.@preserve b begin
```
wsmoses, line 390: cc @avik-pal 

I feel like this is something we should fix within ifrt_copy_array, perhaps?

if not I'm a bit confused by why this fails at the moment

PR #2902, opened 2026-05-16
[MLIR] replace `with_block`, `with_module`, `with_context` for macro-version

mofeing:
i.e. avoid closures

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -3390,7 +3390,7 @@ end
 
     sym_name = Reactant.TracedUtils.__lookup_unique_name_in_module(mod, sym_name)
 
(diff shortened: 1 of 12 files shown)


src/Ops.jl
```diff
@@ -3336,6 +3336,7 @@ We return a NamedTuple with the following fields:
     sym_name::String="mesh",
     location=mlir_stacktrace("mesh", @__FILE__, @__LINE__),
 )
+    @info "1"
```
wsmoses: Remove?

wsmoses (approved):
LGTM minus print

src/Ops.jl
mofeing: ups sorry, I was doing some print debug

PR #2913, opened 2026-05-20
Fix 5-arg mul! for SubArrays

### LINKED ISSUE #2912: 5-arg `mul!(C, A, B, α, β)` into a `SubArray` is missing
The problem is pretty self-explanatory, see code below: 

```Julia 
using Reactant, LinearAlgebra

A = Reactant.to_rarray(rand(Float32, 4, 4))
B = Reactant.to_rarray(rand(Float32, 4, 4))
C = Reactant.to_rarray(rand(Float32, 4, 8))
C_view = view(C, :, 1:4)

# 3-arg mul!(C, A, B) into a SubArray destination — works
@jit mul!(C_view, A, B)

# 5-arg mul!(C, A, B, α, β) into a SubArray destination — fails:
# MethodError: no method matching multiply(::SubArray{TracedRNumber{Float32},2,TracedRArray{Float32,2},...}, ::TracedRArray{Float32,2})
@jit mul!(C_view, A, B, -1.0f0, 1.0f0)
```

Is fixing this as easy as just adding a `materialize_traced_array(C)` to 

https://github.com/EnzymeAD/Reactant.jl/blob/d7761e6d26466b03a8570ad73f5af181f2b8f388/src/stdlibs/LinearAlgebra.jl#L320-L325

Or is this wrong / the materialize avoidable? I suggested this in PR#2913 now.

maximilian-gelbrecht:
Addresses #2912 

I am not sure if this is the right approach, but the tests pass for me locally.

### DIFF
--- a/src/stdlibs/LinearAlgebra.jl
+++ b/src/stdlibs/LinearAlgebra.jl
@@ -318,8 +318,9 @@ function overloaded_mul!(
         if β_is_zero
             res = α_res
         else
+            C_mat = materialize_traced_array(C)
             β_C = @opcall multiply(
-                C, @opcall(fill(Reactant.promote_to(TracedRNumber{T}, β), size(C)))
+                C_mat, @opcall(fill(Reactant.promote_to(TracedRNumber{T}, β), size(C_mat)))
             )
             res = @opcall add(α_res, β_C)
         end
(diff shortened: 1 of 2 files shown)


wsmoses (commented):
yeah the materialize is necessary, but it shouldn't be a performance hit

maximilian-gelbrecht:
Can this be merged then, or shall I test / benchmark something before aside from the added unit test?

PR #2914, opened 2026-05-20
Fix PJRT runtime assert in IFRT traced_type_inner

csvance:
This looks like a simple copy and paste error. The specialization is for IFRT but the assert is for PFRT.


wsmoses (approved):
seems so, have a test that hits this offhand

PR #2922, opened 2026-05-26
Rename `checkpoints` to match Enzyme-JAX

dkytezab:
Enzyme-JAX reads `checkpoint_period` so `Periodic(n)` falls through without this I believe. See https://github.com/EnzymeAD/Enzyme-JAX/blob/d0b1927095c7ed49598353f880e4ae64d7f808d0/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp#L511


Pangoraw (approved):
Good catch thanks!

Issue #2065, opened 2026-01-04
PyTorch StableHLO Support


csvance:
Continuing the discussion: https://discourse.julialang.org/t/node-training-performance-w-lux-vs-torchdiffeq/134860/5

Essentially it would be good if existing models, layers, etc in PyTorch could be converted to StableHLO and used within Reactant. This would reduce the barrier of entry to working with deep learning/neural networks in Julia for people who are heavily invested in the PyTorch ecosystem.

It looks like the easiest way to do this would be using TorchXLA and one of the two methods defined here: https://docs.pytorch.org/xla/master/features/stablehlo.html

Essentially we would just need the glue between `ReactantPythonCallExt` and one of these methods. I would partial to using the exported model approach since it likely is a smaller surface for things to go wrong and does not depend on on Jax.

@wsmoses am I understanding this correctly?

As an aside, it would be extremely useful if we could find a way to automatically bring over the weights from pretrained backbones or with whatever random initialization was used on the PyTorch side.

csvance:
@wsmoses I managed to convert, compile, and execute a PyTorch model to run within Reactant. The interface may need to be different than the Jax version for a few reasons:

1. `@compile` complains about some of the array conversions in `pytorch_to_stablehlo`.
2. PyTorch differs from Jax since it associates modules with parameters. Users are going to want to get the parameters back from the module especially when they are pretrained / already have a good initialization from the PyTorch side. Handling this is what causes problems with `@compile`.

```julia
using PythonCall
using Reactant: Reactant
using OrderedCollections

const torchexportptr = Ref{Py}()
const torchaxexportptr = Ref{Py}()
torchexportptr[] = pyimport("torch.export")
torchaxexportptr[] = pyimport("torchax.export")

# When we convert to StableHLO we get Jax types back for the weights
jax_tensor_to_rarray(x) = Reactant.to_rarray(convert(Array, pyconvert(PyArray, x)))
pytorch_tensor_to_rarray(x) = Reactant.to_rarray(convert(Array, pyconvert(PyArray, x.detach().numpy())))

function ensure_py_import_path()
    path = [pyconvert(String, s) for s in sysptr[].path]
    if pwd() ∉ path
        sysptr[].path.append(pwd())
    end
end

function pytorch_to_stablehlo(pytorch_module::Py, sample_inputs)
    exported = torchexportptr[].export(pytorch_module, sample_inputs)
[...]

PR #2928, opened 2026-05-29
[WIP] PyTorch Import

csvance:
Note: this isn't quite ready for full review yet. 

Implements https://github.com/EnzymeAD/Reactant.jl/issues/2065

It includes monkey patching that is potentially brittle due to an inputs ordering difference in torch.export and torchax. It might be possible to avoid the monkey patch, I'm looking into alternative solutions. For torchscript there likely is not an alternative for the workarounds I implemented here. At the end of the day we are dealing with a format that has been deprecated for over 5 years now.

TODO:

- [x] Investigate potential alternatives to monkey patch for input/param ordering
- [x] Document matmul precision behavior when torchax sees a GPU vs when it doesn't - potential footgun
- [x] FP64 support w/ tests
- [x] Scalar type promotion matches PyTorch w/ Jax x64 mode
- [ ] Find a way to make torch install without requiring CPU version workaround

### DIFF
--- a/ext/ReactantPythonCallExt/pytorch.jl
+++ b/ext/ReactantPythonCallExt/pytorch.jl
@@ -76,6 +76,16 @@ _install_state_order_fix()
 def _torch_to_stablehlo_inputs_first(model, example_inputs, strict):
     blocker = _TritonBlocker()
     sys.meta_path.insert(0, blocker)
+    # Enable jax's 64-bit mode for the lowering so float64 models keep double
+    # precision. jax disables x64 by default and would otherwise canonicalize float64
+    # weights, inputs, and avals to float32, producing an f64/f32 mismatch when
+    # Reactant feeds the original float64 operands to hlo_call. Save and restore the
+    # flag so this stays scoped to the export and does not change global jax behavior
+    # (for example the jax tracing path in pycall.jl). The weights and the exported
+    # program are materialized inside this window, so they retain their dtype after
+    # the flag is restored.
+    prev_x64 = _jax.config.jax_enable_x64
+    _jax.config.update("jax_enable_x64", True)
     try:
         exported = _torch.export.export(model, example_inputs, strict=strict)
         weights, func = _txe.exported_program_to_jax(exported)
@@ -91,6 +101,7 @@ def _torch_to_stablehlo_inputs_first(model, example_inputs, strict):
         jax_export = _jax.export.export(_jax.jit(reordered))((jax_avals,), weights)
         return weights, jax_export, len(jax_avals)
     finally:
+        _jax.config.update("jax_enable_x64", prev_x64)
         try:
             sys.meta_path.remove(blocker)
         except ValueError:
@@ -203,12 +214,15 @@ function pycall_with_torch_export(model::Py, args...)
     matmul_precision = pyconvert(
         String, pyimport("builtins").str(jaxptr[].config.jax_default_matmul_precision)
     )
-    @warn "Tracing a PyTorch model: jax will lower float32 matmuls with \
-           jax_default_matmul_precision=$(matmul_precision). A value of None means jax's \
-           platform default, which can be reduced (TF32-style) precision on a \
-           GPU-initialized jax and will differ from a CPU run. Set \
-           jax_default_matmul_precision=\"highest\" before tracing for full float32 \
-           precision." maxlog = 1
+    @warn "Tracing a PyTorch model: jax is lowering float32 matmuls with \
+           jax_default_matmul_precision=$(matmul_precision), and that precision is frozen \
+           into the exported StableHLO. jax picks it from the platform it detects at \
+           trace time, not the backend Reactant executes on, so the two can disagree: \
+           with GPUs visible jax can bake reduced (TF32-style) precision even if \
+           execution falls back to CPU, and with only CPU detected it bakes full \
+           float32 even if you later run on a GPU (and TF32 itself needs an Ampere or \
+           newer GPU). Set jax_default_matmul_precision=\"highest\" before tracing for \
+           deterministic full float32." maxlog = 1
 
     # Export with strict=false (the non-Dynamo tracer). Besides being required for
     # TorchScript, strict=true routes through Dynamo, whose accelerator/stream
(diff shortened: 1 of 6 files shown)


csvance:
Whenever Jax sees a GPU on startup, it appears to lower matmul precision to something less than FP32. For the test with the linear layer, relerror is around 1e-7 or so, when Jax sees a GPU on startup its around 1e-4, which was causing one of the tests to fail. I'm still getting familiar with how all of the tests are wired up in Reactant and what to expect under different testing conditions.

Also the tests don't seem run with a GPU version of torch installed. I'm trying to figure out a way to automate install of the CPU version via CondaPkg. Installing the same version manually via pip in the environment works.

EDIT: I added a workaround to install the CPU version of torch. Here is the error for the GPU version:

```
ImportError: .../torch/lib/libtorch_cuda.so: undefined symbol: ncclCommResume
```

mofeing:
> Whenever Jax sees a GPU on startup, it appears to lower matmul precision to something less than FP32.

right, you need to do the following:

```python
import jax
jax.config.update("jax_enable_x64", True)
```

ext/ReactantPythonCallExt/pytorch.jl
```diff
@@ -0,0 +1,228 @@
...
+    _torchscript_patched[] = true
+    return nothing
+end
+
+is_torch_module(f::Py) =
+    TORCH_EXPORT_SUPPORTED[] && pyconvert(Bool, pybuiltins.isinstance(f, torchptr[].nn.Module))
+
+is_torchscript_module(f::Py) =
+    TORCH_EXPORT_SUPPORTED[] &&
+        pyconvert(Bool, pybuiltins.isinstance(f, torchptr[].jit.ScriptModule))
```
mofeing: minor thing, but you can skip the Python interpretation and directly call the C API for `isinstance` if you use `pyisinstance` directly

```suggestion
is_torch_module(f::Py) =
    TORCH_EXPORT_SUPPORTED[] && pyisinstance(f, torchptr[].nn.Module)

is_torchscript_module(f::Py) =
    TORCH_EXPORT_SUPPORTED[] && pyisinstance(f, torchptr[].jit.ScriptModule)
```

mofeing (commented):
for the symbol clashes, it's a problem that reappears periodically. I have an idea on how to solve it but @gbaraldi and I are gonna try it first on MLIR.jl

csvance:
I'm going to rework the warning about matmul precision. It shouldn't be written from the assumption that the user is using FP32. We should also clearly communicate Jax's behavior and how what devices it detects impacts the precision used for lowering. If the user wants TF32, they need to have an Ampere or later GPU visible on the machine used for the import.

csvance:
Think I found a way to make the input ordering work without monkey patching. Testing on ~100 different models before pushing changes here.

csvance:
I'm reading that the x64 toggle per model might be an anti pattern / lead to unexpected behavior. I was running into some issues with scalars causing type promotions they shouldn't have been based on the original model semantics. Working on fixing this up now. The goal is the type promotion behavior with the x64 flag enabled should be the same as the original model for the same inputs. 

EDIT 1: we might need to demote  floating point scalars on the PyTorch side to single precision to preserve weak type semantics. If we don't do that Jax will treat weakly typed float scalars as double precision when using the x64 mode, which then causes type promotion to double precision for non scalars.

EDIT 2: Just removed the need for monkey patching + fixed scalar type promotion causing issues with x64 mode.

wsmoses:
alternatively to jax, could we use torchtpu to import the stablehlo directly? https://developers.googleblog.com/torchtpu-running-pytorch-natively-on-tpus-at-google-scale/

cc @GleasonK

csvance:
I think I’ve addressed most of the open issues at this point, aside from the known CUDA symbol problem. The torchtpu path sounds promising. 

Keep in mind with the following I'm not a compiler guy. So sorry if I say something truly asinine 😅

While testing, I ran all of my models through the current PR’s import system and noticed a large jump in compile time compared to how I handle this in my own setup. I suspect it comes from inlining weights as constants on bigger models, particularly once you get past the 1e8 parameter range. I believe the current Jax important code does this as well, which is why I implemented it the same way for the PyTorch path. 

Is inlining weights as constants the best default for general model import, rather than passing them in as arguments? For large models it seems to make compile time explode, and if a user wants to compile the same function across multiple input sizes, baking the weights into the MLIR isn’t ideal unless they’re small enough to be negligible.

For context, I’m building an inference server, and my approach has been to share a single .safetensors as additional model arguments alongside one .mlir per batch size / input shape. In my case it also makes it easy to load weights to GPU on demand.

GleasonK:
> could we use torcht

Once its publicly available, yes you should be able to. Code looks well modularized enough where you just have some layer that exports to mlir and go from there, should be easy enough to swap in

wsmoses (approved):
lgtm, but let @avik-pal and/or @mofeing give it a once over before merging

mofeing:
> ... aside from the known CUDA symbol problem.

Do you mean the `ncclCommReduce` symbol clash?

> I suspect it comes from inlining weights as constants on bigger models, particularly once you get past the 1e8 parameter range.

> Is inlining weights as constants the best default for general model import, rather than passing them in as arguments? For large models it seems to make compile time explode,

I cannot say something for sure without a profile, but I would bet that at that size, the constant propagation / folding mechanism is to blame because it is super simple and running in 1-thread, probably no vectorisation, ...

If it's an option, I would pass the model as arguments. If that's not an option, I would try add a "no-inline" `func.func` that contains and returns the constant.

csvance:
> If it's an option, I would pass the model as arguments. If that's not an option, I would try add a "no-inline" `func.func` that contains and returns the constant.

I think the Jax implementation might be signifigantly less aggressive about inlining than I originally thought. If that's the case I should just be able to make the PyTorch implementation match it, although the problem is its not totally intuitive from a user perspective as to what the order of the inputs should be. Perhaps we could pass the user back a helper function that would take the state dictionary and map it to the correct order. Regardless, it looks like a slightly different interface will be required between the two because of the differences in their semantics (PyTorch uses encapsulated parameters/buffers, Jax uses pure functions). It's obvious what the input ordering should be for a converted Jax function, not so much with PyTorch.

PR #2942, opened 2026-06-03
Emulated complex

wsmoses:


mofeing (approved):
lgtm, but don't we take the size for some `TracedRArray` from MLIR's type inference for some op in Ops? in such case, it would we wrong for emulated complex

PR #2943, opened 2026-06-03
Update Comrade benchmarks to track released versions

ptiede:


ptiede (commented):
This also uses a more stable link for the data downloading

PR #2945, opened 2026-06-04
downgrade HTTP to v1

mofeing:
it's giving problems on pkg registration.


Pangoraw:
it's only used to get the TPU plugin, we could probably use Downloads.jl for that

wsmoses:
cc @csvance

mofeing:
@GunnarFarneback noted that in another registration with the same problem, the common pattern was this.

csvance:
It's probably best to downgrade to HTTP.jl 1.X for now and let the package registration issues get worked out assuming that is the root cause. Downloads.jl could work as a replacement, but it may pull in additional JLL packages that open up potential for a different class of bugs, especially considering how many things Reactant already links with.

Pangoraw:
> Downloads.jl could work as a replacement, but it may pull in additional JLL packages

Downloads.jl is part of the standard library (installed with every julia) whereas HTTP.jl is a julia implementation of the protocol. I agree that it's best to downgrade in the meantime

PR #2972, opened 2026-06-11
Bump ReactantCore version and require it in Reactant

giordano:
#2901 added new API to ReactantCore (Binomial), which is also used in Reactant, but there's no new stable version of ReactantCore available, with the result that end users get a broken Reactant.


wsmoses (approved):

Pangoraw (approved):
Oops good catch thanks

PR #2979, opened 2026-06-12
Ensure raising is set if sharded

wsmoses:


ext/ReactantCUDAExt.jl
```diff
@@ -1438,16 +1438,21 @@ Reactant.@reactant_overlay function CUDA.cufunction(
...
 
+	params = if CUVERSION == 6
+            llvm_sm   = CUDACore.SMVersion(cuda_cap.major,
+				       cuda_cap.minor)
+	    CUDACore.CUDACompilerParams(; sm=llvm_sm, ptx=cuda_ptx)
+	else
+	    CUDACore.CUDACompilerParams(; cap=cuda_cap, ptx=cuda_ptx)
+	end
+
+	@show raising()
```
giordano: Should be removed before merge?

wsmoses: lol yeah, was clearly leftover from my debugging

PR #2984, opened 2026-06-15
Split reads and writes in aliasing optimization

### LINKED ISSUE #2870: Mutable struct aliasing
```julia
using Reactant

mutable struct State{A}
    u  :: A
    u⁰ :: A
end

function step!(state)
    state.u⁰ .= state.u
    state.u = state.u .+ 1.0
    return nothing
end

state_v = State(fill(0.5, 8), fill(0.0, 8))
step!(state_v)
@info "vanilla   u = $(state_v.u)   u⁰ = $(state_v.u⁰)"

state_r = State(ConcreteRArray(fill(0.5, 8)), ConcreteRArray(fill(0.0, 8)))
r_step! = Reactant.@compile step!(state_r)
r_step!(state_r)
@info "reactant  u = $(Array(state_r.u))   u⁰ = $(Array(state_r.u⁰))"
```
Discrepancy between julia and reactant here
```julia
[ Info: vanilla   u = [1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5]   u⁰ = [0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5]
[ Info: reactant  u = [1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5]   u⁰ = [1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5]
```

### LINKED ISSUE #3208: @compile with donated_args=:auto crashes on 2nd call when a mutating function swaps struct fields
## Summary

When `@compile`ing a mutating function whose body swaps two fields of a mutable struct (a common ping-pong buffer pattern in iterative solvers) and then writes into the "new" field in place, the default `donated_args = :auto` donates a buffer that ends up aliased across struct fields. The first call to the compiled thunk succeeds, but a second call to the *same* compiled thunk on the *same* struct crashes with:

```
INVALID_ARGUMENT: Attempt to donate a buffer which is also used by the same call to Execute() (flattened argument 1, replica 0, partition 0, first use: 0). Toy example for this bug: `f(a, donate(a))`.
```

Passing `CompileOptions(; donated_args = :none)` avoids the crash entirely.

This was found while debugging [JuliaDecisionFocusedLearning/CoolPDLP.jl#135](https://github.com/JuliaDecisionFocusedLearning/CoolPDLP.jl/pull/135), where an iterative PDHG solver step does exactly this pointer-swap-then-mutate pattern; I've trimmed it down to the pure-Reactant MWE below. cc @gdalle

## MWE

```julia
using Reactant
using LinearAlgebra
using Random

mutable struct Sol{V}
    x::V
end

mutable struct State{V}
    sol::Sol{V}
    sol_last::Sol{V}
    scratch::V
end

function step!(state::State, A)
    # ping-pong pointer-swap pattern common in iterative solvers
    state.sol, state.sol_last = state.sol_last, state.sol
    x_last = state.sol_last.x
    tmp = mul!(state.scratch, A, x_last)      # in-place matvec into scratch buffer
    state.sol.x .= x_last .+ tmp              # write "new" x into the (former sol_last) buffer
    return state
end

rng = MersenneTwister(0)
n = 8
A = Reactant.to_rarray(randn(rng, Float32, n, n))
x0 = Reactant.to_rarray(ones(Float32, n))
x1 = Reactant.to_rarray(zeros(Float32, n))
scratch = Reactant.to_rarray(zeros(Float32, n))
state = State(Sol(x0), Sol(x1), scratch)

# Compiling WITHOUT donated_args = :none (the default @compile behaviour)
compiled_step! = @compile step!(state, A)

compiled_step!(state, A) # 1st call: succeeds
println("1st call ok, x = ", Array(state.sol.x))
compiled_step!(state, A) # 2nd call: crashes, see error below
println("2nd call ok, x = ", Array(state.sol.x))
```

## Full error

```
1st call ok, x = Float32[1.9792527, 1.7406514, 0.210855, 5.601746, -4.8734264, 0.66939235, 1.401438, 2.8878918]
E0000 00:00:1787304595.561741  172699 common_pjrt_client.cc:2135] ExecutePrepareWithOomRetries failed: INVALID_ARGUMENT: Attempt to donate a buffer which is also used by the same call to Execute() (flattened argument 1, replica 0, partition 0, first use: 0). Toy example for this bug: `f(a, donate(a))`.
=== Source Location Trace: ===
external/xla/xla/util.h:303
external/xla/xla/pjrt/common_pjrt_client.cc:1248
external/xla/xla/pjrt/common_pjrt_client.cc:1954

ERROR: LoadError: INVALID_ARGUMENT: Attempt to donate a buffer which is also used by the same call to Execute() (flattened argument 1, replica 0, partition 0, first use: 0). Toy example for this bug: `f(a, donate(a))`.
```

## Confirmed workaround

Passing `CompileOptions(; donated_args = :none)` to `@compile` avoids the crash and gives correct, stable results over repeated calls:

```julia
compile_options = CompileOptions(; donated_args = :none)
compiled_step! = @compile compile_options=compile_options step!(state, A)
```

## Expected behavior

Either:
- `donated_args = :auto` should not donate a buffer that is (or becomes, through a struct-field swap) aliased by another argument slot used by the same call, or
- if that can't be detected safely in general, `@compile` should surface a clearer error/warning at *compile* time rather than letting the thunk crash non-deterministically on a later call.

## Environment

- Reactant.jl: v0.2.280
- Julia: 1.12.6
- OS: Linux x86_64
- Backend: CPU (PJRT)

Pangoraw:
the optimization that bypasses XLA was reading from args, but if the
value of the argument is over-written from being update the computation
then using the previous value would read the wrong value (already updated).

Closes #2870.


wsmoses:
@Pangoraw can you explai the difference in behavior here, in terms of the code_typed of the generated thunk code? [and/or change to the mlir ]

Pangoraw:
if you take a look at the reproducer:

```julia
mutable struct State{A}
    u::A
    u⁰::A
end

function step!(state)
    state.u⁰ .= state.u
    state.u = state.u .+ 1.0
    return nothing
end
```

the previous pseudo code_lowered would look something like this:


```julia
# u⁰ is donated
linearized_results = XLA.execute_sharded(thunk.exec, thunk.device, (state.u, state.u⁰))

# Unpack result buffers
result_buffer_1 = linearized_results[1]

result = nothing

# Build a cache mapping traced arrays/numbers -> concrete PJRT arrays/numbers
cache_dict = IdDict{Union{TracedRArray, TracedRNumber}, Union{ConcretePJRTArray, ConcretePJRTNumber}}()

# Write result buffer back into state.u
traced_setfield_buffer!(Val{:PJRT}(), cache_dict, result_buffer_1, args[1], 1, (1,))

# Snapshot field 1's data AFTER the result has been written back
argpath_value1 = map(copy, state.u.data)

# Propagate that snapshot into field 2 of args[1] — twice
# once because of (:args, 1, 2) <- (:args, 1, 1)  
traced_setfield!(state.u⁰, :data, argpath_value1, (2,))

# once because of (:resargs, 1, 2) <- (:args, 1, 1)
traced_setfield!(state.u⁰, :data, argpath_value1, (2,))

# therefore state.u⁰ is equal to state.u POST execution
```

with this pr this looks something like this:

```julia
# u⁰ is donated
linearized_results = XLA.execute_sharded(thunk.exec, thunk.device, (state.u, state.u⁰))

result_buffer_1 = Base.getindex(linearized_results, 1)

# read result 1 value before overwriting
argpath_value1 = copy(state.u.data)

# Write result buffer back into state.u
Reactant.Compiler.traced_setfield_buffer!(
  Val{:PJRT}(), cache_dict, result_buffer_1, args[1], 1, (1,))

# (:resargs, 1, 2) <- (:args, 1, 1)  
traced_setfield!(state.u⁰, :data, argpath_value1, (2,))

# noop since already written before (:args, 1, 2) <- (:args, 1, 1)

# therefore state.u⁰ is equal to state.u PRE execution
```

wsmoses:
If we need two different results corresponding to the same logical value I think my preference here would for execute_sharded to return that number of the results [aka mlir should undo the redundant return

wsmoses:
thinking more, I think I'd be fine disbaling [or perhaps changing to a flag, disabled by default] the redundant return elimination opt we have

giordano:
@avik-pal any chance we can get a review for this PR?

gdalle:
Gentle bump here following #3208 :)

PR #2988, opened 2026-06-19
ReactantExtra: Add option to use local Enzyme-JAX folder

Pangoraw:
More convenient than having to push + updating the commit hash.


wsmoses (approved):
lgtm, but presumably something needs to be done about the patches?

PR #3001, opened 2026-06-22
Disable attributor in second SROA wrappers invocation

Pangoraw:
The rationale is that attributes from the first SROA attributor should be attached to the llvm.funcs from the first attributor run and that attributor can be long to run on big kernels (shaves 110s compile time from https://github.com/NumericalEarth/Breeze.jl/pull/687 for example).

### DIFF
--- a/src/compiler/OptimizationPasses.jl
+++ b/src/compiler/OptimizationPasses.jl
@@ -8,7 +8,7 @@ const DEBUG_KERNEL = Ref{Bool}(false)
 const DUMP_LLVMIR = Ref{Bool}(false)
 const DUMP_FAILED_LOCKSTEP = Ref{Bool}(false)
 const OpenMP = Ref{Bool}(true)
-const SROA_ATTRIBUTOR = Ref{Bool}(true)
+const SROA_ATTRIBUTOR = Ref{Union{Bool,Symbol}}(:first_sroa_only)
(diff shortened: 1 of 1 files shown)


src/compiler/OptimizationPasses.jl
```diff
@@ -117,21 +117,36 @@ function optimization_passes(
     end
     if sroa
         push!(passes, "propagate-constant-bounds")
+
+        first_sroa_attributor = if SROA_ATTRIBUTOR[] === true || SROA_ATTRIBUTOR[] === :first_sroa_only
```
wsmoses: you aren't changing attributor here, but sroa [likely not the intention]

Pangoraw: i want to put this to false (defaults to true)

https://github.com/EnzymeAD/Enzyme-JAX/blob/a898c7bd6458674846b9df47842be03b163c93c8/src/enzyme_ad/jax/Passes/SROAWrappers.cpp#L161-L162

wsmoses: yeah but you're setting sroa=false not attributor=false

Pangoraw: ooh yeah thanks!

wsmoses (approved):

PR #3006, opened 2026-06-25
Change default SROA_ATTRIBUTOR to true

Pangoraw:
Output IR is not consistently identical (see https://github.com/NumericalEarth/Breeze.jl/pull/802#discussion_r3473482880)


giordano (approved):
Looks good to me if this makes things work out-of-the-box, but the fact that we got a random floating point exception during compilation is worth investigating, right?  A floating point exception has to be enabled specifically 😕

wsmoses:
yeah the fact this creates an observable side effect is itself a bug worth investigating, but happy to merge regardless

wsmoses:
likely some pass [llvm side? (and thus requiring the xla update for fix)] does some incorrect or unchecked math

wsmoses:
if you have the dumps, please open an issue with

PR #3008, opened 2026-06-26
CUDAExt: Set GPUCompiler.current_job to the job

Pangoraw:
Needed since some passes use that global variable, notably lower cpu features (have_fma).

https://github.com/JuliaGPU/GPUCompiler.jl/blob/03ae9a5a6ac7f69658a71dc20cef3e25849deb1e/src/optim.jl#L429

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -693,6 +693,8 @@ function compile(job)
         end
         opt_level = 2
         tm = GPUCompiler.llvm_machine(job.config.target)
+
+        GPUCompiler.current_job = job
         LLVM.@dispose pb = LLVM.NewPMPassBuilder() begin
             LLVM.register!(pb, GPULowerCPUFeaturesPass())
             LLVM.register!(pb, GPULowerPTLSPass())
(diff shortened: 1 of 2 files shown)


wsmoses (commented):
@maleadt is this the right way to do things?

ext/ReactantCUDAExt.jl
```diff
@@ -693,6 +693,8 @@ function compile(job)
         end
         opt_level = 2
         tm = GPUCompiler.llvm_machine(job.config.target)
+
+        GPUCompiler.current_job = job
```
wsmoses, line 699: @Pangoraw can you save the old current job and restore

wsmoses: what we should repurpose

```
src/driver.jl
85:    # save/restore the `current_job` global that passes read: compilation nests (the relink
88:    global current_job
89:    prev_job = @isdefined(current_job) ? current_job : nothing
130:        current_job = prev_job
```

Pangoraw: i did not find a function that we could use (just set and restore current_job)

wsmoses: yeah that's fine for now

wsmoses (approved):

PR #3017, opened 2026-06-30
Disable argument preservation optimization

Pangoraw:
we can see if there are potential unwanted side effects of disabling, and remove the optimization altogether if there are none.

But this optimization has caused a few issues in the past
 - #2984 
 - #2722


wsmoses (commented):
can we have this as an optional optimization, rather than full disable rn?

Pangoraw:
hitting test failures with an unconditional deactivation of the optimization due to #3023.

PR #3018, opened 2026-06-30
Check if current_job global is defined

vchuravy:
GPUCompiler.jl will remove `current_job` entirely.

https://github.com/JuliaGPU/GPUCompiler.jl/commit/04e9aa9499aac962080b4f85d655db8ebc5bd6dc

So harden Reactant usage so that it shouldn't run into issues. Someone who is more involved with Reactant should double check that if passes are called from GPUCompiler that those now take a `job` argument.

### DIFF
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -701,12 +701,54 @@ function compile(job)
 
         try
             LLVM.@dispose pb = LLVM.NewPMPassBuilder() begin
-                LLVM.register!(pb, GPULowerCPUFeaturesPass())
-                LLVM.register!(pb, GPULowerPTLSPass())
-                LLVM.register!(pb, GPULowerGCFramePass())
-                LLVM.register!(pb, AddKernelStatePass())
-                LLVM.register!(pb, LowerKernelStatePass())
-                LLVM.register!(pb, CleanupKernelStatePass())
+                LLVM.register!(
+                    pb,
+                    if isdefined(GPUCompiler, :CPUFeatures)
+                        GPULowerCPUFeaturesPass(job)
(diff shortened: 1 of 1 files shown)


wsmoses (approved):

ext/ReactantCUDAExt.jl
Pangoraw: In this case, the job should be passed here. I can do in a follow-up.

wsmoses: we can use something like isdefined(GPUCompiler, :LowerPTLS) to see if thes truct exists [and similar]

wsmoses (approved):

ext/ReactantCUDAExt.jl
```diff
@@ -694,18 +694,61 @@ function compile(job)
...
+                    if isdefined(GPUCompiler, :LowerGCFrame)
+                        GPULowerGCFramePass(job)
+                    else
+                        GPULowerGCFramePass()
+                    end,
+                )
+                LLVM.register!(
+                    pb,
+                    if isdefined(GPUCompiler, :AddKernelState)
+                        AddKernelStatePass(job)
```
vchuravy: I originally thought so as well, but note that while this has the same name as the GPUCompiler one, it is owners by Reactant https://github.com/EnzymeAD/Reactant.jl/blob/1fde62a4c29beebdc55e3f17bfe3d42e95e21af9/ext/ReactantCUDAExt.jl#L669

wsmoses: oh yeah this one can go back then

Pangoraw:
thanks! confirmed that test pass with GPUCompiler#main

PR #3024, opened 2026-07-02
[linalg] fix `axp[b]y` with `TracedRNumber` scalars

simeonschaub:
Previously, this would call `T(α::TracedRNumber)` which throws a MethodError when `T` is not a `TracedNumber`

### DIFF
--- a/src/stdlibs/LinearAlgebra.jl
+++ b/src/stdlibs/LinearAlgebra.jl
@@ -514,9 +514,6 @@ function LinearAlgebra._kron!(C::AnyTracedRMatrix, A::AnyTracedRMatrix, B::AnyTr
     return C
 end
 
-_convert(T, x) = T(x)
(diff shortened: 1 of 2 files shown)


src/stdlibs/LinearAlgebra.jl
```diff
@@ -538,8 +541,8 @@ function LinearAlgebra.axpby!(
             ),
         )
     end
-    ax = @opcall multiply(x, Reactant.broadcast_to_size(T(α), size(x)))
-    by = @opcall multiply(y, Reactant.broadcast_to_size(T(β), size(y)))
+    ax = @opcall multiply(x, Reactant.broadcast_to_size(_convert(T, α), size(x)))
```
avik-pal: This is the more common pattern in the codebase

```julia
T1 = unwrapped_eltype(T)
α = Reactant.promote_to(TracedRNumber{T1}, α)
```

avik-pal (approved):

PR #3030, opened 2026-07-08
Add PDMats extension

oschulz:
PDMats' scalar accumulation loops don't trace, use broadcasted reductions instead.

Came across this while making MGVI.jl Reactant-compatible (in flight, depends on this PR).

The `@trace` loop change that was originally part of this PR moved to #3055, as requested in review; this PR now contains only the PDMats extension (the two are independent).

AI-generated (Claude Fable at max effort), human reviewed.


oschulz:
@wsmoses is this acceptable, in principle?

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -397,6 +397,17 @@ function trace_while(expr; track_numbers, mincut, checkpointing, first_arg=nothi
     union!(external_syms, cond_symbols.assignments)
     union!(external_syms, body_symbols.references)
     union!(external_syms, body_symbols.assignments)
+    # Locally defined callables (e.g. closures over traced values) that are
+    # called in the loop must become loop-carried arguments as well, so that
+    # traced values captured by them become while-loop operands instead of
+    # closure captures of the generated body function (which would result in
+    # region block arguments without matching operands). Only plain names can
+    # refer to local callables:
+    for symbols_state in (cond_symbols, body_symbols)
```
wsmoses: @Pangoraw

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -397,6 +397,17 @@ function trace_while(expr; track_numbers, mincut, checkpointing, first_arg=nothi
...
+        for fn in symbols_state.funccalls
```
wsmoses: @oschulz the pdmats extension looks pretty easy to give a quick approve to, this part here requires more scrutiny

Pangoraw: please make another pr for that change if not strictly required here. Also I would maybe apply that only if `!isdefined(__module__, only(fn.parts))` to limit the number of arguments to each closures

oschulz: Taking a look at this now.

oschulz: Done: this PR now contains only the PDMats extension; the `@trace` loop change moved to #3055 with the suggested `isdefined(__module__, name)` guard applied. Note the revert commit here is append-only history — the net diff is just the extension.

oschulz:
@wsmoses good like this now?

wsmoses (approved):

oschulz:
Thanks @wsmoses !

PR #3031, opened 2026-07-08
bump Reactant_jll to 0.0.392

mofeing:


mofeing:
aside from the regular python test fails on macOS and windows, and the new bug in XLA, everything passes so merging and doing a new release

wsmoses:
sg!

PR #3032, opened 2026-07-08
define `REACTANT_TOOLCHAIN_IS_BB` on ReactantBuilder/Yggdrasil toolchains

mofeing:


wsmoses (approved):
not sufficiently familiar with bazel so we should double check it works before merging, otherwise lgtm

PR #3033, opened 2026-07-08
adapt to XLA changes

mofeing:


.github/workflows/CommonCI.yml
```diff
@@ -219,6 +219,7 @@ jobs:
             "Reactant", "xla_runtime" => "IFRT"; force=true
           )
         shell: julia --color=yes --project=. {0}
+
```
mofeing:
```suggestion
```

mofeing:
tests might fail due to the sequential merging process but it has been tested that tests pass with correct repo commits (which will be ok after the sequential merging process)

PR #2170, opened 2026-01-21
Separate Reactant number types

oschulz:
Addresses #1570.

~~Built on top of #3034.~~ [obsolete, #3034 is merged]

PR #3034, opened 2026-07-08
Groundwork for number type separation

oschulz:
Basis for #2170 .

### DIFF
--- a/src/ConcreteRArray.jl
+++ b/src/ConcreteRArray.jl
@@ -725,8 +725,8 @@ for aType in (:ConcretePJRTArray, :ConcreteIFRTArray)
         end
 
         function Base.copyto!(
-            dest::SubArray{TracedRNumber{T1},<:Any,<:$(aType)},
-            src::SubArray{TracedRNumber{T2},<:Any,<:Array},
+            dest::SubArray{<:TracedRNumber{T1},<:Any,<:$(aType)},
+            src::SubArray{<:TracedRNumber{T2},<:Any,<:Array},
         ) where {T1,T2}
             throw(MethodError(copyto!, (dest, src)))
         end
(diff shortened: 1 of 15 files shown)


oschulz:
Test failures are unrelated, I think.

src/stdlibs/LinearAlgebra.jl
```diff
@@ -81,29 +81,29 @@ include("factorization/Factorization.jl")
...
+    x::Transpose{<:TracedRNumber{T},<:AnyTracedRArray}
 ) where {T}
     px = materialize_traced_array(parent(x))
     A = ndims(px) == 1 ? reshape(px, :, 1) : px
     return permutedims(A, (2, 1))
 end
 
 function ReactantCore.materialize_traced_array(
-    x::Adjoint{TracedRNumber{T},<:AnyTracedRArray}
+    x::Adjoint{<:TracedRNumber{T},<:AnyTracedRArray}
```
wsmoses: @oschulz this doesn't need to have the specialization on T [as it's not used], can you remove that here [and similar]

oschulz: Addressed in a new commit. Two exceptions: dropping T from the generic_trimatdiv!/generic_mattridiv! Adjoint/Transpose variants would have made them ambiguous against their own T-bound siblings.

oschulz:
@wsmoses is this fine as-is now?

src/ConcreteRArray.jl
```diff
@@ -484,6 +484,13 @@ end
 
 Base.similar(a::ConcretePJRTArray, dims::Dims) = similar(a, eltype(a), dims)
 
+# A concrete array asked for a traced element type (e.g. via LinearAlgebra's
+# generic `similar`/`\` while tracing) cannot hold traced values in its device
+# buffer, so produce a traced array instead.
+function Base.similar(::ConcretePJRTArray, ::Type{S}, dims::Dims) where {S<:TracedRNumber}
```
wsmoses, line 490: this is a bit sketchy, but fine

wsmoses (approved):
looks fine, but lets make sure everything is green [at least that should be, windows iirc has a cuda issue (which we should separately fix to make debugging prs like this easier)]

oschulz:
Thanks @wsmoses! I'll rebase #2170 .

PR #3037, opened 2026-07-09
Fix if statement variable re-assignemnt

### LINKED ISSUE #3036: rng state is dropped
```julia
function _rand_foo3(rng, α)
    T = typeof(α)
    # shape < 1 via the boost identity `Gamma(a) = Gamma(a+1) · U^(1/a)`.
    boost = one(T)
    @trace if α < one(T)
        boost = rand(rng, T)^inv(α)
    end
    # ensures dv is Traced
    dv = zero(T)
    i = 0
    @trace while (dv == zero(dv))
        x = rand(rng, T) - 0.5
        dv = ifelse(x > 0, x, dv)               # keep the first accepted draw
    end
    return boost * dv
end

rng = Reactant.ReactantRNG()
a = ConcreteRNumber(
rg = @compile _rand_foo3(rng, a)
rg(rng, a) == rg(rng, a) # true
```

wsmoses:
cc @ptiede fixes https://github.com/EnzymeAD/Reactant.jl/issues/3036

from our llm overlords:

Root Cause: Unintended Variable Shadowing by @trace if
In @trace if (

ReactantCore.jl:L723
), all_output_vars determines which variables are returned from the @trace if macro expansion and assigned back into the enclosing scope:

julia
($(all_output_vars...),) = $(traced_if)(
    $(cond_name),
    $(true_branch_fn_name),
    $(false_branch_fn_name),
    $(args_expr);
    ...
)
Before this fix, all_output_vars was defined as:

julia
all_output_vars = all_true_branch_vars ∪ all_false_branch_vars
where all_true_branch_vars included both variables assigned inside the if block (true_branch_assignments) and variables only referenced/read inside the block (true_branch_input_list).

How this broke _rand_foo3 (α >= 1.0 vs α < 1.0)
In your snippet:

julia
@trace if α < one(T)
    boost = rand(rng, T)^inv(α)
end
Inside if α < one(T), rng is referenced (rand(rng, T)). Because inputs were included in all_output_vars, @trace if returned rng as part of its output tuple ((boost, rng)).
When @trace if compiled (if_condition in 

Ops.jl:L2450-L2883
), tracing the branches created new tracer objects (ReactantRNG / TracedRArray copies) for rng.
In-place mutations to inputs across conditional branches are already handled separately via the :resarg mechanism (if_condition updates the .mlir_data on the original rng.seed argument passed into the function).
However, because rng was included in all_output_vars, the assignment:
julia
(boost, rng) = traced_if(...)
re-assigned the local variable rng inside _rand_foo3 to the branch copy returned by traced_if.
Consequently, any subsequent operations—specifically @trace while (dv == zero(dv))—operated on and mutated that local copy (tb_rng).
When _rand_foo3 completed, finalize_mlir_fn inspected the original argument (traced_args[1].seed) to determine its final mutated state (:resarg 1). Because @trace while mutated the copy rather than the original rng, traced_args[1].seed.mlir_data was left pointing to the state immediately after the @trace if block—completely losing the RNG seed update from the while loop when α >= 1.0.
The Fix
By changing line 723 in 

ReactantCore.jl:L723
 to:

julia
all_output_vars = unique(true_branch_assignments ∪ false_branch_assignments)
@trace if only returns and reassigns variables that are explicitly assigned inside either branch (true_branch_assignments ∪ false_branch_assignments).
Input arguments that are merely referenced/mutated (like rng) are no longer shadowed by a returned copy.
The original local variable rng (traced_args[1]) remains intact after @trace if, so @trace while operates on traced_args[1].
Both conditional branch mutations (@trace if via :resarg) and loop mutations (@trace while) correctly accumulate on traced_args[1].seed, ensuring rg(rng, a) == rg(rng, a) is false across all branches.

### DIFF
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -666,7 +666,6 @@ function trace_if(expr; store_last_line=nothing, depth=0, track_numbers)
     true_branch_assignments = [
         ExpressionExplorer.compute_symbols_state(true_block).assignments...
     ]
-    all_true_branch_vars = true_branch_input_list ∪ true_branch_assignments
     true_branch_fn_name = gensym(:true_branch)
 
     else_block, discard_vars, _ = if length(expr.args) == 3
@@ -716,14 +715,13 @@ function trace_if(expr; store_last_line=nothing, depth=0, track_numbers)
     false_branch_assignments = [
         ExpressionExplorer.compute_symbols_state(false_block).assignments...
     ]
-    all_false_branch_vars = false_branch_input_list ∪ false_branch_assignments
     false_branch_fn_name = gensym(:false_branch)
 
-    all_input_vars = true_branch_input_list ∪ false_branch_input_list
+    all_input_vars = unique(true_branch_input_list ∪ false_branch_input_list)
     all_output_vars = unique(true_branch_assignments ∪ false_branch_assignments)
     discard_vars !== nothing && setdiff!(all_output_vars, discard_vars)
 
-    all_vars = all_input_vars ∪ all_output_vars
+    all_vars = unique(all_input_vars ∪ all_output_vars)
 
     true_branch_fn = :(
         (args,) -> begin
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -720,7 +720,7 @@ function trace_if(expr; store_last_line=nothing, depth=0, track_numbers)
     false_branch_fn_name = gensym(:false_branch)
 
     all_input_vars = true_branch_input_list ∪ false_branch_input_list
-    all_output_vars = all_true_branch_vars ∪ all_false_branch_vars
+    all_output_vars = unique(true_branch_assignments ∪ false_branch_assignments)
     discard_vars !== nothing && setdiff!(all_output_vars, discard_vars)
 
     all_vars = all_input_vars ∪ all_output_vars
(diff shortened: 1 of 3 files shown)


jumerckx:
I can look more properly when I'm at my pc tomorrow, but this doesn't capture mutation, I think?

wsmoses:
yeah I didn't look too closely into it, feel free to actual take over this pr/issue investigation

jumerckx:
Yes, can do tomorrow. ~~Paul gave a nice approach during the call, will try it out~~

lib/ReactantCore/src/ReactantCore.jl
Pangoraw: might need unique too here?

or convert to vector later.

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -720,7 +720,7 @@ function trace_if(expr; store_last_line=nothing, depth=0, track_numbers)
     false_branch_fn_name = gensym(:false_branch)
 
     all_input_vars = true_branch_input_list ∪ false_branch_input_list
```
Pangoraw: unique as well?

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -720,7 +720,7 @@ function trace_if(expr; store_last_line=nothing, depth=0, track_numbers)
...
-    all_output_vars = all_true_branch_vars ∪ all_false_branch_vars
+    all_output_vars = unique(true_branch_assignments ∪ false_branch_assignments)
     discard_vars !== nothing && setdiff!(all_output_vars, discard_vars)
 
     all_vars = all_input_vars ∪ all_output_vars
```
Pangoraw: unique?

Pangoraw (approved):

jumerckx:
I'd misunderstood what this was doing (thought this was the same as the globals issue)

jumerckx:
@Pangoraw unless you're now looking into this I'll test if the `unique`s don't break things.

Pangoraw:
Yes go ahead, thanks!

jumerckx (approved):

wsmoses:
I haven't looked closely at this, but @jumerckx / @Pangoraw is this the right fix [if so will merge]

jumerckx:
I think so, yes

PR #3049, opened 2026-07-10
Support Static.jl number types in traced code

oschulz:
Stumbled on this while trying to make MeasureBase.jl Reactant-compatible:

StaticBool/StaticInt/StaticFloat64 are Number subtypes outside ReactantPrimitive, so e.g. x + static(0.0) with a traced x tried to construct TracedRNumber{StaticFloat64{0.0}} and threw a TypeError. The extension maps each static type to its dynamic counterpart via unwrapped_eltype and promote_rule (analogous to the existing AbstractIrrational special-casing), plus a promote_to method that converts through dynamic(x), since Static types don't support direct construction like Float64(static(1.0)).

Code changes generated by Claude Fable at max effort.

Related to (but independent from) #3050 .


ext/ReactantStaticExt.jl
```diff
@@ -0,0 +1,25 @@
+module ReactantStaticExt
+
+using Static: StaticBool, StaticInt, StaticFloat64, dynamic
+using Reactant: Reactant, TracedRNumber
+
+for (ST, T) in ((StaticBool, Bool), (StaticInt, Int), (StaticFloat64, Float64))
+    @eval begin
+        Reactant.unwrapped_eltype(::Type{<:$ST}) = $T
```
wsmoses, line 8: why do we define the unwrapped_eltype this way?

oschulz: Claude Fable says:

`unwrapped_eltype` is what the generic tracing paths use to decide the element type of the `TracedRNumber`/`TracedRArray` they construct (the promotion paths in TracedRNumber.jl / TracedPromotion.jl, the rand overlays, ...). Its fallback for plain `Number`s is identity, so a static number leaks its singleton type into the traced type — `TracedRNumber{StaticFloat64{0.0}}` — which is the TypeError this PR fixes. Overriding it declares that static numbers lower to their dynamic counterparts under tracing, so all generic paths do the right thing; it plays the same role as the inline `AbstractIrrational → Float64` special-case in TracedPromotion.jl, which an extension can't add branches to. The instance methods are there because the instance fallback `unwrapped_eltype(::T) where {T<:Number} = T` doesn't forward to the `Type` method.

That said, Static.jl already encodes exactly this mapping as `Base.eltype` (`eltype(StaticFloat64) === Float64`, and Static's own `promote_rule`s go through it), so I can simplify to `Reactant.unwrapped_eltype(T::Type{<:StaticNumber}) = eltype(T)` instead of restating it — happy to update the PR that way.

oschulz: Which do you prefer @wsmoses ?

wsmoses: okay sure this is fine as is then

wsmoses (approved):

PR #3050, opened 2026-07-10
Fix broadcasting functions with constant results over traced arrays

oschulz:
Stumbled on this while trying to make MeasureBase.jl Reactant-compatible:

when a broadcasted function ignores its arguments, elem_apply returns a plain constant. _copyto! promoted that constant to a single-element traced array and stamped its data into a destination of a different size, producing invalid IR:

`func.call` op operand type mismatch: expected operand type `tensor<10xi1>`, but provided `tensor<1xi1>`
Minimal repro: @jit (x -> all(broadcast(xj -> true, x)))(x) — a shape that comes up naturally in vectorized support checks (insupport-style masks). _copyto! now broadcasts non-traced constants to the destination size.

Code changes generated by Claude Fable at max effort.

Related to (but independent from) #3049 .

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -398,10 +398,12 @@ function _copyto!(dest::AnyTracedRArray, bc::Broadcasted)
 
     args = (Reactant.broadcast_to_size(Base.materialize(a), size(bc)) for a in bc.args)
 
-    res = Reactant.promote_to(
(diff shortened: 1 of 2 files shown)


wsmoses (approved):

oschulz:
I think test failures are unrelated.

PR #3055, opened 2026-07-12
Carry closures called in traced loops as loop arguments

oschulz:
AI-generated (Claude Fable at max effort), human reviewed:

Traced values captured by closures that are called (but not otherwise referenced) inside `@trace` `while`/`for` bodies became block arguments of the while regions without matching while-loop operands, failing MLIR verification with `expect operands to be compatible with body block arguments`.

Treat single-name function calls in the loop condition and body as external symbols, so such closures become loop-carried arguments and the traced values they capture become proper while-loop operands. As suggested by @Pangoraw in https://github.com/EnzymeAD/Reactant.jl/pull/3030#discussion_r2532542427, names that resolve to globals of the calling module (checked via `isdefined(__module__, name)` at macro expansion time) are excluded, to avoid inflating the loop-carried arguments with references to ordinary functions. One consequence: a local callable that shadows a module-level global of the same name won't be detected, and module functions defined *after* the macro expansion site are still carried (harmless, just an extra argument).

Split out of #3030 as requested there.

### DIFF
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -258,10 +258,10 @@ macro trace(args...)
     Meta.isexpr(expr, :if) && return esc(trace_if(expr; track_numbers))
 
     Meta.isexpr(expr, :for) &&
-        return (esc(trace_for(expr; track_numbers, checkpointing, mincut)))
+        return (esc(trace_for(__module__, expr; track_numbers, checkpointing, mincut)))
 
     Meta.isexpr(expr, :while) &&
-        return (esc(trace_while(expr; track_numbers, checkpointing, mincut)))
+        return (esc(trace_while(__module__, expr; track_numbers, checkpointing, mincut)))
 
     return error(
         "Only `if-elseif-else` blocks, function definitions, `function calls`, `for` and \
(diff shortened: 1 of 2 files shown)


oschulz:
Should I remove the lengthy `# Locally defined callables (e.g. closures over traced values) that are ...` comment from the code?

oschulz:
Thanks @Pangoraw !

lib/ReactantCore/src/ReactantCore.jl
```diff
@@ -397,6 +397,20 @@ function trace_while(expr; track_numbers, mincut, checkpointing, first_arg=nothi
     union!(external_syms, cond_symbols.assignments)
     union!(external_syms, body_symbols.references)
     union!(external_syms, body_symbols.assignments)
+    # Locally defined callables (e.g. closures over traced values) that are
+    # called in the loop must become loop-carried arguments as well, so that
+    # traced values captured by them become while-loop operands instead of
+    # closure captures of the generated body function (which would result in
+    # region block arguments without matching operands). Only plain names that
+    # don't resolve to globals of the calling module can refer to local
+    # callables:
```
Pangoraw:
```suggestion
    # Add func calls to closures as references
```

Pangoraw:
```suggestion
    # Add func calls to closures as references
```

Pangoraw:
While not a full fix (if the closure has the same name as a module constant), i believe this is an improvement and not as convoluted as #3057.

oschulz:
Thanks @Pangoraw !

PR #3056, opened 2026-07-12
Self-host Triton dialect C-API wrappers inside ReactantExtra

ImanHosseini:
- Move Triton C-API wrapper declarations (`triton_dialect_capi.h/.cpp`) and TableGen wrapper (`triton.td`) directly into `deps/ReactantExtra/triton_capi/`.
- Add `:triton_dialect_capi_objects`, `:triton_dialect_capi_test`, and `:TritonDialectTdFiles` targets in `deps/ReactantExtra/BUILD`.
- Replace `@jax//jaxlib/triton` dependency in `ReactantExtraLib`, `julia_proto_bindings`, and `TritonJLIncGen` with self-hosted targets.
- Add unit test verifying pointer type creation and query functions.

### DIFF
--- a/deps/ReactantExtra/triton_capi/triton_dialect_capi.cpp
+++ b/deps/ReactantExtra/triton_capi/triton_dialect_capi.cpp
@@ -1,0 +1,56 @@
+#include "triton_dialect_capi.h"
+
+#include <optional>
+
+#include "llvm/Support/Casting.h"
+#include "mlir-c/IR.h"
+#include "mlir-c/Support.h"
+#include "mlir/CAPI/IR.h"
+#include "mlir/CAPI/Registration.h"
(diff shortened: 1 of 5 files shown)


deps/ReactantExtra/triton_capi/triton_dialect_capi.cpp
```diff
@@ -0,0 +1,56 @@
+#include "triton_dialect_capi.h"
+
```
wsmoses, line 2: @ImanHosseini do we want to move this to enzyme-jax [there are some other triton exts, though no c apis]

ImanHosseini:
Moving to enzyme-jax so we can reuse later.
https://github.com/EnzymeAD/Enzyme-JAX/pull/2613

PR #3058, opened 2026-07-13
Fix loop assignment tracing

### LINKED ISSUE #3053: Incorrect value returned from loop
The following code should return `[10.0,]` but returns `[0.0,]`

```julia
using Reactant

function bug(m::T) where {T}
    n = 0
    oneT = one(T)

    @trace while (n < 10)
        m += oneT
        n += 1
    end

    return T(n)
end

ms = Float64[0.01, ]
m_r = Reactant.to_rarray(ms)

Reactant.@jit bug.(m_r)
```

output

```
1-element ConcretePJRTArray{Float64,1}:
 0.0
```

The code returns the correct value if I remove the `m += oneT` in the loop

```julia
using Reactant

function bug(m::T) where {T}
    n = 0
    oneT = one(T)

    @trace while (n < 10)
        n += 1
    end

    return T(n)
end

ms = Float64[0.01, ]
m_r = Reactant.to_rarray(ms)

Reactant.@jit bug.(m_r)
```

```
1-element ConcretePJRTArray{Float64,1}:
 10.0
```

wsmoses:
fixes https://github.com/EnzymeAD/Reactant.jl/issues/3053

I don't totally understand/follow the fix, but llm made it go

@Pangoraw / @jumerckx can you read and make this reasonable in practice?

@dominic-chang

### DIFF
--- a/lib/ReactantCore/src/ReactantCore.jl
+++ b/lib/ReactantCore/src/ReactantCore.jl
@@ -441,6 +441,8 @@ function trace_while(expr; track_numbers, mincut, checkpointing, first_arg=nothi
     )
     varnames_expr = Expr(:tuple, QuoteNode.(external_syms)...)
 
+    args_res_sym = gensym("args_res")
+
     reactant_code_block = quote
         let $args_sym = $(args_init)
             $pre_alias_ids_sym = $(pre_ids_expr)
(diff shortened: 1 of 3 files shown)


Pangoraw (approved):

wsmoses:
there's a loooot of red here for ci...

PR #3059, opened 2026-07-13
Respect order of parameter specification

penelopeysm:
@wsmoses @sbrantq So I started playing around with this a bit and ran into a rather unexpected ordering of the output. I think it's quite unexpected to have parameters be sorted behind the scenes, most people would expect the output to have the parameters in the order that they were specified. This PR removes the sort.

I also started doing some docs tweaks to make it look more like a typical PPL tutorial (simulating data with the ground truth rather than hardcoding the data)

I'm not sure if you'd rather I open issues, or just do PRs directly. Let me know!

Regardless, I should say I'm trying to look at how to package some of the work here neatly up into something that has less boilerplate. More results on this to follow at some point in time.

cc also @Red-Portal

### DIFF
--- a/docs/src/tutorials/probprog/index.md
+++ b/docs/src/tutorials/probprog/index.md
@@ -45,13 +45,14 @@ fitted value `slope · x_i + intercept` with fixed noise (standard deviation 1).
 
 ### The data
 
-Synthetic data drawn from `slope = -2, intercept = 10`. The `xs` /
-`ys` pair is what we'll condition on.
+We start by drawing synthetic data using a known slope and intercept of `-2` and `10` respectively.
 
 ```@example probprog_index
-xs = [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0]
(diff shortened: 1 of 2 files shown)


sbrantq:
Hi @penelopeysm thanks for doing this! Direct PRs are totally fine.

We should remove the sort for sure. (It was an ad hoc "canonical order" to align with the convention of Enzyme-side trace construction that's no longer needed)

sbrantq (approved):

penelopeysm:
FYI: I started putting this together https://github.com/penelopeysm/Implus.jl. It's still very very rudimentary, but that can hopefully get better over time!

It felt a bit too presumptuous for me to take Impulse.jl, but I think it would make sense to do that if you want to have a "self-contained" high-level package for probabilistic programming. If you want, I'm happy to move the package to EnzymeAD org and rename the package.

penelopeysm:
(I reckon there's also an argument that the ProbProg module could be moved to a separate package, but that might be more hassle than it's worth)

PR #3063, opened 2026-07-14
Fix `Random.seed!(::ReactantRNG, ::Integer)` failure on Julia 1.10

penelopeysm:
The relevant method is defined here:

https://github.com/EnzymeAD/Reactant.jl/blob/e0cb37d73536ab46b7f891f0ff2179b1af21ba22/src/stdlibs/Random.jl#L24-L32

The issue is that `Random.hash_seed` was only introduced in Julia 1.11 (https://github.com/JuliaLang/julia/pull/51436) so this throws a UndefVarError on 1.10.

On top of that, `Random.hash_seed` is actually internal so we can't necessarily assume it will always exist / have the same behaviour.

This PR takes the most direct route to fix it, which is to just copy over the current definition of `Random.hash_seed`, which uses only public API from the `SHA` stdlib.

I couldn't test locally (for some reason a couple of Reactant extensions hang when precompiling with 1.10 on my Mac, maybe a separate issue) but I've abused some GitHub runners and the added tests pass.


src/stdlibs/Random.jl
```diff
@@ -21,13 +22,32 @@ end
     return Random.rand!(rng, Vector{UInt64}(undef, 2))
 end
 
+# This is the definition of Random.hash_seed from base Julia, but inlined because it's not
```
wsmoses: can we do something like

 if VERSION <= v"1.11"
  this def
else
   using Random: _hash_seed
end

penelopeysm: Yes we could -- I didn't opt for that as that is a bit more fragile towards newer versions removing /changing the behaviour, but if you prefer I could change it to an `isdefined` check

wsmoses: yeah I agree it's more fragile in an exported sense, but it is also more robust if the defn changes internally [and we want to maintain the same rng style algorithm/bias/etc]

penelopeysm: Yes agreed:)

wsmoses (approved):

PR #3065, opened 2026-07-14
Fix StructArrayStyle broadcast over separate arrays

oschulz:
Ran into this why trying to make PropertyFunctions.jl Reactant-compatible:

This fixes `StructArrays.StructArrayStyle` broadcasts where the input is not a `StructArray` but one or several plain arrays.

Limitation: struct-returning broadcasts over plain arrays, with no StructArrays involvement anywhere, (like a simple `broadcast((x,y) -> (s=x+y,), a, b)` without `StructArrayStyle`) will still fail as before. Fixing that would requite Reactant depending on StructArrays.

Claude Fable-5:

elem_apply_via_while_loop allocated its own result via similar(Broadcasted(f, args), T_res) even when its callers in ReactantStructArraysExt already hold a correctly-typed destination. For struct-returning functions broadcast over plain Reactant arrays under a StructArray broadcast style (as done e.g. by PropertyFunctions.jl to broadcast over individual columns), that internal allocation fails, since similar on the plain Reactant broadcast style only supports primitive and TracedRNumber element types.

Split out elem_apply_via_while_loop!(result, f, args...), which writes into a caller-provided destination, and use it in the StructArrays extension's copyto! methods. This also removes an allocate-then-copy round trip.

### DIFF
--- a/src/TracedUtils.jl
+++ b/src/TracedUtils.jl
@@ -1154,18 +1154,39 @@ scalar_arg(arg) = arg isa Base.RefValue || !(arg isa AbstractArray)
 flattenarg(arg) = ReactantCore.materialize_traced_array(vec(arg))
 flattenarg(arg::Ref) = RefFillVector(arg)
 
-function elem_apply_via_while_loop(f, args::Vararg{Any,Nargs}; kwargs...) where {Nargs}
+function elem_apply_via_while_loop!(
+    result, f, args::Vararg{Any,Nargs}; kwargs...
+) where {Nargs}
     non_ref_args = [arg for arg in args if !scalar_arg(arg)]
     if !isempty(non_ref_args)
         @assert allequal(size.(non_ref_args)) "All args must have the same size"
     end
-    out_size = isempty(non_ref_args) ? () : size(first(non_ref_args))
     L = isempty(non_ref_args) ? 1 : length(first(non_ref_args))
     # flattening the tensors makes the auto-batching pass work nicer
     flat_args = [flattenarg(arg) for arg in args]
 
+    ind_var = zeros(TracedRArray{Int}, ())
+    f_ref = Ref(f)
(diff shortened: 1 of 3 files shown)


oschulz:
Working on the test failures.

oschulz:
#3068 should hopefully fix the test failures here.

oschulz:
@wsmoses remaining test failures should be unrelated to this PR (I think).

oschulz:
@wsmoses good like this?

src/TracedUtils.jl
```diff
@@ -1154,18 +1154,41 @@ scalar_arg(arg) = arg isa Base.RefValue || !(arg isa AbstractArray)
...
+    return result
+end
+
+function elem_apply_via_while_loop(f, args::Vararg{Any,Nargs}; kwargs...) where {Nargs}
+    non_ref_args = filter(!scalar_arg, args)
+    out_size = isempty(non_ref_args) ? () : size(first(non_ref_args))
 
     # This wont be a mutating function so we can safely execute it once
-    scalar_seed_args = [@allowscalar(arg[1]) for arg in flat_args]
+    scalar_seed_args = map(arg -> @allowscalar(flattenarg(arg)[1]), args)
```
wsmoses: avoid the use of closures

oschulz: Replaced the anonymous function with a named top-level helper (`scalar_seed_arg`) in b39d50f. Left `filter(!scalar_arg, ...)` as is — `!f` functors are already used in the tree (`filter(!isnothing, ...)` in TracedRArray.jl); happy to switch it to a named predicate as well if preferred.

oschulz:
@wsmoses anything you'd still like me to change here?

oschulz:
Gentle ping @wsmoses .

PR #3067, opened 2026-07-14
Support broadcasting tuple/struct-returning functions over traced arrays

### LINKED ISSUE #1566: Reactant compile fails with broadcasts of functions that return tuples
Reactant current doesn't seem to be able to deal with broadcasts of functions that return tuples. Given

```julia
using Reactant
Reactant.set_default_backend("cpu")
A = Reactant.to_rarray(rand(Float32, 10))
foo(a) = (a, a)
foo.(Array(A)) isa AbstractVector{<:Tuple}
```

running

```julia
@code_hlo foo.(A)
```

(or `@compile`) results in

```
ERROR: MethodError: no method matching similar(::Base.Broadcast.Broadcasted{…}, ::Type{…}, ::Tuple{…})
The function `similar` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  similar(::Type{Reactant.XLA.PJRT.AsyncBuffer}, ::Any...; kwargs...)
   @ Reactant /user/.julia/packages/Reactant/hB4Fs/src/xla/PJRT/AsyncBuffer.jl:19
  similar(::Reactant.XLA.PJRT.AsyncBuffer, ::Any...)
   @ Reactant /user/.julia/packages/Reactant/hB4Fs/src/xla/PJRT/AsyncBuffer.jl:15
  similar(::AbstractArray, ::Type{T}, ::Tuple{Union{Integer, Base.OneTo}, Vararg{Union{Integer, Base.OneTo}}}) where T
   @ Base abstractarray.jl:830
  ...

Stacktrace:
  [1] macro expansion
    @ /user/.julia/packages/Reactant/hB4Fs/src/utils.jl:0 [inlined]
  [2] call_with_reactant(::Reactant.MustThrowError, ::typeof(similar), ::Base.Broadcast.Broadcasted{…}, ::Type{…}, ::Tuple{…})
    @ Reactant /user/.julia/packages/Reactant/hB4Fs/src/utils.jl:875
  [3] similar
    @ ./broadcast.jl:223 [inlined]
  [4] (::Nothing)(none::typeof(similar), none::Base.Broadcast.Broadcasted{…}, none::Type{…})
    @ Reactant ./<missing>:0
```

(This is an MWE, distilled down from the same problem in a more complex use case.)

oschulz:
Closes #1566.

Somewhat related, but technically orthogonal to #3065.

Claude Fable-5 max:

Broadcasting a function that returns a tuple or struct over a traced array
(e.g. `foo.(A)` with `foo(a) = (a, a)`) failed with a `MethodError` on
`similar`, because the broadcast eltype is a `Tuple`/struct and no `similar`
method existed for such eltypes under `AbstractReactantArrayStyle`.

The downstream machinery already handled this case: `elem_apply` returns a
struct-of-arrays result and the StructArrays extension already has
`alloc_sarr` for building struct-of-arrays containers plus the matching
`copyto!` fill path. Only the container-allocation hook was missing.

This adds:

- `similar(::Broadcasted{AbstractReactantArrayStyle{N}}, ::Type{T}, dims)` in
  the StructArrays extension, which builds a `StructArray` via `alloc_sarr`
  for tuple/struct eltypes (and otherwise rethrows the original `MethodError`
  so unrelated cases are unaffected). `alloc_sarr` is generalized to thread
  `dims` through instead of always using `axes(bc)`.
- A `MethodError` hint in `Reactant.__init__` telling users to load
  StructArrays.jl when this broadcast is attempted without it.
- Integration tests covering tuple and named-tuple results.

The result is a `StructVector` (still `<: AbstractVector{<:Tuple}`, as the
issue expects), and the broadcast optimizes to clean batched HLO — e.g.
`(sin.(a), cos.(a))` lowers to `stablehlo.sine` + `stablehlo.cosine` over
whole tensors.

### DIFF
--- a/ext/ReactantStructArraysExt.jl
+++ b/ext/ReactantStructArraysExt.jl
@@ -31,6 +31,8 @@ function __init__()
     Reactant.@skip_rewrite_func StructArrays.index_type
 end
 
+Reactant.is_extension_loaded(::Val{:StructArrays}) = true
+
 StructArrays.always_struct_broadcast(::AbstractReactantArrayStyle) = true
 
 function Base.copy(
@@ -81,14 +83,14 @@ function Base.copyto!(
     return copyto!(dest, res)
 end
 
-function alloc_sarr(bc, T)
+function alloc_sarr(bc, T, dims=axes(bc))
     # Short circuit for Complex since in Reactant they are just a regular number
-    T <: Complex && return similar(bc, T)
-    asa = Base.Fix1(alloc_sarr, bc)
+    T <: Complex && return similar(bc, T, dims)
+    asa = FT -> alloc_sarr(bc, FT, dims)
     if StructArrays.isnonemptystructtype(T)
         return StructArrays.buildfromschema(asa, T)
     else
-        return similar(bc, T)
+        return similar(bc, T, dims)
     end
 end
 
@@ -98,6 +100,16 @@ function Base.similar(
     bc′ = convert(Broadcasted{S}, bc)
     # It is possible that we have multiple broadcasted arguments
     return alloc_sarr(bc′, ElType)
+end
+
+# Broadcasted functions returning tuples or structs over traced arrays get a
+# struct-of-arrays result (see issue #1566)
+function Base.similar(
+    bc::Broadcasted{AbstractReactantArrayStyle{N}}, ::Type{T}, dims
+) where {T,N}
+    (T <: Complex || !StructArrays.isnonemptystructtype(T)) &&
+        throw(MethodError(similar, (bc, T, dims)))
+    return alloc_sarr(bc, T, dims)
 end
 
 Base.@propagate_inbounds function StructArrays._getindex(
--- a/ext/ReactantStructArraysExt.jl
+++ b/ext/ReactantStructArraysExt.jl
@@ -83,12 +83,17 @@ function Base.copyto!(
     return copyto!(dest, res)
 end
 
+struct AllocSArr{B,D} <: Function
+    bc::B
+    dims::D
+end
+(asa::AllocSArr)(T) = alloc_sarr(asa.bc, T, asa.dims)
+
 function alloc_sarr(bc, T, dims=axes(bc))
     # Short circuit for Complex since in Reactant they are just a regular number
     T <: Complex && return similar(bc, T, dims)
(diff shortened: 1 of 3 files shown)


ptiede:
I'm concerned about changing the array type under the hood on users here. This means code executed under Reactant may take a separate dispatch/code path when compiled which could make debugging rather annoying and potentially introduce strange bugs. This is at least part of the reason I never implemented this when updating the StructArrays extension.

oschulz:
> I'm concerned about changing the array type under the hood on users here. [...]

Yes, that is less than perfect. On the other hand Reactant "redirects" quite a few things already, so running under Reactant often needs to very verified/tested in addition to Julia default execution. But yes ... I just don't know how else to solve this.

I think a solution that is both type-faithful and efficient may be fundamentally impossible - the question is, which is the lesser evil, changing type or not supporting these kinds of broadcasts? I tend to say changing type here is defensible, but I do understand the worry.

And

```
function Base.similar(
    bc::Broadcasted{AbstractReactantArrayStyle{N}}, ::Type{T}, dims
)
```

is kinda-sorta "extension type piracy" in a way.

oschulz:
@wsmoses , @ptiede could we make StructArrays a hard dependency of Reactant? In the end, Reactant needs some SoA representation with an AoS interface, I'd say, and that's what StructArrays provides.  I think it could open many doors (i.e. open a lot of non-ML code for Reactant).

A rule that the broadcast style of Reactant would be a `StructArray`-style would be principled and predictable for the user, right? But for that we'd really need to depend on StructArrays.

penelopeysm:
> I think it could open many doors (i.e. open a lot of non-ML code for Reactant).

FWIW, I ran into the same issue when trying to adapt the equivalent of `Distributions.product_distribution` to work on Reactant (#3192), so this is definitely true. It's not a _fundamental_ blocker because I can rewrite _my_ code to use SoA instead of AoS, but I can't rewrite Distributions 😉.

I haven't thought about it too hard, but I agree that StructArrays is conceptually the right tool to handle it. I had Claude do up a 'simpler' fix that just overloaded `similar` plus adjusted `_copyto!` to use scalar iteration instead of `elem_apply`, and that did enough in the MWE in my issue, but it doesn't vectorise well at all.

PR #3068, opened 2026-07-14
Make convert to non-concrete TracedRArray types a no-op when possible

oschulz:
`Base.convert(T::Type{<:TracedRArray}, x::AbstractArray)` unconditionally
forwards to `promote_to`, which has methods for `TracedRArray`,
`TracedRArray{T}`, and `TracedRArray{T,N}` — but not for non-concrete
targets like `TracedRArray{<:Any,N}`. Julia produces exactly such targets
via `promote_typejoin` when `collect` widens over mixed-eltype traced
arrays (on 1.10/1.11, where inference-based `collect` falls back to
`grow_to!`/`push_widen`). The resulting `convert` call then throws a
`MethodError`, so an innocent comprehension like

```julia
v = [x for x in (a_f64, b_f32)]  # both TracedRArrays
```

fails during tracing. Without Reactant's `convert` override, Base's
generic `convert(::Type{T}, x::T) = x` would have handled this case; this
PR restores that behavior by returning `x` directly when `x isa T`.
Conversions between concrete traced types are unaffected (for those,
`promote_to` was already the identity whenever `x isa T` holds).

Relationship to other PRs:

- This is the root-cause follow-up to the CI failures on #3065, whose new
  mixed-eltype broadcast test was the first to collect heterogeneous
  traced arrays. #3065 now sidesteps the issue locally (tuples instead of
  `Vector`s in `elem_apply_via_while_loop[!]`) — a good change on its own —
  while this PR defuses the whole class for user code. No textual overlap;
  both apply independently.
- Independent of #3067 (tuple-returning broadcasts).

Tests: direct `convert` identity checks (including the previously-throwing
non-concrete target, which fails on all Julia versions without this fix)
plus a traced mixed-eltype `collect` regression test that exercises the
1.10/1.11 widening path on CI.

Created by generative AI.

### DIFF
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -27,7 +27,12 @@ Base.elsize(::Type{<:TracedRArray{T,N}}) where {T,N} = sizeof(T)
 
 # This is required otherwise we will copy a tracedrarray each time
 # we use it
-Base.convert(T::Type{<:TracedRArray}, x::AbstractArray) = Reactant.promote_to(T, x)
+function Base.convert(T::Type{<:TracedRArray}, x::AbstractArray)
+    # non-concrete `T` (e.g. from `promote_typejoin` when collecting mixed-eltype
+    # traced arrays) has no `promote_to` methods; `x isa T` needs no conversion
+    x isa T && return x
+    return Reactant.promote_to(T, x)
+end
 
 # Base.complex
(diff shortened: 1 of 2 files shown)


wsmoses (approved):

oschulz:
Anything I should still change here?

oschulz:
Gentle ping @wsmoses .

PR #3076, opened 2026-07-19
Improve docs for local builds

penelopeysm:
Fable and I spent a while figuring out the easiest way to make it build on my Mac. While I'm sure some of these fixes could be hardcoded in the build process, I think it's more flexible to just document them since I don't know whether other people would actually run into the same problem in practice. Text is written by me.


deps/ReactantExtra/WORKSPACE
```diff
@@ -4,7 +4,7 @@ NSYNC_COMMIT = "82b118aa7ace3132e517e2c467f8732978cf4023"
 
 NSYNC_SHA256 = ""
 
-ENZYMEXLA_COMMIT = "94c1d963b66df169478591a5539281005dd85f12"
+ENZYMEXLA_COMMIT = "d87746dd2e242fe7477cc06ad38f999b83d65ec5"
```
penelopeysm: cf. https://github.com/EnzymeAD/Enzyme-JAX/pull/2636 this fixes another of the build errors I had

docs/src/tutorials/local-build.md
```diff
@@ -40,11 +49,11 @@ optional arguments:
   --gcc_host_compiler_path GCC_HOST_COMPILER_PATH
                         Path to the gcc host compiler. (default:
                         "/usr/bin/gcc")
-  --cc CC                (default: "/usr/bin/cc")
+  --cc CC                (default: "/opt/homebrew/opt/llvm/bin/clang")
```
penelopeysm: Hmm, I updated the output of `--help`, didn't realise that this would pick up what was on my system. Can revert if desired but I figure it doesn't really matter since it's just illustration

docs/src/tutorials/local-build.md
```diff
@@ -7,11 +7,14 @@ In the `deps/` subdirectory of the Reactant repository there is a script to do l
 * Julia.  If you don't have it already, you can obtain it from the [official Julia website](https://julialang.org/downloads/)
 * A reasonably recent C/C++ compiler, ideally GCC 12+.
   Older compilers may not work.
-* Bazel. If you don't have it already, you can download a build for your platform from [the latest `bazelbuild/bazelisk` release](https://github.com/bazelbuild/bazelisk/releases/latest) and put the `bazel` executable in `PATH`
-* not necessary in general, but for debug builds with CUDA support, you'll need a fast linker, like `lld` or `mold`.
+* Bazel. If you don't have it already, you can download a build for your platform from [the latest `bazelbuild/bazelisk` release](https://github.com/bazelbuild/bazelisk/releases/latest) and put the `bazel` executable in `PATH`.
+* A fast linker like `lld` or `mold`.
```
wsmoses, line 11: this bullet is reworded a bit weirdly, I may revert

docs/src/tutorials/local-build.md
wsmoses: yeah I may recommend reverting the help defaults [saying cc or <MAX NUMBER OF CPUS> explicitly feels more helpful]

docs/src/tutorials/local-build.md
```diff
@@ -102,6 +117,98 @@ Finally, you may have to add the flag `--extraopt "--sandbox_writable_path=/path
...
+If you are building on macOS and you see an error such as
+
+```console
+external/riegeli/riegeli/base/new_aligned.h:137:27: error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on macOS 10.13 or newer
+```
+
+(note that this error can be triggered from multiple different source files), this can be fixed by adding the options
+
+```console
+julia [...] --copt "-mmacosx-version-min=11.0" --extraopt "--host_copt=-mmacosx-version-min=11.0"
```
wsmoses: I feel like we should just have these args added on macos automatically

docs/src/tutorials/local-build.md
```diff
@@ -102,6 +117,98 @@ Finally, you may have to add the flag `--extraopt "--sandbox_writable_path=/path
...
+Depending on your compiler toolchain and Bazel version you may encounter an error like:
+
+```console
+external/com_google_absl/absl/status/statusor.h:58:10: error: module com_google_absl//absl/status:statusor does not depend on a module exporting 'absl/types/span.h'
+```
+
+This is caused by Bazel performing a strict version of dependency checking between C++ modules, which can be disabled with
+
+```console
+julia [...] --extraopt "--features=-layering_check" --extraopt "--host_features=-layering_check"
```
wsmoses, line 145: the layering check should automatically added as an arg already?

docs/src/tutorials/local-build.md
penelopeysm: The build script already adds `-fuse-ld=lld`

https://github.com/EnzymeAD/Reactant.jl/blob/ef202e8582682f5edcf4e83b6a3dbfe049629828/deps/build_local.jl#L276

even for non-CUDA builds, so I didn't entirely like that the current version implies that it's only needed for CUDA.

To avoid the awkward phrasing, how about just simplifying to this?

```
* A fast linker like [`lld`](https://lld.llvm.org/) or [`mold`](https://github.com/rui314/mold) (the build script uses `lld` by default).
```

Mold's github release tarballs already include an `ld.mold` symlink and i think it should be obvious to anyone building stuff that executables need to be on path so the rest can be skipped imo

docs/src/tutorials/local-build.md
penelopeysm: I'm not opposed to that at all

penelopeysm: I added them to the build script, can confirm it works locally

docs/src/tutorials/local-build.md
penelopeysm: Nothing in the build config enables it, but for some reason it's enabled by default on my system. I think it depends on compiler and/or Bazel versions. And I have to disable it to get it to build

deps/build_local.jl
```diff
@@ -309,7 +316,7 @@ push!(build_cmd_list, "--copt=-Wno-private-header")
 push!(build_cmd_list, "--color=$(parsed_args["color"])")
 push!(build_cmd_list, ":libReactantExtra.so")
 
-@info "About to run Bazel" build_cmd_list
+@info "About to run Bazel\n" * join(build_cmd_list, " \\\n  ")
```
penelopeysm, line 319: long command lists get truncated in `@info` so this spells it out

docs/src/tutorials/local-build.md
```diff
@@ -21,17 +24,23 @@ julia --project -e 'using Pkg; Pkg.instantiate()' # needed only the first time t
...
+There are a few options you may want to use to tweak the build.
+For more information run the command
+
+```bash
+julia --project build_local.jl --help
+```
+
+which prints the following (note that the output may be out of date; you should run it locally to see the options available to you):
 
 ```console
```
giordano: If you remove the prompt line, you can also remove the console type here
````suggestion
```
````

File: docs/src/tutorials/local-build.md [hunk 8]
```diff
@@ -102,6 +117,82 @@ Finally, you may have to add the flag `--extraopt "--sandbox_writable_path=/path
 At the end of a successful build, the `build_local.jl` script will create a `LocalPreferences.toml` file (see [`Preferences.jl` documentation](https://juliapackaging.github.io/Preferences.jl/stable/) for more information) in the top-level of the Reactant repository, pointing `libReactantExtra` to the new local build. 
 If you instantiate this environment, Reactant will automatically use the new local build.
 
+You can check that the local build is being used by running
+
+```julia
+using Reactant; Reactant.Reactant_jll.libReactantExtra_path
```
giordano: I'd more simply do
```suggestion
using Reactant_jll; Reactant_jll.libReactantExtra_path
```
`Reactant_jll` must be a top-level dependency of the environment anyway for the preference to be effective.

giordano (approved):

docs/src/tutorials/local-build.md
penelopeysm: Oh there actually wasn't a prompt so I just went ahead and removed it!

docs/src/tutorials/local-build.md
wsmoses: okay I'm fine with the bullet here as is

deps/ReactantExtra/WORKSPACE
wsmoses: revert this?

wsmoses (approved):
okay this all lgtm modulo some final small comments

deps/build_local.jl
```diff
@@ -275,6 +275,13 @@ push!(build_cmd_list, "--spawn_strategy=local")
 
 push!(build_cmd_list, "--linkopt=-fuse-ld=lld")
 
+# Many deps require C++17 aligned allocation which is not available on
+# old macOS libc++
+if Sys.isapple()
+    push!(build_cmd_list, "--copt=-mmacosx-version-min=11.0")
+    push!(build_cmd_list, "--host_copt=-mmacosx-version-min=11.0")
+end
+
```
mofeing, line 282: Are these flags needed? We already build for x86-64 macOS and it works.

These flags are fine for me but would like to know if you found some error that forced you to add them.

docs/src/tutorials/local-build.md
```diff
@@ -7,11 +7,14 @@ In the `deps/` subdirectory of the Reactant repository there is a script to do l
...
   Older compilers may not work.
-* Bazel. If you don't have it already, you can download a build for your platform from [the latest `bazelbuild/bazelisk` release](https://github.com/bazelbuild/bazelisk/releases/latest) and put the `bazel` executable in `PATH`
-* not necessary in general, but for debug builds with CUDA support, you'll need a fast linker, like `lld` or `mold`.
+* Bazel. If you don't have it already, you can download a build for your platform from [the latest `bazelbuild/bazelisk` release](https://github.com/bazelbuild/bazelisk/releases/latest) and put the `bazel` executable in `PATH`.
+* A fast linker like `lld` or `mold`.
+  This is strictly speaking not necessary in general, but `lld` is used by default, and for debug builds with CUDA support, you'll need it.
   Binutils `ld` won't work, don't even try using it.
   You can obtain `mold` for your platform from the [latest `rui314/mold` release](https://github.com/rui314/mold/releases/latest) and put the executable in `PATH` with name `ld.mold`.
 
+On MacOS the latter two requirements can be installed with `brew install bazelisk lld`.
```
mofeing: micro fix (can ignore it)

```suggestion
On macOS the latter two requirements can be installed with `brew install bazelisk lld`.
```

penelopeysm: haha no i appreciate this sort of thing

deps/build_local.jl
penelopeysm: Yes indeed - an older commit had the exact error, something like

```
external/riegeli/riegeli/base/new_aligned.h:137:27: error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on macOS 10.13 or newer
```

Also curiously --...=10.13 didn't work for me, I had to use 11.0. But i didn't investigate that further.

deps/ReactantExtra/WORKSPACE
penelopeysm: yup i can fix this when merging main in

penelopeysm:
Sorry I dropped this for a while! Should be good to go now

(formatting failure is in unrelated files)

deps/build_local.jl
mofeing: it actually seems like we target macOS 12.3 on ReactantBuilder https://github.com/EnzymeAD/ReactantBuilder/blob/760b466b19846b78d2797f747e61a1b2240a1827/R/Reactant/build_tarballs.jl#L657-L661 so it's fine 👍

mofeing (approved):

PR #3096, opened 2026-07-24
Unify julia cache key

wsmoses:
we shouldn't need different caches for different test args / events, right?

reason being that we evict the gh cache too much

### DIFF
--- a/.github/workflows/CommonCI.yml
+++ b/.github/workflows/CommonCI.yml
@@ -91,7 +91,7 @@ jobs:
       - uses: julia-actions/cache@v3
         id: julia-cache
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/CommonCI.yml
```diff
@@ -91,7 +91,7 @@ jobs:
       - uses: julia-actions/cache@v3
         id: julia-cache
         with:
-          cache-name: julia-cache;workflow=${{ inputs.julia_version }}-${{ inputs.os }}-${{ inputs.runtime }}-assertions=${{ inputs.assertions }}-${{ github.event_name }}-${{ inputs.test_args }}-${{ inputs.downgrade_testing }}-${{ inputs.localjll || inputs.enzymejax_commit != '' }}-${{ inputs.enzymejax_commit }}
+          cache-name: julia-cache;workflow=${{ inputs.julia_version }}-${{ inputs.os }}-${{ inputs.runtime }}-assertions=${{ inputs.assertions }}-${{ inputs.downgrade_testing }}-${{ inputs.localjll || inputs.enzymejax_commit != '' }}-${{ inputs.enzymejax_commit }}
```
giordano, line 94: I don't know how it works in GCP, but on GitHub actions you'd have race conditions if you don't fully disambiguate the key indices, ending up with broken caches (for exmample one job may be fetching it while another one is writing it)

PR #3097, opened 2026-07-24
Add more specific convert methods for CuTracedRNumber{Float}

Pangoraw:


ext/ReactantCUDAExt.jl
```diff
@@ -192,7 +192,8 @@ end
     )
 end
 
-Base.convert(::Type{<:CuTracedRNumber{T}}, x::CuTracedRNumber{T}) where {T} = x
+Base.convert(::Type{<:CuTracedRNumber{Float32,1}}, x::CuTracedRNumber{Float32,1}) = x
```
wsmoses, line 197: @Pangoraw why was this needed??

Pangoraw: i think both of these methods are competing:

```julia
@inline function Base.convert(CT::Type{<:CuTracedRNumber{Float32,1}}, x::Number)
Base.convert(CT::Type{T}, x::T) where {T <: Number}

wsmoses: fair enough

wsmoses: we also need bfloat16 though, so keep the default method and also add the bf16 one

wsmoses (approved):

PR #3121, opened 2026-07-27
Ignore sharding metadata when comparing user and XLA shardings

wsmoses:
The [Enzyme-JAX jax bump](https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/30293798235/job/90069681247) fails a bunch of sharding tests with e.g.

```
AssertionError: Sharding provided by the user ({devices=[1,8]<=[8]}) does not match the sharding
computed by XLA ({devices=[1,8]<=[8] metadata={op_type="sdy::reduction_op" op_name="MAX"}}).
```

The only difference is the metadata: newer XLA parses a reduction op out of the `sdy::reduction_op` metadata entry (`HloSharding::ExtractReductionOpFromMetadata`) and compares `reduction_op_` in `xla::HloSharding::operator==`, so shardings that lay the data out identically now compare unequal.

Since that assertion only cares about how the data is laid out across devices, this adds a separate `hlo_sharding_check_eq_ignoring_metadata` C API (leaving `hlo_sharding_check_eq`/`==` untouched) which compares the two shardings after round-tripping them through the proto with the metadata cleared — that also clears the metadata-derived reduction op, which `HloSharding::WithoutMetadata` does not.

Note: the same CI job also hits `StackOverflowError`s inside XLA compilation for a few tests; those are a separate issue and are not addressed here.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -2648,9 +2648,36 @@ hlo_sharding_tile_assignment_devices(xla::HloSharding *hloSharding,
   }
 }
 
+static void clear_op_sharding_metadata(xla::OpSharding &op_sharding) {
+  op_sharding.clear_metadata();
+  for (auto &tuple_sharding : *op_sharding.mutable_tuple_shardings())
+    clear_op_sharding_metadata(tuple_sharding);
+}
+
+// Drops the metadata of `hlo_sharding`. Round-tripping through the proto (as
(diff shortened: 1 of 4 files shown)


wsmoses:
Follow-up: the metadata itself turned out to be junk we produce ourselves — our `sdyTensorShardingAttrGet` binding is missing shardy's trailing `uint32_t reductionOp` parameter, so every `sdy.sharding` gets an uninitialized reduction op. Fixed in #3122, which should also remove the `StackOverflowError`s. This PR is then just hardening of the assertion.

PR #3131, opened 2026-07-29
enable sroa on `:after_enzyme` pipeline

mofeing:


src/compiler/Compiler.jl
```diff
@@ -429,50 +431,38 @@ function compile_mlir!(
         run_pass_pipeline!(mod, "raise-triton-custom-call", "raise_triton_custom_call")
     end
 
-    if compile_options.optimization_passes === :all
+    if compile_options.optimization_passes === :all || compile_options.optimization_passes === :after_enzyme
+
+
+    elseif compile_options.optimization_passes === :after_enzyme
```
mofeing: I think this conditional is not ok...

wsmoses: fixed that, bad merge

PR #3136, opened 2026-07-31
Adds foundations for GPU MPI support via NCCL

romanlee:
Add the foundations for GPU MPI support (via NCCL) to ReactantMPIExt. Mainly, this amounts to adding logic and state to properly set up NCCL. This logic is concentrated in `ext/ReactantMPIExt/Nccl.jl`, and specifically in the function `ReactantMPIExt.initialize!(MPI.COMM_WORLD)` which:
* explicitly initializes an XLA gpu client with exactly one allowed device per MPI rank and then sets the NCCL device accordingly.
* sets up the NCCL communicator. The address of the communicator ultimately gets passed to EnzymeJAX as a pass option on LowerEnzymeXLAPass.

We also add backend gpu tests for MPI Allreduce, which will be supported when https://github.com/EnzymeAD/Enzyme-JAX/pull/2699 lands

NOTE: Ultimately we might want to move NCCL communicator creation to EnzymeJAX, but this is what we agreed on for a first pass.


ext/ReactantNCCLExt/ReactantNCCLExt.jl
avik-pal, line 1: XLA:CUDA already ships with NCCL, why do we need to load these via an extension?

romanlee: Hm yeah I suppose we could do it that way. Would that be better?

romanlee: Yeah actually on reflection that is definitely better. I will refactor this.

romanlee: @avik-pal let me know how things look to you now, see https://github.com/EnzymeAD/Reactant.jl/pull/3136/commits/4f9dab6d4f3fbf998b419b93f348082a04f45b3a.

romanlee: Note also that in the future, we'll probably move nccl communicator creation to enzymejax, but setting it up somewhat clunkily in reactant was what @wsmoses and I decided on for a first pass.

ext/ReactantMPIExt/Nccl.jl
```diff
@@ -0,0 +1,132 @@
+const ncclComm_t = Ptr{Cvoid}
+
+struct ncclUniqueId
```
mofeing: Style check: type names should be in CamelCase.

```suggestion
const NcclComm_t = Ptr{Cvoid}

struct NcclUniqueId
```

test/runtests.jl
```diff
@@ -135,9 +134,14 @@ test_worker = custom_test_worker ? tpu_custom_worker_launcher : Returns(nothing)
         @testset "MPI" begin
             using MPI
             nranks = 2
-            run(
-                `$(mpiexec()) -n $nranks $(Base.julia_cmd()) --project=$(Base.active_project()) $(joinpath(@__DIR__, "integration", "mpi.jl"))`,
-            )
+            withenv(
+                "XLA_REACTANT_GPU_MEM_FRACTION" => 1 / (nranks + 0.1),
+                "XLA_REACTANT_GPU_PREALLOCATE" => false,
```
mofeing: just curious, is preallocation giving problems? also, if you don't preallocate, then does `XLA_REACTANT_GPU_MEM_FRACTION` does anything?

test/runtests.jl
```diff
@@ -125,8 +125,7 @@ test_worker = custom_test_worker ? tpu_custom_worker_launcher : Returns(nothing)
     end
 
     if (
-        # MPI is only supported on CPU
-        (BACKEND == "cpu" || BACKEND == "auto") && (
+        BACKEND in ("auto", "cpu", "cuda", "gpu") && (
```
mofeing, line 128: mmm `gpu` is CUDA but also ROCm, which can right now gives problems because we don't yet support RCCL (which will be trivial to support once we merge this or other features)

src/Reactant.jl
```diff
@@ -272,6 +272,18 @@ const TracedType = Union{TracedRArray,TracedRNumber,MissingTracedValue}
...
+    ext === nothing && error("ReactantMPIExt is not loaded; load MPI first")
+    return ext.default_comm_handle()
+end
+
+function set_nccl_device!(::Integer)
+    return error(
+        "GPU MPI requires CUDA.jl; load CUDA before initializing the NCCL communicator"
+    )
+end
+
```
mofeing: So... you shouldn't call `Base.get_extension` inside regular lib code. Instead, you should declare it but left undefined, and define it in the appropriate extension.

But since whole ReactantMPIExt/Nccl.jl file doesn't depend on anything inside the `ReactantMPIExt` module, maybe better move to its own file inside the `src/` folder?

test/integration/mpi.jl
```diff
@@ -31,267 +48,347 @@ datatypes = [
...
-    # Operations that only work with integer/boolean types
-    integer_bool_ops = Set([MPI.LAND, MPI.LOR, MPI.LXOR, MPI.BAND, MPI.BOR, MPI.BXOR])
-
-    for (opname, op) in operations
-        for T in datatypes
-            # Skip some invalid combinations of T and op
-            if op in integer_bool_ops && !(T <: Integer || T <: Bool)
-                continue
-            end
+try
```
mofeing: maybe moving the `try`/`finally` to the runtests.jl file would make it more legible? wdyt?

ext/ReactantMPIExt/Nccl.jl
```diff
@@ -0,0 +1,132 @@
...
+    internal::NTuple{128,UInt8}
+end
+
+const DEFAULT_COMM = Ref{ncclComm_t}(C_NULL)
+const DEFAULT_COMM_HANDLE = Ref{UInt}(0)
+const DEFAULT_XLA_DEVICE = Ref{Union{Nothing,Reactant.XLA.AbstractDevice}}(nothing)
```
mofeing: mind prefixing sth like `REACTANT_NCCL_` here? the reasoning being that there is overlap with MPI and both are on the same module

ext/ReactantMPIExt/Nccl.jl
```diff
@@ -0,0 +1,132 @@
...
+    return override
+end
+
+function Reactant.XLA.default_device(client::Reactant.XLA.PJRT.Client)
+    return require_default_xla_device(client)
+end
+
+function Reactant.XLA.default_device(client::Reactant.XLA.IFRT.Client)
+    return require_default_xla_device(client)
+end
```
mofeing: is this piracy? it certainly doesn't feel right. why do you need to override these methods from XLA?

ext/ReactantMPIExt/Nccl.jl
mofeing: if you move to its own module, that would be another solution.

mofeing (changes requested):
@avik-pal we agreed to get this done like @romanlee to get some MPI runs working soon. I agree we should better use the XLA-built NCCL (and since we do that, we can maybe write our own custom calls directly in Enzyme-JAX), but we can fix it and improve it incrementally. That's what also the new `comm` dialect is about.

@romanlee I feel like this should better live in its own NCCL extension. You're using a lot of MPI to set it up, but the user is gonna know if it's going to execute on a GPU cluster with NCCL. Ultimately, the NCCL backend for MPI should be used for converting user-code, but the set up should be done with NCCL and not MPI. It's true that we are gonna use XLA-built NCCL, which can lead to some conflicts, and would like to know your opinion.

EDIT: I see that you had your code in a NCCL extension first, and I think we should discuss it.

romanlee:
> I feel like this should better live in its own NCCL extension. You're using a lot of MPI to set it up, but the user is gonna know if it's going to execute on a GPU cluster with NCCL. Ultimately, the NCCL backend for MPI should be used for converting user-code, but the set up should be done with NCCL and not MPI. It's true that we are gonna use XLA-built NCCL, which can lead to some conflicts, and would like to know your opinion.

@mofeing I actually think it's better to not have a NCCL extension for a few reasons: (1) we avoid incompatibility issues between nccl.jl's and xla's packaged nccl libs, (2) since what we're doing with nccl is relatively straightfoward (just setting up a communicator) it's relatively easy to just create the wrappers ourselves and that allows us to avoid another dependency, and (3) ultimately we'll move communicator creation out of Reactant anyway, at which point we won't need the extension, so it's fine to do something lightweight for now. Curious what @avik-pal's opinion is, but I assume based on his original comment that he was also thinking better to avoid an extension unless necessary.

Re using a lot of MPI to set it up, that's just how you set up NCCL. See https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/communicators.html#creating-a-communicator.

test/runtests.jl
romanlee: Glad you pointed this out. This shouldn't be necessary so I'll remove it, was a hydra thing. 

Incidentally, I assume MEM_FRACTION sets the budget regardless of whether it's preallocated or allocated on demand? That's what other usage would suggest, eg just above in the same file

test/runtests.jl
romanlee: Thanks, see https://github.com/EnzymeAD/Reactant.jl/pull/3136/commits/82c9567d2d3041175701d7f89e2dd54e56c5fa34

src/Reactant.jl
romanlee: Ah thanks, this didn't feel quite right to me either. See https://github.com/EnzymeAD/Reactant.jl/pull/3136/commits/54a04fc76208b01945f64492895358647a8721a3. Re `set_nccl_device!`, I think it's appropriate as is. Let me know if you disagree.

And re moving `Nccl.jl` to `src/`, I think it makes sense to leave it in the `ReactantMPIExt` module even if it doesn't strictly need to be there, since it is always used together with the mpi extension. But if that's not the right style/you feel strongly I'm happy to move it

test/integration/mpi.jl
romanlee: I don't think that's really possible bc we can't access the stuff in the `finally` block in runtests.jl? Anyway, that feels like it would be slightly less readable to me, so I'd prefer to keep it where it is. 

That said, I did make some minor changes to make it more readable here https://github.com/EnzymeAD/Reactant.jl/pull/3136/commits/a0ebd0439eb05fffecdabd61926b0af9d3fa76eb

test/runtests.jl
romanlee: Wait, thinking a bit more about this, are you sure this is right? Doesn't "gpu" *sometimes* mean "cuda"? How do we make sure to catch that case properly? 

Also, as written I have "auto" routing to `set_default_backend("cpu")`, which probably isn't right either. Because "auto" defaults to "cuda" or "rocm" with higher priority than "cpu", which we probably want to respect.

If you could clarify the intended usage of "REACTANT_BACKEND_GROUP" here that would be very helpful.

romanlee:
After thinking more about it, I decided to refactor the way the NCCL is initialized. See https://github.com/EnzymeAD/Reactant.jl/pull/3136/commits/15662fc0ddfdf551e05185a49de1a957ac2100a4.

Specifically, previously we selected the NCCL device after XLA client creation. This was fragile because it left all devices addressable. This refactor adds a function `ReactantMPIExt.initialize!`, which explicitly initializes the XLA client with exactly one allowed GPU per local MPI rank and then sets the NCCL device accordingly.

ext/ReactantMPIExt/Nccl.jl
romanlee: Good idea. I renamed to `DEFAULT_COMM_*` to `NCCL_COMM_*`. And `DEFAULT_XLA_DEVICE` got separately refactored away.

test/runtests.jl
romanlee: Alright, I sorted this out. See https://github.com/EnzymeAD/Reactant.jl/pull/3136/commits/14a4e7631ae5df79a9de5946fe0706a77db5b9cd. 

Feel free to double check my logic, but it should be all good

src/Reactant.jl
```diff
@@ -272,6 +272,14 @@ const TracedType = Union{TracedRArray,TracedRNumber,MissingTracedValue}
 include("ControlFlow.jl")
 include("Tracing.jl")
 
+function default_nccl_comm_handle end
+
+function set_nccl_device!(::Integer)
```
wsmoses, line 277: I don't totally understand this, if we ship the nccl symbols and the cuda devices, why do we need cuda.jl for the translation?

romanlee: We don't necessarily need it, but it does simplify things slightly. The alternative would be something along these lines

```diff
diff --git a/deps/ReactantExtra/API.cpp b/deps/ReactantExtra/API.cpp
index 722f0e455..f5ea06db2 100644
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -885,6 +885,10 @@ void __cudaRegisterLinkedBinary_28b8d6c6_20_separate_callback_cu_a85cd5ea_29231(

 REACTANT_ABI int32_t ReactantHermeticCudaGetVersion() { return CUDA_VERSION; }

+REACTANT_ABI int32_t ReactantCudaSetDevice(int32_t device_id) {
+  return static_cast<int32_t>(cudaSetDevice(device_id));
+}
+
 REACTANT_ABI int32_t ReactantCudaDeviceGetComputeCapalilityMajor() {
   CUdevice cuDevice;
   ReactantHandleCuResult(cuDeviceGet(&cuDevice, 0));
@@ -1054,6 +1058,8 @@ REACTANT_ABI int32_t ReactantCudaDriverGetVersion() { return 0; }

 REACTANT_ABI int32_t ReactantHermeticCudaGetVersion() { return 0; }

+REACTANT_ABI int32_t ReactantCudaSetDevice(int32_t) { return -1; }
+
 REACTANT_ABI int32_t ReactantCudaDeviceGetComputeCapalilityMajor() { return 0; }

 REACTANT_ABI int32_t ReactantCudaDeviceGetComputeCapalilityMinor() { return 0; }
diff --git a/deps/ReactantExtra/BUILD b/deps/ReactantExtra/BUILD
index e362f6e10..31f01ac27 100644
--- a/deps/ReactantExtra/BUILD
+++ b/deps/ReactantExtra/BUILD
@@ -1201,6 +1201,7 @@ cc_library(
             "-Wl,-exported_symbol,_ReactantFuncSetArgAttr",
             "-Wl,-exported_symbol,_ReactantHermeticCudaGetVersion",
             "-Wl,-exported_symbol,_ReactantCudaDriverGetVersion",
+            "-Wl,-exported_symbol,_ReactantCudaSetDevice",
             "-Wl,-exported_symbol,_ReactantLLVMParseCommandLineOptions",
             "-Wl,-exported_symbol,_ReactantCudaDeviceGetComputeCapalilityMajor",
             "-Wl,-exported_symbol,_ReactantCudaDeviceGetComputeCapalilityMinor",
diff --git a/ext/ReactantCUDAExt.jl b/ext/ReactantCUDAExt.jl
index c52d8987d..0b2cd8024 100644
--- a/ext/ReactantCUDAExt.jl
+++ b/ext/ReactantCUDAExt.jl
@@ -36,8 +36,6 @@ const KA = KernelAbstractions

 Reactant.is_extension_loaded(::Val{:CUDA}) = true

-Reactant.set_nccl_device!(ordinal::Int) = CUDA.device!(ordinal)
-
 Base.Experimental.@MethodTable(REACTANT_CUDA_METHOD_TABLE)

 macro reactant_cuda_overlay(def)
diff --git a/src/Reactant.jl b/src/Reactant.jl
index b501c2707..5560237a0 100644
--- a/src/Reactant.jl
+++ b/src/Reactant.jl
@@ -274,10 +274,10 @@ include("Tracing.jl")

 function default_nccl_comm_handle end

-function set_nccl_device!(::Integer)
-    return error(
-        "GPU MPI requires CUDA.jl; load CUDA before initializing the NCCL communicator"
-    )
+function set_nccl_device!(ordinal::Integer)
+    status = MLIR.API.ReactantCudaSetDevice(ordinal)
+    status == 0 || error("cudaSetDevice($ordinal) failed with CUDA status $status")
+    return nothing
 end

 include("compiler/Compiler.jl")
diff --git a/src/mlir/libMLIR_h.jl b/src/mlir/libMLIR_h.jl
index c4a5ce779..449202f19 100755
--- a/src/mlir/libMLIR_h.jl
+++ b/src/mlir/libMLIR_h.jl
@@ -16682,6 +16682,10 @@ function ReactantHermeticCudaGetVersion()
     @ccall Reactant_jll.libReactantExtra.ReactantHermeticCudaGetVersion()::Int32
 end

+function ReactantCudaSetDevice(device_id)
+    @ccall Reactant_jll.libReactantExtra.ReactantCudaSetDevice(device_id::Int32)::Int32
+end
+
 function ReactantCudaDeviceGetComputeCapalilityMajor()
     @ccall Reactant_jll.libReactantExtra.ReactantCudaDeviceGetComputeCapalilityMajor()::Int32
 end
```

romanlee: Which is actually better than what I initially thought the alternative was. Maybe this is the best way then actually?

wsmoses: yeah we should do this -- at minimum it avoid an accidental "called the wrong cuda library" mismatch -- especially since our cuda/nccl is potentially different from the cuda.jl one

PR #3139, opened 2026-07-31
Stop KA ConcreteRArray allocation within compile

dkytezab:
```julia
using Reactant

y = Reactant.to_rarray(zeros(4))

f(x) = (fill!(y, 1.0); x .+ y)

x = Reactant.to_rarray(rand(4))

@compile f(x)
```
```
ERROR: LoadError: StackOverflowError:
Stacktrace:
     [1] _has_debugcache()
       @ Reactant.Compiler ~/.julia/packages/Reactant/NeviQ/src/compiler/Compiler.jl:1535
     [2] debugcache(; throw_error::Bool)
       @ Reactant.Compiler ~/.julia/packages/Reactant/NeviQ/src/compiler/Compiler.jl:1540
     [3] debugcache
       @ ~/.julia/packages/Reactant/NeviQ/src/compiler/Compiler.jl:1539 [inlined]
     [4] push_debug_stack!(fname::String, file::String, line::Int64)
       @ Reactant ~/.julia/packages/Reactant/NeviQ/src/utils.jl:682
     [5] call_with_reactant(::Reactant.EnsureReturnType{Any}, ::typeof(fill!), ::ConcreteIFRTArray{Float64, 1, Nothing}, ::Float64)
       @ Reactant ~/.julia/packages/Reactant/NeviQ/src/utils.jl:1195
     [6] fill!
       @ ~/.julia/packages/Reactant/NeviQ/src/ConcreteRArray.jl:811--- the above 2 lines are repeated 19992 more times ---
```

### DIFF
--- a/src/ConcreteRArray.jl
+++ b/src/ConcreteRArray.jl
@@ -780,12 +780,24 @@ function Base.zero(x::ConcreteIFRTArray{T,N}) where {T,N}
     )
 end
 
+# `fill!` on a concrete array is implemented by compiling a `fill!` kernel and running it. Reached
+# from inside a trace, that compile step traces `fill!`, which lands back on the same method, and
+# the recursion is unbounded. It shows up whenever an array is allocated *during* tracing, since
+# allocation routines such as `KernelAbstractions.zeros` fill their result before returning it.
+#
+# Broadcast assignment lowers directly and does not re-enter `fill!`, so it is the traced path.
+@inline function fill_while_tracing!(a, val)
+    a .= val
+    return a
+end
+
 function Base.fill!(a::ConcretePJRTArray{<:TracedRNumber}, val::TracedRNumber)
     throw(MethodError(fill!, (a, val)))
 end
 
 function Base.fill!(a::ConcretePJRTArray{T,N}, val) where {T,N}
     isempty(a) && throw("Cannot setindex! to empty buffer")
+    within_compile() && return fill_while_tracing!(a, val)
 
     wait(a)
     if buffer_on_cpu(a) && !Sharding.is_sharded(a)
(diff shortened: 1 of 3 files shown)


src/ConcreteRArray.jl
```diff
@@ -806,6 +818,7 @@ end
 
 function Base.fill!(a::ConcreteIFRTArray{T,N}, val) where {T,N}
     isempty(a) && throw("Cannot setindex! to empty buffer")
+    within_compile() && return fill_while_tracing!(a, val)
```
wsmoses: How would this happen we shouldn't ever have a concreterarray from within compile

dkytezab: ok yeah i see now that this MWE is a little misleading, seeing as we have the concreterarray that is captured in the function body only and is not included in the signature, but i think there is an issue with KAext related to this

dkytezab: Namely doesn't this produce this problem?
https://github.com/EnzymeAD/Reactant.jl/blob/0a24c520bc9e7e979d581fd3a9ce04963e65853c/ext/ReactantKernelAbstractionsExt.jl#L27-L35

wsmoses: okay that allocate should have an override if within compile to return zeros or something then

dkytezab: in oceananigans we have https://github.com/CliMA/Oceananigans.jl/blob/7a16203d1e8bc72f0c449d8a0cfd8fbd90d765c1/src/Grids/zeros_and_ones.jl#L8-L9 which then blocks us whenever we try to build a field inside of a compiled body.

wsmoses (commented):
can you add a test [and re-title the PR appropriately]

PR #3142, opened 2026-08-01
Add Ops.geqrf and Ops.orgqr for QR factorization

albertomercurio:
Wrap the `enzymexla.lapack.geqrf` and `enzymexla.lapack.orgqr` dialect ops,
which lower to LAPACKE on CPU, cuSOLVER on CUDA and XLA's Qr /
ProductOfElementaryHouseholderReflectors on TPU.

The high-level `enzymexla.linalg.qr` op is deliberately bypassed: its lowering
forwards the QR op's (Q, R, info) result types straight into geqrf's
(output, tau, info), so it only produces verifiable IR when Q and R have the
same type, and the "R" it yields is the packed matrix rather than a
triangular one.

The CPU lowering of geqrf hardcodes its result types and then replaces the
op's uses, so the declared types have to match exactly: packed factors with
the shape of the input, tau of length min(m, n), and a rank-0 BlasInt info
(not Int32 as for `enzymexla.linalg.lu`, which does respect declared types).

Also register the LAPACKE symbols the CPU lowering binds. These use the C
interface rather than the Fortran one, so they need their own table; the
lookup is non-throwing so a BLAS backend without LAPACKE does not break
loading Reactant.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -14,6 +14,7 @@ using ..Reactant:
     unwrapped_eltype
 using ReactantCore: ReactantCore
 using GPUArraysCore: GPUArraysCore
+using LinearAlgebra: BlasInt
 
 const GELU_APPROXIMATION_MAP = Dict(
     "NONE" => MLIR.API.ENZYMEXLA_GELU_APPROXIMATION_NONE,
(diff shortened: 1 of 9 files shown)


mofeing:
hey @albertomercurio

this was actually being worked out in #2877 but found out that I need to refactor some of the passes in Enzyme-JAX passes before offering the QR directly (it works but there are some performance pitfalls, we are no longer using LAPACKE bindings but LAPACK, ...) and life got in between before I could fix it all 😅.

if you don't mind, I would want to continue in that PR. we have a deadline this Monday but after it, my plan for this month is to work on LAPACK and MPI support.

I'm not closing this PR because we can merge both PRs once everything is worked out since there is more stuff here, but it shouldn't be merged yet because I'm gonna do breaking changes on QR in Enzyme-JAX.

albertomercurio:
Ok I didn't know that there was already a PR working on it. I am closing this

PR #3156, opened 2026-08-04
Enable speculate_partial_if in canonicalize-loops raising pass

Pangoraw:
## Summary

- Sets `speculate_partial_if=true` on both `func.func(canonicalize-loops)` invocations in the default raise pass string (`compile_mlir!` in `src/compiler/Compiler.jl`).

## Why

EnzymeAD/Enzyme-JAX#2697 templated `PartialIfToSelect` on a new `speculate_partial_if` option defaulting to `false`. Before that change, this pattern always ran in full-speculation mode, converting a partial `scf.if`/`affine.if` (where only some values are conditionally yielded) into a single `arith.select` of a precomputed expression, regardless of branch-body size. With the new default off, only trivial single-op branch bodies get this treatment; anything larger is left as an unconverted conditional, which then gets raised to StableHLO via a less optimal path.

## Evidence

Bisected against a real regression in Breeze.jl (NumericalEarth/Breeze.jl#884): the `CBL/compressible_explicit/1M_MixedNonEquilibrium [Float32]` benchmark dropped ~24% in throughput after bumping `Reactant_jll` across a range that picked up EnzymeAD/Enzyme-JAX#2697 (among ~73 other commits). Building `enzymexlamlir-opt` at the two pinned Enzyme-JAX commits and running the real raise pipeline (truncated to just after `raise-affine-to-stablehlo`) on the same traced IR:

| op | before pin | after pin (default) | after pin + `speculate_partial_if=true` |
|---|---|---|---|
| `stablehlo.select` | 504 | 1209 (+705) | 537 (+33) |
| `stablehlo.and` | 255 | 858 (+603) | 390 (+135) |
| `arith.select` | 2083 | 1552 (-531) | 2008 (-75) |

Re-enabling speculation for just this pass closes ~96% of the extra `stablehlo.select` count and ~78% of the extra `stablehlo.and` count introduced by the default flip. A small residual gap remains, likely from the LLVM 23→24 bump or other AffineCFG.cpp changes in the same commit range — this PR addresses the dominant contributor.

## Test plan

- [ ] CI passes (existing raising/lit tests)
- [ ] Re-run the Breeze.jl benchmark suite (CBL/compressible_explicit/1M_MixedNonEquilibrium, ScalarTendency/BF16) to confirm the regression recovers end-to-end, not just at the IR op-count level


wsmoses:
so this is definitely one option, however is there a different optimizationw e can do.

Part of why this was disabled was because it can accidentally produce the need for strong-zero where otherwise it wouldn't have been required

Pangoraw:
fair, let me try to recover the same ir with optimizations

Pangoraw:
@copilot make it an option like sroa attributor global ref

Pangoraw:
@copilot make default false

wsmoses:
Rather than a global maybe we make this a compile option? So it can be more easily ab tested

Pangoraw:
https://github.com/EnzymeAD/Reactant.jl/pull/3212

PR #3160, opened 2026-08-04
Hand CUDA kernel modules to MLIR as bitcode

### LINKED ISSUE #3137: [0.2.276] Some boundschecked loops in kernels fail to compile
MWE reduced by Claude below. For my actual project in this case (Terrarium.jl, based on Oceananigans.jl) I already found a workaround because the boundscheck wasn't actually done on purpose. 

This might be related to this LLVM change: https://discourse.llvm.org/t/rfc-enforce-single-operand-format-for-all-enable-metadata-nodes/90571

The real code doesn't use `OffsetArray` but `Oceananigans.Fields` at an offset. 

```Julia 
# MWE: Reactant 0.2.276 aborts the process when compiling a KernelAbstractions kernel
# that contains a bounds-checked counted loop.
#
#   $ julia --project=. reactant_loop_distribute_mwe.jl
#   Reactant v0.2.276
#   Expected one operand for llvm.loop.distribute metadata
#   !219 = !{!"llvm.loop.distribute.enable", i1 false}
#   LLVM ERROR: Broken module found, compilation aborted!
#   [9049] signal 6: Abort trap: 6
#
# Works on Reactant 0.2.275. CPU only — no GPU is needed; CUDA.jl is loaded only because
# Reactant compiles KernelAbstractions kernels through its CUDA extension.
#
# Project.toml:
#   [deps]
#   CUDA                 = "052768ef-5323-5732-b1bb-66c8b64840ba"
#   KernelAbstractions   = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
#   OffsetArrays         = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
#   Reactant             = "3c362404-f566-11ee-1572-e11a4b42c853"
#
# ---------------------------------------------------------------------------------------
# Cause
#
# The loop below is a counted loop with a *dynamic* trip count (`length(zs)`, from a view
# with runtime bounds) over an array with *runtime* offsets, so the bounds check cannot be
# hoisted out of the loop. LLVM's IRCE pass therefore splits it into pre/main/post loops and
# tags the clones via DisableAllLoopOptsOnLoop() in
# llvm/lib/Transforms/Utils/LoopConstrainer.cpp, which emits the two-operand form
#
#     !{!"llvm.loop.distribute.enable", i1 false}
#
# LLVM 18 (Julia 1.12's LLVM) both emits and accepts that. Reactant serialises the Julia
# module to text and re-parses it with its own, much newer LLVM (ConvertLLVMStrToMLIR in
# ext/ReactantCUDAExt.jl). That LLVM contains llvm/llvm-project#201077 (merged 2026-07-21),
# which mandates the single-operand form `!{!"llvm.loop.distribute.enable"}` and adds a
# verifier check for it — with no AutoUpgrade for existing IR. The verifier then calls
# report_fatal_error, which aborts the whole Julia process.
#
# Reactant 0.2.275 -> Reactant_jll 0.0.397 (old LLVM, fine)
# Reactant 0.2.276 -> Reactant_jll 0.0.399 (new LLVM, aborts)
# ---------------------------------------------------------------------------------------

using Reactant, CUDA, KernelAbstractions, OffsetArrays

@kernel function findfirst_kernel!(out, zs, sat)
    i, j = @index(Global, NTuple)
    n = length(zs)                                     # dynamic trip count
    idx = -1
    for k in 1:n
        found = (idx < 0) & (sat[i, j, k] < 1.0f0)     # bounds-checked, runtime offsets
        idx = ifelse(found, k, idx)
    end
    out[i, j] = zs[ifelse(idx > 0, idx, n)]
end

function run!(out, zs, sat)
    findfirst_kernel!(KernelAbstractions.get_backend(parent(out)))(
        out, zs, sat; ndrange = (1, 1)
    )
    return nothing
end

zs = view(Reactant.to_rarray(collect(range(-1.0f0, 0.0f0; length = 13))), 1:11)
sat = OffsetArray(Reactant.to_rarray(rand(Float32, 3, 3, 16)), -1, -1, -3)
out = OffsetArray(Reactant.to_rarray(zeros(Float32, 3, 3)), -1, -1)

println("Reactant v", pkgversion(Reactant))
f = Reactant.@compile run!(out, zs, sat)
f(out, zs, sat)
println("compiled and ran OK: ", Array(parent(out)))
```

wsmoses:
Fixes #3137. Requires `Reactant_jll` >= 0.0.404, which exports the new `ConvertLLVMBCToMLIR` entry point; the compat bound is bumped here.

Independent of #3159 — that PR carries the C++ source for the entry point and should still merge so the tree matches the JLL, but this one does not depend on it and can go in on its own.

### Problem

`ReactantCUDAExt` was moving the compiled kernel module from Julia's LLVM into Reactant's LLVM by printing textual IR and reparsing it. Textual IR is not stable across LLVM versions — `LLParser` performs no auto-upgrade.

Julia 1.12 (LLVM 18) emits the two-operand loop metadata form:

```llvm
!{!"llvm.loop.distribute.enable", i1 false}
```

Current LLVM enforces the single-operand form in the Verifier, so the reparse aborts the whole Julia process:

```
Expected one operand for llvm.loop.distribute metadata
LLVM ERROR: Broken module found, compilation aborted!
```

The trace pins it to the textual parser specifically:

```
ConvertLLVMStrToMLIR → llvm::parseIR → LLParser::Run
  → LLParser::validateEndOfModule → UpgradeDebugInfo → report_fatal_error
```

`UpgradeDebugInfo` is where `verifyModule` runs. LLVM *does* know how to rewrite the old form (`upgradeInstructionLoopAttachment` in `AutoUpgrade.cpp`), but its only caller is `Bitcode/Reader/MetadataLoader.cpp` — the `.ll` parser never reaches it.

The bitcode path has the opposite ordering: `BitcodeReader::materializeModule` calls `materializeMetadata()` (which runs the upgrade) first and `UpgradeDebugInfo` last, so the metadata is fixed up before the verifier ever sees it. MLIR's `LoopAnnotationImporter` already understands the new single-operand spelling, so this works end to end.

### Change

Hand the module over as bitcode: `convert(Vector{UInt8}, mod)` (LLVM.jl's bitcode writer) → `ConvertLLVMBCToMLIR`, plus the corresponding binding in `libMLIR_h.jl` matching what the generator emits.

### Verification

Against the published `Reactant_jll` 0.0.404, Julia 1.12.4, RTX 5090:

```
Reactant v0.2.279
libReactantExtra: ~/.julia/artifacts/3f76162f.../lib/libReactantExtra.so
compiled and ran OK: Float32[0.0 0.0 0.0; 0.0 -1.0 0.0; 0.0 0.0 0.0]
```

The same MWE on `main` with jll 0.0.403 aborts with `signal 6`.

https://claude.ai/code/session_01PH4zH9zvavbXWXomJbCq3B


wsmoses:
Verified end to end against the published `Reactant_jll` 0.0.404 (compat bumped in aabc950), Julia 1.12.4, RTX 5090:

```
Reactant v0.2.279
libReactantExtra: ~/.julia/artifacts/3f76162f.../lib/libReactantExtra.so
compiled and ran OK: Float32[0.0 0.0 0.0; 0.0 -1.0 0.0; 0.0 0.0 0.0]
```

`nm -D` on that artifact confirms `ConvertLLVMBCToMLIR` is exported.

The same MWE on `main` with jll 0.0.403 aborts the process with `signal 6`, and the trace pins it to the textual parser specifically:

```
ConvertLLVMStrToMLIR → llvm::parseIR → LLParser::Run
  → LLParser::validateEndOfModule → UpgradeDebugInfo → report_fatal_error
```

`UpgradeDebugInfo` is where `verifyModule` runs. The bitcode path has the opposite ordering — `BitcodeReader::materializeModule` calls `materializeMetadata()` (which runs `upgradeInstructionLoopAttachment` via `MetadataLoader.cpp:2632`) first, and `UpgradeDebugInfo` last — so the metadata is upgraded before the verifier ever sees it.

PR #3169, opened 2026-08-09
Give a plain array asked for a traced eltype a TracedRArray

### LINKED ISSUE #3167: Stack overflow with matrix multiplication between traced / untraced array
This errors dramatically at the compilation stage with a stack overflow:

```julia
A, B, C = 9, 40, 50 # Might need to be bigger depending on setup, not sure, but this fails for me

struct MatMul{X}
    X::X
end
m = MatMul(ones(B, C))
function (m::MatMul)(W)
    return sum(W * m.X)
end
p = ones(A, B)
@info m(p)

using Reactant
p2 = Reactant.to_rarray(p)
m2 = @compile m(p2)
@info m2(p2)
```

The error isn't super helpful. However, prior to me minimising this I got some stack traces which complained about type inference in `typed_vcat`.

```
julia> m2 = @compile m(p2)

Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
ERROR: StackOverflowError:
```

This can be worked around on the user side by converting `m.X` to a traced array as well, although having to do that is kind of annoying if `m.X` is supposed to be a compile-time constant.

```julia
mtraced = Reactant.to_rarray(m)
m3 = @compile mtraced(p2)
@info m3(p2)
```

wsmoses:
Fixes #3167.

> Rewritten: the first version of this PR overlaid 2-argument `*`, which broke `x' * y` (see "First attempt" at the bottom). This targets the root cause instead.

### Cause

`W * X` where `W` is traced and `X` is an ordinary array. Generic Base algorithms allocate their destination from whichever operand they were handed:

```julia
(*)(A::AbstractMatrix, B::AbstractMatrix) =
    mul!(similar(B, promote_op(matprod, eltype(A), eltype(B)), (size(A,1), size(B,2))), A, B)
```

`promote_op(matprod, TracedRNumber{Float64}, Float64)` is `TracedRNumber{Float64}`, and `similar` is called on **`B`** — the *untraced* operand — so the destination came back as an `Array{TracedRNumber{Float64},2}`, an array-of-structs, rather than a `TracedRArray`.

The contraction itself was still a single correct `dot_general`, but the result was then written into that AoS destination **one element at a time** and reassembled with a `concatenate` (the `typed_vcat` the reporter saw in inference). Each scalar write is traced, so constant-propagation recursion grows with the number of output elements and eventually overflows the stack.

This predicts the size dependence exactly — failure tracks the output element count `A*C`, not the contraction dimension `B`:

| A | B | C | A·C | result |
|---|---|---|---|---|
| 9 | 32 | 32 | 288 | ok |
| 9 | 40 | 50 | 450 | `StackOverflowError` |
| 2 | 40 | 50 | 100 | ok |
| 9 | 40 | 2 | 18 | ok |
| 9 | **2** | 50 | 450 | `StackOverflowError` |

It also explains the reported workaround: with `X` traced, `similar` returns a `TracedRArray` and the elementwise path never happens.

### Fix

The concrete array types already redirect exactly this case — see the existing `similar` methods for `ConcretePJRTArray`/`ConcreteIFRTArray`, whose comment describes the same hazard. Plain `Array` was missing one.

The method is restricted to arrays of untraced primitives (`Array{T} where T<:ReactantPrimitive`). `Array{TracedRNumber{...}}` is itself an `Array`, so without that restriction an existing array-of-structs asking for another one would have been silently redirected to SoA; only the plain-to-traced widening is caught.

Callers are narrower than the signature suggests. Instrumenting the method across matmul, broadcast, `map`, `hcat`, `kron`, `cov` and `\` shows only two reach it, both in matmul, and both with a plain `Float64` source:

```
1  Float64  <- *@matmul.jl:60
1  Float64  <- matprod_dest@matmul.jl:129
```

### Effect

Not just a crash fix — every mixed traced/untraced matmul was emitting O(n) scalar ops. Unoptimized IR for `9x32 * 32x32` (a size that stayed just under the crash threshold):

| | before | after |
|---|---|---|
| IR lines | 3474 | **10** |
| `stablehlo.dot_general` | 1 | 1 |
| `stablehlo.reshape` | 289 | **0** |
| `stablehlo.broadcast_in_dim` | 290 | **0** |
| `stablehlo.concatenate` | 1 | **0** |

### Testing

New testset in `test/integration/linear_algebra.jl` covering the reported reproducer, both operand orders, the matrix–vector case, an assertion that the IR contains exactly one `dot_general` and no `concatenate`, and one that an array-of-structs asking for another one still gets it.

Running identical workloads with the old AoS behaviour restored versus this patch, exactly one outcome changes:

| workload | before | after |
|---|---|---|
| `mat * mat`, untraced rhs | `StackOverflowError` | ok |
| `mat * vec`, untraced rhs | ok | ok |
| untraced lhs `*` mat | ok | ok |
| `adjoint(traced) * traced vec` | ok | ok |
| `cov(v, v)` | ok | ok |
| broadcast / `map` / `hcat`, mixed | ok | ok |

Three workloads fail identically before and after, so they are pre-existing and out of scope here: `adjoint(traced) * untraced vec` (`no method matching dot_general`), `kron` with an untraced operand, and untraced matrix `\` traced vector (both "Scalar indexing is disallowed"). Happy to file those separately.

Suites, locally on CPU (Julia 1.12.4):

| suite | result |
|---|---|
| `integration/linear_algebra` | 125 pass / 0 fail (119 + 6 new) |
| `integration/statistics` | 41 / 0 |
| `core/array_ops` | 290 / 0 |
| `core/map_reduction` | 148 / 0 |
| `core/bcast` | 16 / 0 |
| `integration/cuda` | 26 / 0 |

### First attempt, for the record

The original version of this PR added a `@reactant_overlay` for 2-argument `*`. That was wrong: `Adjoint{T,<:AbstractVector}` is an `AbstractMatrix`, so `x' * y` matched the overlay, and since overlays win over more-specific Base methods it started returning a 1-element array instead of the scalar `dot`. That broke `Statistics.cov` and turned CI red. Carving the case out inside the overlay body does not work either — the `call_with_native` fallback runs Base's method without tracing inner calls, so the `dot` inside hits "Scalar indexing is disallowed". Fixing `similar` avoids the dispatch hazard entirely and fixes any generic Base algorithm that allocates this way, not only `*`.

https://claude.ai/code/session_01PH4zH9zvavbXWXomJbCq3B

### DIFF
--- a/src/Overlay.jl
+++ b/src/Overlay.jl
@@ -393,6 +393,30 @@ end
     end
 end
 
+# `*` allocates its destination from one of its operands -- `similar(B, TS, ...)` for
+# matrix-matrix, `similar(x, TS, axes(A, 1))` for matrix-vector -- so a traced matrix times
+# an untraced right-hand side lands in an `Array{TracedRNumber}` that then costs one traced
+# op per element to fill, overflowing inference's constant-propagation stack once the output
+# is large enough (#3167). Only this operand order is affected: with the right-hand side
+# traced, `similar` is called on a `TracedRArray` and already does the right thing.
+#
+# Deliberately restricted to genuine traced matrices rather than `AbstractMatrix`:
+#  * an `Adjoint`/`Transpose` of a *vector* is also an `AbstractMatrix`, but `x' * y` is a
+#    scalar dot product, and an overlay matching it would win over Base's more specific
+#    method and silently return a 1-element array;
+#  * structured wrappers (`Diagonal`, `UpperTriangular`, ...) keep Base's dispatch, which
+#    picks destinations that preserve their structure.
+@reactant_overlay function Base.:(*)(
+    a::Union{
+        TracedRArray{T,2},
+        LinearAlgebra.Adjoint{TracedRNumber{T},<:TracedRArray{T,2}},
+        LinearAlgebra.Transpose{TracedRNumber{T},<:TracedRArray{T,2}},
+    },
+    b::AbstractVecOrMat,
+) where {T}
+    return call_with_native(TracedLinearAlgebra.overloaded_mul, a, b)
+end
+
 # 3 arg multiplication is specialized in Base, but we can reorder the computation
 # as an MLIR optimization
 @reactant_overlay function Base.:(*)(a::AbstractArray, b::AbstractArray, c::AbstractArray)
--- a/test/integration/linear_algebra.jl
+++ b/test/integration/linear_algebra.jl
@@ -119,6 +119,54 @@ end
     @test C5_ra ≈ C5 atol = 1e-3 rtol = 1e-2
 end
 
+# A traced operand multiplied by an ordinary array used to allocate its destination via
+# `similar(::Matrix, TracedRNumber, ...)`, i.e. an array-of-structs, and then fill it one
+# traced op per element. Big enough outputs overflowed inference's const-prop stack.
+struct ConstMatMul{X}
+    X::X
+end
+(m::ConstMatMul)(W) = sum(W * m.X)
+
+@testset "Multiplication with an untraced operand (#3167)" begin
+    A, B, C = 9, 40, 50
+    X = Reactant.TestUtils.construct_test_array(Float32, B, C)
+    W = Reactant.TestUtils.construct_test_array(Float32, A, B)
+    W_ra = Reactant.to_rarray(W)
+
+    m = ConstMatMul(X)
+    @test @jit(m(W_ra)) ≈ m(W) rtol = 1e-3
+
+    mul_const_rhs(w) = w * X
+    @test Array(@jit(mul_const_rhs(W_ra))) ≈ W * X rtol = 1e-3
+
+    # the untraced operand on the left, and the matrix-vector case
+    X_ra = Reactant.to_rarray(X)
+    mul_const_lhs(x) = W * x
+    @test Array(@jit(mul_const_lhs(X_ra))) ≈ W * X rtol = 1e-3
+
+    # matrix-vector allocates its destination separately from matrix-matrix (and does not
+    # go through `matprod_dest`), so exercise it with an output big enough to have blown
+    # the const-prop stack too
+    v = Reactant.TestUtils.construct_test_array(Float32, B)
+    mul_const_vec(w) = w * v
+    @test Array(@jit(mul_const_vec(W_ra))) ≈ W * v rtol = 1e-3
+
+    tall = Reactant.TestUtils.construct_test_array(Float32, 600, B)
+    tall_ra = Reactant.to_rarray(tall)
+    @test Array(@jit(mul_const_vec(tall_ra))) ≈ tall * v rtol = 1e-3
+
+    # one `dot_general`, not one op per output element
+    hlo = repr(@code_hlo(optimize = false, mul_const_rhs(W_ra)))
+    @test count("stablehlo.dot_general", hlo) == 1
+    @test count("stablehlo.concatenate", hlo) == 0
+
+    # `x' * y` is a scalar dot product, not a matmul; it must not be diverted
+    u = Reactant.TestUtils.construct_test_array(Float32, B)
+    u_ra = Reactant.to_rarray(u)
+    dot_prod(x) = x' * x
+    @test Reactant.to_number(@jit(dot_prod(u_ra))) ≈ u' * u rtol = 1e-3
+end
+
 @testset "triu & tril" begin
     A = Reactant.TestUtils.construct_test_array(Float32, 4, 6)
     A_ra = Reactant.to_rarray(A)


wsmoses:
Good push — I went and traced where the `Array{TracedRNumber}` is actually born. Short answer: it is `similar`, and there are three converging paths, only one of which has a designed hook.

| path | allocation site | hook |
|---|---|---|
| `mat * mat`, Julia ≥ 1.11 | `matprod_dest(A, B, TS)` → `similar(B, TS, dims)` | `LinearAlgebra.matprod_dest` |
| `mat * vec`, all versions | `similar(x, TS, axes(A, 1))`, inline in `*` (matmul.jl:57/60) | none |
| both, Julia 1.10 | `similar(B, TS, ...)`, inline in `*` (matmul.jl:53/57/106) | none — `matprod_dest` was added in 1.11 |

Instrumenting the method across matmul, broadcast, `map`, `hcat`, `kron`, `cov` and `\` confirms only matmul reaches it, and only with a plain source:

```
1  Float64  <- *@matmul.jl:60
1  Float64  <- matprod_dest@matmul.jl:129
```

So overloading `matprod_dest` would be the tidier hook, but it fixes only matrix-matrix on 1.11+. Matrix-vector is affected too — it is just less visible because the output is usually small. Measured on current main:

| case | output elems | IR lines | reshapes | concat |
|---|---|---|---|---|
| `mat*mat` 9x20 | 180 | 2178 | 181 | 1 |
| `mat*mat` 9x50 | 450 | StackOverflow | | |
| `mat*vec` 9x40 * 40 | 9 | 99 | 10 | 1 |
| `mat*vec` 600x40 * 40 | 600 | **StackOverflow** | | |

That last row is the point: matrix-vector blows up in exactly the same way once the result is big enough, and a `matprod_dest` overload would not touch it, nor anything on 1.10.

`similar` is the single point all three converge on, so that is where I left the fix. I have added the reasoning as a comment next to the method so it does not look like an arbitrarily broad override, and added the 600-element matrix-vector case to the tests since it is a distinct path from the matrix-matrix one.

Happy to switch to `matprod_dest` plus a separate matrix-vector path plus a 1.10 fallback if you would rather have the designed hook where one exists — it is three code paths instead of one, which is why I did not start there, but it is your call.

wsmoses:
Followed the thread further and instrumented Reactant's own side, since "where do the N ops actually come from" is the other half of the question.

They come from `aos_to_soa` (src/Reactant.jl:160), called by the `mul!` overlay on the destination Base allocated:

```julia
function aos_to_soa(x::Array{TracedRNumber{T}}) where {T}
    isa_traced_soa(ancestor(x)) && return x
    for i in eachindex(x)
        if !isassigned(x, i)
            x[i] = promote_to(TracedRNumber{T}, 0)   # N constants
        end
    end
    return @opcall reshape(vcat(x...), size(x)...)   # N-way vcat
end
```

Instrumented on a `9x40 * 40x20` product, it is called exactly once, with all 180 elements unassigned:

```
aos_to_soa calls: [((9, 20), 180, 180)]     # (size, length, #undef)
N=180   broadcast=182   reshape=181   concat=1
```

which accounts for the IR exactly: 180 `broadcast_in_dim` from the `promote_to(..., 0)` fill, 180 `reshape` + 1 `concatenate` from `vcat(x...)`. Every one of those zeros is then discarded, because `overloaded_mul!` overwrites the whole buffer (`β == 0`). The `vcat` is also where the reporter's `typed_vcat` came from.

So I tried the best case for an override-level fix: shortcut `aos_to_soa` when the array is entirely unassigned (no per-element work needed at all), keeping Base's AoS destination.

| approach | N=180 | N=450 | ops |
|---|---|---|---|
| main | 2178 lines, 182 bcast, 181 reshape, 1 concat | StackOverflow | ~3N |
| `aos_to_soa` shortcut, AoS destination kept | 1815 lines, 2 bcast, 180 reshape | 4515 lines, 450 reshape | ~N |
| `similar` returning a `TracedRArray` (this PR) | 10 lines | 10 lines | O(1) |

The shortcut is a real improvement — it deletes the wasted zero-fill and the `vcat`, and pushes the crash threshold past 450 — but it stays linear, because the AoS destination still has to be filled element by element on the way back out (`C .= C2` in the overlay), and anything downstream that consumes the returned `Array{TracedRNumber}` pays another N to convert it again. It moves the cliff rather than removing it.

That is why I think the allocation site is the right place: it is the only change that makes this O(1) rather than a larger constant.

Separately, the `aos_to_soa` shortcut looks worth having on its own merits for any other path that hands it an all-undef array — happy to open that as a separate PR if you want it, but it is not a fix for #3167.

wsmoses:
Reworked: no `Base.similar` override any more. You were right that `similar(::Array, T, dims)` has to return an `Array` — Reactant can define that for its own array types (as it does for `ConcretePJRTArray`), but not for `Base.Array`.

I tried `LinearAlgebra.matprod_dest` next, since it is the designed hook, and it does not work either — the tests caught it with 6 failures:

```
MethodError: matprod_dest(::Diagonal{TracedRNumber{Float32}, TracedRArray{Float32,1}},
                          ::TracedRArray{Float32,2}, ::Type{TracedRNumber{Float32}}) is ambiguous.
```

LinearAlgebra defines about eight `matprod_dest` methods specialized on `StructuredMatrix`, so dispatching on the element type is ambiguous with all of them. And `matprod_dest` is not matmul-specific: `/`, `\`, `ldiv!`, `_rdiv!` and `hessenberg` route through it too, and several overloads deliberately return *structured* destinations (`_matprod_dest_diag(A::UnitUpperTriangular, TS) = UpperTriangular(similar(parent(A), TS))`). Taking it over on element type would flatten those to dense well outside matmul, and disambiguating would mean mirroring LinearAlgebra's method lattice and rechecking it every Julia version.

So this is now a single narrow overlay, dispatching on genuine traced matrices:

```julia
@reactant_overlay function Base.:(*)(
    a::Union{
        TracedRArray{T,2},
        LinearAlgebra.Adjoint{TracedRNumber{T},<:TracedRArray{T,2}},
        LinearAlgebra.Transpose{TracedRNumber{T},<:TracedRArray{T,2}},
    },
    b::AbstractVecOrMat,
) where {T}
    return call_with_native(TracedLinearAlgebra.overloaded_mul, a, b)
end
```

It is narrow by construction rather than by a runtime check:

* an `Adjoint`/`Transpose` of a *vector* cannot match, so `x' * y` stays Base's scalar dot product — that is the regression that reddened CI on the first attempt, and it is now unrepresentable rather than guarded;
* structured wrappers keep Base's dispatch, so `Diagonal`/`UpperTriangular` destinations stay structured (the "Wrapper Types Matrix Multiplication" testset is back to 7/7);
* `AbstractVecOrMat` covers matrix-matrix and matrix-vector in one method on every supported Julia, with no `matprod_dest` version split.

Only the traced-LHS × untraced-RHS order needs this. With the right-hand side traced, `similar` is called on a `TracedRArray` and already does the right thing — measured:

| | before | after |
|---|---|---|
| `traced_mat * plain_vec` | 99 lines, 10 reshapes | 10 lines, 0 |
| `traced_mat * plain_mat` | StackOverflow at 450 out | 10 lines, 0 |
| `plain_mat * traced_vec` | 10 | 10 |
| `plain_mat * traced_mat` | 10 | 10 |

Matrix-vector was degraded too and would overflow on its own at 600 outputs; it is covered here and has a test.

Suites on CPU (Julia 1.12.4): `linear_algebra` 126/0, `statistics` 41/0, `array_ops` 290/0, `map_reduction` 148/0, `bcast` 16/0.

PR #3174, opened 2026-08-09
Query ROCm device properties for kernel lowering instead of sm_XX garbage

wsmoses:
Fixes issue 3 of @ftynse's [ROCm findings](https://gist.github.com/ftynse/5c3816a1cea5daa7dc189b937b62f9ca): eager KernelAbstractions kernels fail deep inside codegen on ROCm with no indication of the actual problem.

**Reworked per review** — ROCm no longer raises like TPU; it queries real device properties (or falls back to reasonable defaults on older jlls).

## The bug, in two layers

1. **`XLA.device_properties` returned uninitialized memory on non-CUDA platforms.** `src/xla/Device.jl` created `Ref{MLIR.API.DeviceProperties}()`, only filled it when `pname == "cuda"`, warned otherwise — then read and *cached* the unfilled Ref.
2. **The GPU branch of `compile_mlir!` built `cubinChip = "sm_$(major)$(minor)"` from those properties unconditionally**, so on ROCm every compile fed a garbage `sm_XX` string into `lower-jit`.

## The fix

**C++ (`API.cpp` + `BUILD`)** — mirror the CUDA query for HIP:
- `ReactantHipDeviceGetProperties(DeviceProperties*, device_id)` via `hipGetDeviceProperties` (`hipDeviceProp_t` is field-for-field compatible with the CUDA-derived `DeviceProperties` struct, so no ABI change).
- `ReactantHipDeviceGetGCNArchName(device_id)` returning the device's LLVM target id, e.g. `"gfx950:sramecc+:xnack-"` (same source XLA's own `RocmExecutor` uses).
- Real under `REACTANT_ROCM`, throwing stubs otherwise; `@local_config_rocm//rocm:rocm_headers` added to the `if_rocm` deps for the include.

**Julia:**
- `XLA.device_properties` dispatches on platform: CUDA query, HIP query on `"rocm"` — with a **fallback to conservative CDNA defaults** (wave64, 64 KiB LDS, 1024 threads/block, …, plus a warning) when running against a libReactantExtra that predates the new entry points, detected via `dlsym`. Platforms with no implementation now error instead of returning uninitialized memory.
- New `XLA.device_gcn_arch(device)` with the same old-jll fallback (`"gfx942"`, warned).
- `compile_mlir!` derives the lowering target from the target id on ROCm: chip = token before the first `:` (`gfx950`), features = remaining `name±` tokens converted to LLVM's `±name` spelling (`+sramecc,-xnack`). CUDA keeps the `sm_XX` / `+ptxNN` path untouched.
- Bindings hand-added to `libMLIR_h.jl` in generator style (the regenerate workflow will keep them).

## Verification

On CUDA hardware (RTX 5090): end-to-end `@jit` compile unchanged, `device_properties` still `sm_120`; target-id→chip/features conversion unit-tested (`gfx950:sramecc+:xnack-` → `gfx950` / `+sramecc,-xnack`, bare `gfx1100` → empty features); the fallback defaults struct constructs; `device_gcn_arch` correctly rejects non-ROCm devices. The HIP C++ block passes a standalone syntax check and mirrors `ReactantCudaDeviceGetProperties` 1:1 — **no AMD GPU available locally**, so the ROCm build/runtime legs need testing on real hardware (@ftynse). Needs a jll rebuild for the new entry points; until then ROCm hits the warned fallback path rather than garbage.

https://claude.ai/code/session_015drPT8jCwS74HnX6vEBCxz

### DIFF
--- a/src/compiler/Compiler.jl
+++ b/src/compiler/Compiler.jl
@@ -327,14 +327,30 @@ function compile_mlir!(
             jit = "lower-jit{openmp=$(OpenMP[]) backend=cpu},symbol-dce"
         end
     else
+        is_rocm = lowercase(XLA.platform_name(client)) == "rocm"
+        gpu_backend = is_rocm ? "rocm" : "cuda"
+        # HSACO serialization links the ROCm device libraries from the toolkit.
+        is_rocm && (toolkit = get(ENV, "ROCM_PATH", "/opt/rocm"))
+
         kern = if is_raising
             "lower-kernel{backend=cpu},symbol-dce,canonicalize"
         else
-            "lower-kernel,canonicalize"
+            "lower-kernel{backend=$(gpu_backend)},canonicalize"
         end
 
-        device_properties = XLA.device_properties(XLA.default_device(client))
-        cubinChip = "sm_$(device_properties.major)$(device_properties.minor)"
+        if is_rocm
+            # LLVM target id, e.g. "gfx950:sramecc+:xnack-": the token before
+            # the first ':' is the chip; the remaining tokens are target
+            # features in "name±" form, which LLVM spells "±name".
+            arch = XLA.device_gcn_arch(XLA.default_device(client))
+            parts = split(arch, ':')
+            cubinChip = String(first(parts))
+            chipFeatures = join(("$(p[end])$(p[1:(end - 1)])" for p in parts[2:end]), ",")
+        else
+            device_properties = XLA.device_properties(XLA.default_device(client))
+            cubinChip = "sm_$(device_properties.major)$(device_properties.minor)"
+            chipFeatures = cubinFeatures()
+        end
 
         if DEBUG_KERNEL[]
             curesulthandler = dlsym(
@@ -346,7 +362,7 @@ function compile_mlir!(
         else
             extra_lowerjit_options = ""
         end
-        jit = "lower-jit{$(extra_lowerjit_options)cuOptLevel=$(cuOptLevel[]) cubinFormat=$(cubinFormat[]) indexBitWidth=$(cuindexBitWidth[])  cubinChip=$(cubinChip) cubinFeatures=$(cubinFeatures()) run_init=true toolkitPath=$toolkit},symbol-dce"
+        jit = "lower-jit{$(extra_lowerjit_options)backend=$(gpu_backend) cuOptLevel=$(cuOptLevel[]) cubinFormat=$(cubinFormat[]) indexBitWidth=$(cuindexBitWidth[])  cubinChip=$(cubinChip) cubinFeatures=$(chipFeatures) run_init=true toolkitPath=$toolkit},symbol-dce"
     end
 
     recognize_comms = true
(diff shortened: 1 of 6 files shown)


wsmoses:
Follow-up: the enzymexla side of this — an actual amdgcn/ROCDL lowering path in `lower-jit`/`lower-kernel` so the chip string this PR now provides gets consumed — is being PR'd to Enzyme-JAX (ROCDL pipeline + `hipModuleLoadData`/`hipModuleGetFunction`/`hipModuleLaunchKernel` launch ABI + ROCM FFI handler registration). This PR's `backend=rocm` pass options are inert on the currently-pinned enzyme_ad (the option exists and defaults to cuda; verified CUDA compiles are unaffected with the option passed explicitly) and activate once the Enzyme-JAX commit is pulled.

src/compiler/Compiler.jl
```diff
@@ -327,14 +327,30 @@ function compile_mlir!(
...
 
-        device_properties = XLA.device_properties(XLA.default_device(client))
-        cubinChip = "sm_$(device_properties.major)$(device_properties.minor)"
+        if is_rocm
+            # LLVM target id, e.g. "gfx950:sramecc+:xnack-": the token before
+            # the first ':' is the chip; the remaining tokens are target
+            # features in "name±" form, which LLVM spells "±name".
+            arch = XLA.device_gcn_arch(XLA.default_device(client))
+            parts = split(arch, ':')
+            cubinChip = String(first(parts))
```
ftynse, line 347: Cubin?

src/compiler/Compiler.jl
```diff
@@ -327,14 +327,30 @@ function compile_mlir!(
...
+            chipFeatures = join(("$(p[end])$(p[1:(end - 1)])" for p in parts[2:end]), ",")
```
ftynse, line 348: Would be a good idea to check the syntax with +/- flags instead of blindly assuming here

src/compiler/Compiler.jl
```diff
@@ -327,14 +327,30 @@ function compile_mlir!(
...
+        else
+            device_properties = XLA.device_properties(XLA.default_device(client))
+            cubinChip = "sm_$(device_properties.major)$(device_properties.minor)"
```
ftynse, line 351: Are we sure about sm_ ?

src/xla/Device.jl
```diff
@@ -52,13 +52,80 @@ function device_properties(device::AbstractDevice)
...
+end
+
+# Conservative CDNA (Instinct-class) defaults, for libReactantExtra builds that
+# predate `ReactantHipDeviceGetProperties`. Sizes we cannot guess are 0.
+function _default_rocm_device_properties()
+    return MLIR.API.DeviceProperties(
+        0,                          # totalGlobalMem (unknown)
+        65536,                      # sharedMemPerBlock: 64 KiB LDS
+        65536,                      # regsPerBlock
+        64,                         # warpSize: CDNA wavefront
```
ftynse, line 91: Nit, this changes in gfx12**

ftynse (approved):
You may want a second look on the Julia side

PR #3175, opened 2026-08-09
Disable XLA command buffers by default on ROCm

wsmoses:
Mitigates issue 6 of @ftynse's [ROCm findings](https://gist.github.com/ftynse/5c3816a1cea5daa7dc189b937b62f9ca): `SIGSEGV` in `rocr::core::InterceptQueue::Submit` → `RocmCommandBuffer::LaunchGraph` on the first execution of a while-loop command buffer, on gfx950 / ROCm 7.1 — deterministic, after compile/spinup succeeded, and gone entirely with command buffers off (their `smoke08` vs `smoke09` runs).

## Change

`XLA.get_debug_options` / `make_compile_options` now take the client's `platform`; on `"rocm"` the `xla_gpu_enable_command_buffer` list is cleared — the per-compile equivalent of XLA's own documented escape hatch `--xla_gpu_enable_command_buffer=`. Both compile paths (PJRT `compile_mlir!` and the IFRT resharding path) pass the platform through.

The override ordering keeps this a *default*, not a lock-out: user kwargs are applied after, so `xla_debug_options=(; xla_gpu_enable_command_buffer=[...])` re-enables them — the right knob for testing whether a newer ROCm stack still crashes.

## Cost

Real and worth stating: a time-stepping loop is many small kernels, exactly the shape command buffers accelerate — disabling them pays full per-launch overhead every iteration. But a deterministic crash on the first executed step costs more, and the flag flip back is one line once the HSA/rocprofiler/XLA interaction is isolated upstream.

## Verification

Unit-checked on CUDA hardware (no AMD GPU available locally): `platform="rocm"` produces an empty command-buffer list, `platform="cuda"` keeps XLA's defaults (CONDITIONAL, CUBLAS, …, FUSION), an explicit user override wins over the ROCm default, and an end-to-end `@jit` compile on CUDA is unaffected.

https://claude.ai/code/session_015drPT8jCwS74HnX6vEBCxz

### DIFF
--- a/src/compiler/Compiler.jl
+++ b/src/compiler/Compiler.jl
@@ -1337,6 +1337,7 @@ function compile_xla(
                 num_replicas=mlir_fn_res.num_replicas,
                 num_partitions=mlir_fn_res.num_partitions,
                 mesh_ids=mlir_fn_res.is_sharded ? global_device_ids : nothing,
+                platform=lowercase(XLA.platform_name(client)),
                 xla_debug_options=compile_options.xla_debug_options,
                 xla_executable_build_options=merge(
                     (;
--- a/src/xla/CompileOptions.jl
+++ b/src/xla/CompileOptions.jl
@@ -35,13 +35,24 @@ function get_default_compile_options()
     return proto
 end
 
-function get_debug_options(; kwargs...)
+function get_debug_options(; platform::String="", kwargs...)
     debug_options = get_default_debug_options()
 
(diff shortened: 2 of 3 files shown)


wsmoses:
Upstream issue filed: openxla/xla#46952 — once resolved there, reverting this default is a one-liner (drop the `platform == "rocm"` block in `get_debug_options`).

PR #3186, opened 2026-08-14
add `traced_type_inner` method for `Union{A,B}`

### LINKED ISSUE #3185: Tracing structs with Union type bounds
Here's a bug I think I ran into while trying to apply `Reactant.to_rarray` to objects from [CoolPDLP.jl](https://github.com/JuliaDecisionFocusedLearning/CoolPDLP.jl). I think this can be solved on your end?

MWE found and commented by Claude below (I ran it and checked that it errors as planned):

```julia
using Reactant

const BatchedNumber{T <: Number} = Union{T, AbstractVector{T}}

struct Scratch{T <: Number, V <: AbstractVecOrMat{T}, S <: BatchedNumber{T}}
    x::V
    b::S
end

scratch = Scratch(rand(Float32, 3), 1.0f0)

# Works: without `track_numbers`, the scalar field `b` is left untouched, so Reactant
# never needs to reconstruct `Scratch`'s type with a new `S`.
Reactant.to_rarray(scratch)

# Fails: `track_numbers = true` also traces the scalar field `b`, forcing Reactant to
# reconstruct `Scratch{T′, V′, S′}` for the new traced types. While resolving the
# `Union`-bound type parameters `V` and `S`, `Reactant.typevar_dict` calls `.parameters`
# on a `Union` object, which doesn't have that field (only `DataType`/`UnionAll` do):
#
#   ERROR: FieldError: type Union has no field `parameters`, available fields: `a`, `b`
#   Stacktrace:
#     [1] getproperty(x::Type, f::Symbol)
#     [2] typevar_dict(t::Type)
#         @ Reactant Tracing.jl:1021
#     [3] apply_type_with_promotion(wrapper::Type, params::Vector{Any}, relevant_typevars::Dict{Any, Any})
#         @ Reactant Tracing.jl:1004
#     [4] apply_type_with_promotion(wrapper::Type, params::Vector{Any})
#         @ Reactant Tracing.jl:959
#     [5] traced_type_inner(...)
#         @ Reactant Tracing.jl:790
#     [6] traced_type(...)
#         @ Reactant Tracing.jl:1043
#     ...
Reactant.to_rarray(scratch; track_numbers = true)
```

mofeing:
fixes #3185 (well, still checking out)

cc @gdalle

### DIFF
--- a/src/Tracing.jl
+++ b/src/Tracing.jl
@@ -27,6 +27,12 @@ Base.@nospecializeinfer function traced_type_inner(
     return T
 end
 
+Base.@nospecializeinfer function traced_type_inner(@nospecialize(_::Type{Union{A,B}}), @nospecialize(args...)) where {A,B}
+    tA = traced_type_inner(A, args...)
+    tB = traced_type_inner(B, args...)
+    return Union{tA,tB}
+end
+
 for T in (
     DataType,
     Module,


glou-nes:
Got infinite precompilation (and no print at all), even without the recursive part. For instance, with this in 1.12:
```julia
Base.@nospecializeinfer function traced_type_inner(
    @nospecialize(_::Type{Union{A,B}}), @nospecialize(args...)
) where {A,B}
    ccall(:jl_, Cvoid, (Any,), name * " " * string(stacktrace()))
    return Nothing
end
```

PR #3198, opened 2026-08-19
[RFC] basic support for the sparse tensor dialect

simeonschaub:
Introduces `Reactant.CSRMatrix`, hooks it up to https://mlir.llvm.org/docs/Dialects/SparseTensorOps/ and adds the ability to lower simple spmv and spmm ops to either cuSPARSE or hipSPARSE. I verified the cuSPARSE path locally.

Putting this up for comments, since I'm not sure this is the right approach, but this would be really useful to us in CoolPDLP.jl!


Pangoraw:
Hey Simeon! Thanks for the PR. Would it be possible to port the lowerings from this PR in [Enzyme-JAX](https://github.com/EnzymeAD/Enzyme-JAX). That way we can also include other missing pieces such as AD rules for the sparse ops. You can create a subdialect `enzymexla.sparse` instead of matching into `sparse.assemble`+`stablehlo.dot_general` if that's more convenient.

simeonschaub:
Sure, I can give that a go!

PR #3201, opened 2026-08-20
basic support for sparse tensors (2)

simeonschaub:
Companion to EnzymeAD/Enzyme-JAX#2922. Replaces #3198 and moves the lowering to Enzyme-JAX, which gains a `sparse.spmm` op. I also added support for emitting a single cuSPARSE/hipSPARSE call, even if `mul!` is called with different alpha and beta (they have to be constant)


src/compiler/Compiler.jl
```diff
@@ -430,6 +447,13 @@ function compile_mlir!(
 
     legal_to_run_shardy_passes = compile_options.optimization_passes === :all
 
+    # Lower sparse_tensor CSR products to library custom calls before anything
+    # else runs: XLA cannot consume sparse-encoded tensor types. With `:none`
+    # the sparse IR is kept as-is for inspection.
+    if compile_options.optimization_passes !== :none && has_sparse_tensor_ops(mod)
+        run_pass_pipeline!(mod, "lower-sparse-csr", "lower_sparse_csr")
+    end
```
Pangoraw: add as part of the existing pass-pipeline

PR #3202, opened 2026-08-20
refactor MPI integration

mofeing:
this pr introduces the following changes:

- MPI bindings now emit `comm.mpi` MLIR ops. EnzymeXLA MPI ops are deprecated and will be deleted soon.
- support for tracing over MPI communicators
- communicators and request can now cross the compile boundary (i.e. you can pass them as args or return them from a Reactant-compiled function)
- it can optionally overlay MPI calls with no Reactant type involved and emit them as MLIR ops with constants 
- allows for both lowering to FFI handlers and JIT calls
- datatype is automatically handled by EnzymeXLA

# to do

- [ ] update and check tests
- [x] add compiler passes to the pipeline, with options


ext/ReactantMPIExt/Ops.jl
```diff
@@ -1,268 +1,216 @@
...
-    MPI.MIN.val => MLIR.API.ENZYMEXLA_MPI_MIN,
-    MPI.PROD.val => MLIR.API.ENZYMEXLA_MPI_PROD,
-    MPI.REPLACE.val => MLIR.API.ENZYMEXLA_MPI_REPLACE,
-    MPI.SUM.val => MLIR.API.ENZYMEXLA_MPI_SUM,
-    MPI.NO_OP.val => MLIR.API.ENZYMEXLA_MPI_NO_OP,
-)
-
-function get_mpi_op_enum(op)
-    return get(MPI_OP_MAP, op.val) do
-        throw(ArgumentError("Unknown MPI op `$op`"))
```
romanlee, line 228: why'd you get rid of my nice error messages? :( 

Also FYI in case you run into the same issue, the `.val` thing was needed to pass CI on some systems

mofeing: > why'd you get rid of my nice error messages? :(

Ah, because it will already error the same way but with a similar error by just doing `MPI_OP_MAP[mpi_op]`. For example,

```julia
julia> d = Dict([:a => 1, :b => 2])
Dict{Symbol, Int64} with 2 entries:
  :a => 1
  :b => 2

julia> d[:c]
ERROR: KeyError: key :c not found
Stacktrace:
 [1] getindex(h::Dict{Symbol, Int64}, key::Symbol)
   @ Base ./dict.jl:477
 [2] top-level scope
   @ REPL[2]:1
```

I can put it back.

> Also FYI in case you run into the same issue, the .val thing was needed to pass CI on some systems

~alright, that seems a lil bit weird knowing that the ops are compile time constants~

okay, it looks like `MPI.Op` also saves a function pointer for custom reductions and there is some load-time hooks that may be interfering. https://github.com/JuliaParallel/MPI.jl/blob/fa4013bfbb66bf80acc9fd40e380492e1aaae8b7/src/operators.jl#L21-L56

mofeing: actually, I put it back so it also handles the `.val` thing and the error messages.

PR #3212, opened 2026-08-24
Make speculate_partial_ifs a compile option

Pangoraw:
NFC since default is false as the pass default. I think this can speed compile and run times on *raised* programs.


wsmoses (approved):

wsmoses:
long term we should still fix this on the ejax side to have optimizations which mean this isn't necessary, since otherwise we hit the same nan potential problems in raised derivatives

PR #3213, opened 2026-08-24
lazy initialize path to libReactantExtra for relocability

### LINKED ISSUE #3190: Reactant is not relocatable due to `MLIR.API.mlir_c`
`MLIR.API.mlir_c` captures the artifact path at precompile time, breaking depot relocation

If a depot is relocated after precompilation, `using Reactant` fails: `__init__` tries to `dlopen` the library at the old path.
```
julia> using Reactant
ERROR: InitError: could not load library "/private/var/folders/df/.../T/jl_nm5ERo/Jumbo.app/.../libReactantExtra.dylib"
Stacktrace:
  [1] __init__()
    @ Reactant.XLA .../Reactant/src/xla/XLA.jl:203
during initialization of module XLA
```
The `jl_nm5ERo` prefix is the temp dir the depot was precompiled in; it was loaded from `/Applications/Jumbo.app`. The JLL itself relocates fine — in the same session `Reactant_jll.libReactantExtra_path` returns the correct `/Applications/...` path, and the artifact is present there.

The stale value comes from `src/mlir/MLIR.jl:11`:
```julia
    Reactant_jll.libReactantExtra
else
    missing
end
```
Being a top-level `const`, this is evaluated at precompile time and serialized into the pkgimage, so it isn't refreshed by `Reactant_jll.__init__`. `XLA.__init__` uses it in the first statement touching the library (`XLA.jl:203`, `cglobal((:ReactantThrowError, MLIR.API.mlir_c), ...)`), which fails immediately. Every other call site goes through `Reactant_jll.libReactantExtra_handle`, which is refreshed correctly — `mlir_c` is the only stale reference.

Resolving it lazily instead of at precompile time should fix it.

mofeing:
@JanisErdmanis mind trying whether this PR fixes your issue?


Pangoraw (approved):

mofeing:
switching to draft so that we don't merge before @JanisErdmanis confirms whether this fixes #3190 or not

also, it seems to be breaking Windows builds 😫

giordano:
What's the issue this is trying to solve?

mofeing:
I have no experience on building pkgimages or relocability, but apparently Reactant is not relocatable as of #3190. @JanisErdmanis gave a talk at JuliaCon this year about his Julia "Jumbo distribution" were he addressed this.

specifically, the issue seems to be the `mlir_c` global const that we use for referring to libReactantExtra path.

JanisErdmanis:
The fix seems proper. I rebuilt the Jumbo Julia distribution with this Reactant branch this evening, and it has resolved #3190. I have no idea whether there are other issues, since running `test Reactant` recompiles the pkgimage cache and hence is unusable for relocatability tests. For now we know that loading works on macOS, and by tomorrow the Windows build should be finished, so I'll be able to test that as well.

JanisErdmanis:
I can also report that the Windows build succeeded and resolved the loading issue #3190. I asked Claude to put together an example, and the following works with the Jumbo Julia distribution:
```
using Reactant

# Convert regular Julia arrays into Reactant's array type
input1 = Reactant.to_rarray(ones(10))
input2 = Reactant.to_rarray(ones(10))

# A plain Julia function — nothing special needed
function sinsum_add(x, y)
    return sum(sin.(x) .+ y)
end

# Compile it ahead-of-time
compiled_f = @compile sinsum_add(input1, input2)

# Run the compiled version
result = compiled_f(input1, input2)
```

mofeing:
closed in favour of #3248

PR #3223, opened 2026-08-26
Reserve an address range per XLA allocation

wsmoses:
The handle `reactantXLAMalloc` returned was an 8-byte heap slot, which made pointer arithmetic on device handles unsound: interior pointers (`x.Read() + offset`, ubiquitous in MFEM) walked past the slot into neighbors, foreign pointers resolved to a garbage (buffer, ~30TB offset) pair, and a stray host write through a device handle silently replaced the stored `PjRtBuffer*` with data (observed: the bit pattern of -0.5, crashing much later in `CopyToBuffer`).

Reserve an inaccessible `mmap(PROT_NONE, MAP_NORESERVE)` range as large as the logical buffer instead, tracked as base → {buffer, size}: interior pointers resolve to (buffer, offset) with a bounds check, foreign pointers get a precise "does not belong to any reactant allocation" report, frees unmap the range, and any host dereference of a device handle faults at the offending instruction instead of corrupting state. Exec's zero-offset diagnostic now prints the pointer and resolved base.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -1,3 +1,4 @@
+#include <sys/mman.h>
 #include <iostream>
 
 #include "mlir-c/IR.h"
@@ -3428,8 +3429,14 @@ struct LinkableRuntime {
                                   xla::PjRtLoadedExecutable *>>
       executables;
 
-  // Set of allocated pointers to size
-  std::set<void *, std::greater<void *>> allocations;
+  // Each allocation reserves an inaccessible address range as large as the
+  // buffer it stands for, so pointer arithmetic on the handle stays inside
+  // its own range (and a stray dereference faults at the offender).
+  struct AllocationInfo {
+    xla::PjRtBuffer *buffer;
+    size_t size;
+  };
+  std::map<void *, AllocationInfo, std::greater<void *>> allocations;
 
   LinkableRuntime(const std::string &backend) : registry() {
     InitializeRegistry(wrap(&registry));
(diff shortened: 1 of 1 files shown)


wsmoses:
Pushed two more linkable-runtime fixes from the MFEM campaign:
- **DtoD copies**: mfem's `Vector` copy constructor issues `cudaMemcpy(DeviceToDevice)`, which fell into an `llvm_unreachable` arm (UB in release — it fell through into the HtoD path and faulted in a stream host callback). PJRT exposes no raw buffer-to-buffer copy, so it now stages through a host bounce buffer.
- **Null kernel arguments**: kernels capture optional buffers as null pointers (`geom ? geom->J.Read() : nullptr`) and guard every access on a flag; the runtime now hands the executable a small fresh dummy buffer for those arguments (fresh per use, since donated arguments cannot be reused within a call).

wsmoses:
Pushed two more runtime fixes found while running the MFEM CUDA suite end-to-end:

- **Keep the written view's result for aliased buffers** (`fff8c811b`): a kernel whose output aliases an input (mfem's `add(z, beta, d, d)` inside CGSolver) resolves to the same buffer twice; keeping the first view's result silently kept the stale input whenever the written view came later, so in-place updates vanished and preconditioned CG stalled at its initial residual. The executable now records per-view writtenness (a view is read-only when its result resolves back to its own argument through layout ops and untouched while iter-args) and each duplicate group keeps the written view's result.
- **Retry the pipeline without slice-motion on invalid IR** (`1d89255cd`): an EnzymeHLOOpt pattern interplay rooted in the slice-of-broadcast motion group can leave type-inconsistent IR; the post-pipeline verify catches it and the executable rebuilds once with that pattern group disabled instead of dying. Also adds debug plumbing (`REACTANT_XLA_DEBUG_EXEC` argument tracing, failing-module dumps).

wsmoses:
Split into a stacked series, one PR per feature, in this order (each based on the previous; the combined tip is byte-identical to this branch's `API.cpp`):

1. #3235 — Reserve an address range per XLA allocation
2. #3236 — Stage device-to-device copies through the host
3. #3237 — Hand null kernel arguments a dummy buffer
4. #3238 — Keep the written view's result for aliased buffers
5. #3239 — Exec-pipeline retries on invalid IR (slice-motion off, then no hlo-opt) + `REACTANT_XLA_NO_HLO_OPT`
6. #3240 — Stage offset argument views
7. #3241 — Environment-gated debugging switches (`REACTANT_XLA_DEBUG_EXEC` / `REACTANT_XLA_DUMP_DIR` / `REACTANT_XLA_DUMP_MOD_ON_ERR`)

Every intermediate stage compiles standalone (`ReactantExtraLib`, cuda config). Closing in favor of the series.

PR #3224, opened 2026-08-26
Let Dates arithmetic work when the operand is traced

Pangoraw:
Three gaps stop ordinary `Dates` code from working once one of its operands becomes a `TracedRNumber`, even though `ReactantDatesExt` already provides a full set of counterpart types and accessors.

### 1. Period constructors

`Dates.Second`, `Dates.Millisecond` and friends store a concrete `Int64`, so they cannot hold a traced value:

```julia
julia> Dates.Second(x::TracedRNumber{Int64})
ERROR: MethodError: no method matching Int64(::TracedRNumber{Int64})
```

Constructing a period from a traced number now yields the Reactant counterpart. This keeps *generic* code working unchanged — code that writes `Millisecond(round(Int, 1000t))` without knowing whether `t` is traced.

### 2. Concrete `TimeType` with a traced period

The extension already defines `ReactantDateTime ± Period` (traced datetime, possibly-concrete period). The mirror case was missing: a wall-clock epoch advanced by a *traced* offset. Base's `+` builds a `UTInstant{ReactantMillisecond{TracedRNumber}}` and then tries to pour it into a plain `DateTime`, whose instant must hold a concrete `Int64`:

```
MethodError: no method matching Int64(::UTInstant{ReactantMillisecond{TracedRNumber{Int64}}})
```

Promoting the datetime to its Reactant counterpart first fixes it, for `DateTime`, `Date` and `Time`.

### 3. `Adapt` did not recurse into the date types

Kernel arguments are walked by `Adapt`, which swaps `TracedRNumber` for the device-side `CuTracedRNumber` — but only for types that say how. The date types wrap exactly one traced number and had no `adapt_structure`, so that number reached the kernel still host-side. Every operation on it then tried to *emit MLIR* instead of computing, dragging Reactant's own tracing machinery into device code:

```
Reason: unsupported dynamic function invocation (call to Reactant.MLIR.IR.Location)
  @ Reactant/src/Ops.jl:153
```

This is the situation Reactant's own kernel-argument checker already diagnoses and prescribes a fix for. It can't be a plain `Adapt.@adapt_structure`, because the traced number isn't a direct field — it sits behind `UTInstant` and the period wrapper — so the value is adapted and the wrappers rebuilt around it, mirroring what `ReactantCUDAExt` does for `TracedStepRangeLen` and `Base.TwicePrecision`.

Written against a generic adaptor rather than `ReactantKernelAdaptor`, so the extension stays free of a CUDA dependency.

## Motivation

Found while making an atmosphere model's radiation compute a real solar zenith angle from a *traced* model clock instead of a compile-time constant. With these three changes the caller's seconds-to-datetime conversion and its per-column KernelAbstractions kernel both work with **no change on the caller's side** — the downstream `cos_solar_zenith_angle(datetime, λ, φ)` kernel compiles and produces per-column values matching a host computation exactly.

## Note for reviewers

The judgement call here is item 1: `Dates.Millisecond(::TracedRNumber)` returns a `ReactantMillisecond`, so it is a type-changing constructor. It seemed the only representable answer, and it is what lets untouched generic code keep working — but it is the piece most worth a second opinion.

## Tests

`test/integration/dates.jl` gains three testsets covering the period constructors, concrete-plus-traced arithmetic in both directions and both operand orders, and that `Adapt` reaches the wrapped payload (via a marker adaptor, so it needs no GPU).

Locally: **639 passed, 0 failed** for `test/integration/dates.jl` (635 pre-existing + 4 new), on CPU.

https://claude.ai/code/session_01LShNSUDBfvosAH2onkcq3q

### DIFF
--- a/ext/ReactantDatesExt/adapt.jl
+++ b/ext/ReactantDatesExt/adapt.jl
@@ -1,30 +1,3 @@
-# Adapt support, so Reactant date/time types can be passed as GPU KERNEL ARGUMENTS.
-#
-# Kernel arguments are walked by `Adapt`, which replaces `TracedRNumber`/`TracedRArray` with their
-# device-side `CuTracedRNumber`/`CuTracedArray` counterparts. `Adapt` only recurses into types that
-# say how; a type with no method is treated as opaque and its fields are left alone.
-#
-# `ReactantDateTime` and friends wrap exactly one traced number, so without these methods that number
-# survives into the kernel still HOST-side. Every operation on it then tries to emit MLIR instead of
-# computing, and Reactant's own tracing machinery is dragged into device code:
-#
-#     Reason: unsupported dynamic function invocation (call to Reactant.MLIR.IR.Location)
-#       @ Reactant/src/Ops.jl:153
-#
-# Reactant's own kernel-argument checker diagnoses this and prescribes the fix:
-#
-#     GPU kernel argument of type $T contains an unadapted traced value at field: $bad
-#     … some struct in the hierarchy is missing `Adapt.@adapt_structure`, so its fields were not
-#     recursed into during GPU adaptation.
-#
-# These cannot be a plain `Adapt.@adapt_structure`, because the traced number is not a direct field:
-# it sits behind `UTInstant` and the period wrapper (`ReactantDateTime.instant.periods.value`), so
-# the value has to be adapted and the wrappers rebuilt around it. This mirrors what Reactant already
-# does for `TracedStepRangeLen` and `Base.TwicePrecision` in ReactantCUDAExt.
-#
-# Written against a generic adaptor rather than the CUDA one: adapting a wrapper by adapting its
-# payload is correct for ANY adaptor, and it keeps this extension free of a CUDA dependency.
-
 function Adapt.adapt_structure(to, dt::ReactantDateTime)
     return ReactantDateTime(UTInstant(ReactantMillisecond(Adapt.adapt(to, value(dt)))))
 end
--- a/ext/ReactantDatesExt/adapt.jl
+++ b/ext/ReactantDatesExt/adapt.jl
@@ -1,0 +1,43 @@
+# Adapt support, so Reactant date/time types can be passed as GPU KERNEL ARGUMENTS.
+#
+# Kernel arguments are walked by `Adapt`, which replaces `TracedRNumber`/`TracedRArray` with their
+# device-side `CuTracedRNumber`/`CuTracedArray` counterparts. `Adapt` only recurses into types that
+# say how; a type with no method is treated as opaque and its fields are left alone.
+#
+# `ReactantDateTime` and friends wrap exactly one traced number, so without these methods that number
+# survives into the kernel still HOST-side. Every operation on it then tries to emit MLIR instead of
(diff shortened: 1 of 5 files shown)


ext/ReactantDatesExt/adapt.jl
```diff
@@ -0,0 +1,16 @@
+function Adapt.adapt_structure(to, dt::ReactantDateTime)
+    return ReactantDateTime(UTInstant(ReactantMillisecond(Adapt.adapt(to, value(dt)))))
```
maximilian-gelbrecht, line 2: Why not just do a simple adapt as below for the period types or is adapt not defined for `UTInstant`?

ext/ReactantDatesExt/arithmetic.jl
```diff
@@ -156,3 +156,42 @@ end
...
+# for a traced elapsed time `t`.
+#
+# Without these, Base's `+(::DateTime, ::Period)` builds a
+# `UTInstant{ReactantMillisecond{TracedRNumber}}` and then tries to pour it into a plain `DateTime`,
+# whose instant must hold a concrete `Int64`:
+#
+#     MethodError: no method matching Int64(::UTInstant{ReactantMillisecond{TracedRNumber{Int64}}})
+#
+# Promoting the datetime to its Reactant counterpart first is the only representable answer.
+const ReactantDatePeriod = Union{
```
maximilian-gelbrecht, line 174: Minor coding style choice: I would always move these `const` definitions to the beginning of the file or even the module definition file, but that's just my opinion.

ext/ReactantDatesExt/conversions.jl
```diff
@@ -24,6 +27,22 @@ for (S, T) in (
...
+#
+#     julia> Dates.Second(x::TracedRNumber{Int64})
+#     ERROR: MethodError: no method matching Int64(::TracedRNumber{Int64})
+#
+# Without these methods, ordinary date arithmetic written against `Dates` fails the moment its
+# operand becomes traced, even though every Reactant counterpart type already exists. Returning the
+# Reactant period is the only representable answer, and it keeps generic code — code that says
+# `Millisecond(round(Int, 1000t))` without knowing whether `t` is traced — working unchanged.
+for (S, T) in _PERIOD_PAIRS
+    @eval Dates.$S(x::Reactant.TracedRNumber) = $T(x)
```
maximilian-gelbrecht: This does somewhat violate Julia standards, right? I thought about adding that, but didn't. But I would be fine with it and it also seems the most pragmatic choice.

maximilian-gelbrecht (commented):
looks good to me, just some very minor comments

ext/ReactantDatesExt/conversions.jl
Pangoraw: Right, I don't think this qualify strictly as type piracy since Reactant owns TracedRNumber but I removed it since it did not have a motivating use case.

ext/ReactantDatesExt/adapt.jl
Pangoraw: I don't think UTInstant has an adapt method. Should it be defined in Adapt.jl?

maximilian-gelbrecht: Probably yes, in Adapt.jl

PR #3227, opened 2026-08-26
Keep the batching rewrites out of the post-optimization cleanup

### LINKED ISSUE #3225: `:all` pipeline never terminates on bf16-compute graphs: a three-pattern rewrite cycle in the post-optimization cleanup
## Summary

`@compile` never returns on any model that keeps fp32 master weights and computes in bf16,
once the model has two or more transformer layers. Not slow -- unbounded: the
post-optimization cleanup stage of the `:all` pipeline runs a rewrite cycle that grows the
module on every round.

I hit this porting a small GPT (rotary, QK-norm, GQA, sliding-window masked softmax,
ReLU^2, value embeddings) to Reactant. Symptoms:

- `@compile Enzyme.gradient(...)` of the bf16 model: **no output after 45 minutes**, on CPU
  and on CUDA alike. Depth 1 compiles in 91 s; depth 2 and 3 never finish.
- The fp32 version of the same graph compiles fine at any depth.
- Not Enzyme-specific: a hand-written backward pass (plain array ops, no `enzyme.autodiff`
  in the module) hangs identically -- >25 min at depth 4, where fp32 takes 156 s.
- `optimize=:only_enzyme` and `optimize=:before_enzyme` both compile the same graph in ~95 s
  and give correct gradients, which narrows it to the passes `:all` adds.
- Individual primitives (matmul+casts, rmsnorm, ReLU^2 MLP, masked-softmax attention,
  softcap) each compile in seconds.

## Root cause

Only `:all` runs `post_op_transpose_reshape`: the full pattern set re-run with
reshape/transpose propagation **reversed** (`down`, then `up`, then `down`). Replaying the
traced module stage by stage with `Reactant.Compiler.run_pass_pipeline_on_source` and the
strings from `Reactant.Compiler.optimization_passes` shows every stage finishing in 0-2 s
except the first `down` pass, which never returns.

gdb on the hung process lands in `ConcatInsertDimToBatchBase::matchAndRewriteImpl`
(`AutoBatching.cpp`) creating yet another `enzymexla_unbatched_ConcatInsertDimToBatch_*`
wrapper function inside `applyPatternsGreedily` -- the module keeps growing rather than
looping tightly between two patterns.

An exclusion bisect (`CompileOptions(excluded_passes=...)`, one subprocess per trial under
`timeout`) pins it down: excluding `concat_insert_dim_elementwise` alone brings the stage to
0.2 s; excluding all 21 down-propagation-only patterns, 0.4 s; excluding any *single* down
pattern, or the whole reshape-down group, or the whole transpose-down group, still hangs.

Reduced (ddmin over the 471-pattern down pipeline, then cut by hand) it is exactly three
patterns on a six-line function:

```
start:                            convert(concat(reshape(a), reshape(b)))
convert_concat                 -> concat(convert(reshape(a)), convert(reshape(b)))
elementwise_reshape_like       -> concat(reshape(convert(a)), reshape(convert(b)))
concat_insert_dim_elementwise  -> convert(concat(reshape(a), reshape(b)))   == start
```

with a new unbatched wrapper function per round.
(`reorder_elementwise_and_shape_op` is `elementwise_reshape_like`'s counterpart in the
transpose-down group, so that side loops the same way.)

## Reproducer

No model needed -- this is the cleanup pipeline on its own:

```julia
using Reactant

source = """
module {
  func.func @main(%a: tensor<16x64x2xf32>, %b: tensor<16x64x2xf32>) -> tensor<2x16x64x2xbf16> {
    %ra = stablehlo.reshape %a : (tensor<16x64x2xf32>) -> tensor<1x16x64x2xf32>
    %rb = stablehlo.reshape %b : (tensor<16x64x2xf32>) -> tensor<1x16x64x2xf32>
    %c = stablehlo.concatenate %ra, %rb, dim = 0 : (tensor<1x16x64x2xf32>, tensor<1x16x64x2xf32>) -> tensor<2x16x64x2xf32>
    %cv = stablehlo.convert %c : (tensor<2x16x64x2xf32>) -> tensor<2x16x64x2xbf16>
    return %cv : tensor<2x16x64x2xbf16>
  }
}
"""

patterns = "concat_insert_dim_elementwise;elementwise_reshape_like;convert_concat"
pipeline = "enzyme-hlo-generate-td{patterns=$(patterns)},transform-interpreter,enzyme-hlo-remove-transform"
Reactant.Compiler.run_pass_pipeline_on_source(source, pipeline)   # never returns
```

No output after 10 minutes; RSS climbs steadily while it spins (1.33 GB -> 1.70 GB over
three minutes). All three patterns are needed -- drop any one and it converges immediately.
A rotary embedding is enough to put this shape in a real graph: `concat` of two reshaped
halves, then the bf16 cast.

## Fix

PR to follow: run the post-optimization cleanup with the concat/slice-to-batch groups
disabled (`disable_slice_to_batch_passes` / `disable_concat_to_batch_passes`). Batching has
already been applied by the main pipeline at that point, so nothing is lost. Measured after
the change: the depth-2 bf16 Enzyme gradient compiles under `:all` in 96.5 s (was >45 min),
the depth-4 hand-written backward in 7.5 s (was >25 min), compiled-vs-eager bf16 loss
difference 9.5e-7. The reproducer above returns in 0.14 s once the batching patterns are out
of the run.

The rewrite cycle itself is an Enzyme-JAX bug and is filed separately with an
`enzymexlamlir-opt` reproducer: EnzymeAD/Enzyme-JAX#2983. The Reactant-side change is
still worth having on its own -- it also covers anyone using `reshape_propagate=:down` in
the main pipeline, and works on today's jll.

## Environment

Reactant.jl main (v0.2.283 source) with `Reactant_jll` v0.0.405, Julia 1.12.7, Linux x86_64.
Reproduces on the CPU backend and on CUDA (H100, driver 580.105, CUDA 13.1 runtime).

## Possibly related

- #2652 -- `GreedyPatternRewriteDriver` non-termination in `ReduceMulToDotGeneral`
- #3195 -- reshaped traced matrix * untraced matrix, compile-time blow-up
- #2805 -- `ConvertMulConvert` correctness error, on the same convert/elementwise chains

AsafManela:
Fixes #3225.

`compile_mlir!` finishes the `:all` pipeline with a cleanup stage that re-runs the whole
pattern set with transpose/reshape propagation reversed (`down`, `up`, `down`). In that
configuration three EnzymeXLA patterns rewrite each other in a cycle:

```
start:                            convert(concat(reshape(a), reshape(b)))
convert_concat                 -> concat(convert(reshape(a)), convert(reshape(b)))
elementwise_reshape_like       -> concat(reshape(convert(a)), reshape(convert(b)))
concat_insert_dim_elementwise  -> convert(concat(reshape(a), reshape(b)))   == start
```

and each trip through the batching rewrite adds another unbatched wrapper function, so the
greedy driver never reaches a fixed point and the module grows without bound. Any graph with
fp32 master weights and bf16 compute hits this -- a rotary embedding's `concat(reshape(...))`
feeding a bf16 cast is enough -- and `@compile` then runs for hours with no output.
Individual primitives compile in seconds; two or more transformer layers hang. The issue has
the full bisect and the gdb trace.

Batching has already been applied by the main pipeline before this stage, so run the cleanup
with the concat/slice-to-batch groups disabled.

### Results

- Depth-2 bf16 Enzyme gradient of a small GPT: compiles under `:all` in **96.5 s**, was
  >45 min with no output.
- Depth-4 hand-written backward pass (no `enzyme.autodiff` in the module -- this is not an
  Enzyme problem): **7.5 s**, was >25 min. Compiled-vs-eager bf16 loss difference 9.5e-7.
- The pass-pipeline reproducer from the issue returns in **0.14 s**.

### Tests

`test/core/optimization_passes.jl`: the cleanup pipeline strings must not contain the
batching patterns while the main pipeline still does, and the minimal
`convert(concat(reshape))` module must survive `down, up, down` (10/10 locally). Note that
before this change the second testset does not fail, it hangs.

### Relationship to the Enzyme-JAX fix

The rewrite cycle itself is an Enzyme-JAX bug, filed and fixed separately
(EnzymeAD/Enzyme-JAX#2983, EnzymeAD/Enzyme-JAX#2988: `ConvertConcat` declines the batchable
shape). This change is still worth having on its own -- it works on today's jll, and it also
covers anyone passing `reshape_propagate=:down` to the main pipeline, where the same three
patterns meet.

Verified on Julia 1.12.7 / `Reactant_jll` v0.0.405, CPU backend and CUDA (H100).

https://claude.ai/code/session_012k6KZhhCNSCvZFKzErHaHH

### DIFF
--- a/src/CompileOptions.jl
+++ b/src/CompileOptions.jl
@@ -427,6 +427,53 @@ function __compile_options_with_reversed_propagation(compile_options::CompileOpt
     )
 end
 
+# Copy of `compile_options` with the concat/slice-to-batch rewrites turned off. Used for
+# the post-optimization transpose/reshape cleanup: under reversed propagation
+# `concat_insert_dim_elementwise` and the reshape-propagation patterns rewrite each other
+# in a cycle and the greedy rewriter never terminates, and batching has already been
+# applied by the main pipeline anyway.
+function __compile_options_without_batching_passes(compile_options::CompileOptions)
+    return CompileOptions(
+        compile_options.optimization_passes,
+        compile_options.no_nan,
+        compile_options.all_finite,
+        compile_options.inline,
+        compile_options.transpose_propagate,
+        compile_options.reshape_propagate,
+        compile_options.max_constant_threshold,
+        compile_options.raise,
+        compile_options.raise_first,
+        compile_options.legalize_chlo_to_stablehlo,
+        compile_options.cudnn_hlo_optimize,
+        compile_options.shardy_passes,
+        compile_options.optimize_then_pad,
+        compile_options.optimize_communications,
+        compile_options.raise_triton_custom_call,
+        compile_options.lower_triton,
+        compile_options.assert_nonallocating,
+        compile_options.donated_args,
+        compile_options.sync,
+        compile_options.xla_executable_build_options,
+        compile_options.xla_compile_options,
+        compile_options.xla_debug_options,
+        compile_options.disable_scatter_gather_optimization_passes,
+        compile_options.disable_pad_optimization_passes,
+        compile_options.disable_licm_optimization_passes,
+        compile_options.loop_unswitch_threshold,
+        compile_options.excluded_passes,
+        compile_options.disable_reduce_slice_fusion_passes,
+        true, # disable_slice_to_batch_passes
+        true, # disable_concat_to_batch_passes
+        compile_options.disable_loop_raising_passes,
+        compile_options.disable_structured_tensors_detection_passes,
+        compile_options.disable_structured_tensors_passes,
+        compile_options.strip_llvm_debuginfo,
+        compile_options.speculate_partial_ifs,
+        compile_options.strip,
+        compile_options.multifloat,
+    )
+end
+
 function __compile_options_with_updated_sync(compile_options::CompileOptions, sync::Bool)
     if compile_options.sync == sync
         return compile_options
--- a/src/compiler/Compiler.jl
+++ b/src/compiler/Compiler.jl
@@ -739,11 +739,12 @@ function compile_mlir!(
             is_sharded,
             raise_shlo_to_blas_lapack=false,
         )
+        cleanup_opts = Reactant.__compile_options_without_batching_passes(compile_options)
         opt_passes_down = optimization_passes(
-            Reactant.__compile_options_with_reversed_propagation(compile_options);
+            Reactant.__compile_options_with_reversed_propagation(cleanup_opts);
             common_kwargs...,
         )
-        opt_passes_up = optimization_passes(compile_options; common_kwargs...)
+        opt_passes_up = optimization_passes(cleanup_opts; common_kwargs...)
         run_pass_pipeline!(
             mod,
             join([opt_passes_down, opt_passes_up, opt_passes_down], ","),
--- a/test/core/optimization_passes.jl
+++ b/test/core/optimization_passes.jl
@@ -1,0 +1,65 @@
+using Reactant, Test
+
+# The `:all` pipeline finishes with a cleanup stage that re-runs the pattern set with
+# transpose/reshape propagation reversed. The concat/slice-to-batch rewrites have to stay
+# out of that stage: with propagation reversed, `concat_insert_dim_elementwise` and the
+# reshape-propagation patterns rewrite each other in a cycle and the greedy driver never
+# reaches a fixed point (layered bf16 graphs hung `@compile` indefinitely). Batching has
+# already been applied by the main pipeline at that point.
+
+function cleanup_pipelines(compile_options)
+    cleanup = Reactant.__compile_options_without_batching_passes(compile_options)
+    down = Reactant.Compiler.optimization_passes(
+        Reactant.__compile_options_with_reversed_propagation(cleanup);
+        backend="cpu",
+        raise_shlo_to_blas_lapack=false,
+    )
+    up = Reactant.Compiler.optimization_passes(
+        cleanup; backend="cpu", raise_shlo_to_blas_lapack=false
+    )
+    return down, up
+end
+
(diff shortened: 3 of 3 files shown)


AsafManela:
Closing this: I built the fix and it turns out this PR is unnecessary and not free.

The root cause is the Enzyme-JAX rewrite cycle in #3225 / EnzymeAD/Enzyme-JAX#2983. I have
now compiled that fix (EnzymeAD/Enzyme-JAX#2988) into a JLL and measured it against an
otherwise identical build. On a real optimization-stage module that hangs today, with the
**full** pattern set -- i.e. with the batching rewrites this PR disables still enabled:

```
without the Enzyme-JAX fix   HANG (killed at 300 s)
with the Enzyme-JAX fix      OK, 0.7 s
```

So once the cycle is fixed at its source, nothing needs to be excluded from the
post-optimization cleanup. This PR would permanently disable the concat/slice-to-batch
rewrites in that stage for every user, to work around a bug that no longer exists. Its
secondary justification -- that it also covers `reshape_propagate=:down` -- is moot for
the same reason: the cycle cannot form once `ConvertConcat` declines the batchable shape.

The measurements quoted in the original description stand and now belong to the
Enzyme-JAX fix instead. Two independent real-model confirmations, both bf16/fp8 training
graphs with a hand-written backward on a single H100:

- depth 24, bf16: killed after 12 min without the fix, 164 s with it
- depth 24, fp8:  killed after 23 min without the fix, 215 s with it
- depth 4, fp8:   the stalling stage alone, >300 s vs 0.4 s

One piece worth keeping: the regression test. Happy to open a small test-only PR adding
`test/core/optimization_passes.jl` -- the `convert(concat(reshape))` module driven through
the real cleanup pipeline, asserting it terminates -- once the JLL carries the Enzyme-JAX
fix. Say the word and I'll put that up.

PR #3228, opened 2026-08-26
Point cuDNN's runtime-compiled engines at the NVRTC bundled in Reactant_jll

### LINKED ISSUE #3226: cuDNN's runtime-compiled engines (fused attention) can never build a plan: NVRTC/builtins version mix in the bundle
## Summary

cuDNN's runtime-compiled engines -- fused attention / SDPA and runtime fusion, i.e. the
`__cudnn$fmhaSoftmax` custom call and anything the `--xla_gpu_enable_cudnn_fmha` rewriter
would produce -- can never build an execution plan through `Reactant_jll`. Every plan build
fails in NVRTC:

```
CUDNN_STATUS_INTERNAL_ERROR_COMPILATION_FAILED; Reason: Encountered runtime kernel
compilation failure at: compilationResult != NVRTC_SUCCESS
[cudnn_frontend] Error: No valid execution plans built.
```

The cause is an NVRTC/builtins version mix inside the static link, and there is a one-line
environment hook that works around it. Details below; PR to follow.

## Reproducer

A hand-emitted `stablehlo.custom_call @__cudnn$fmhaSoftmax` with JAX's JSON backend config
(the op `jax.nn.dot_product_attention(implementation="cudnn")` lowers to): BNTH layout,
bf16, causal, B=2, H=6, T=S=512, D=128, 3-D f32 softmax stats. XLA:GPU accepts it -- the
config parses, the cuDNN graph is built, the descriptors validate -- and then the plan build
fails as above. A self-contained version is in the PR as `test/integration/cudnn.jl`.

## What actually fails

- `libReactantExtra.so` statically links NVRTC *and* its builtins (`libnvrtc_static_*`
  symbols, `nvrtc-builtins.compute_{75..121}.c`), so this is not a missing `libnvrtc`.
- `strace -f -e openat` on the process: no `libnvrtc*` / `libnvptxcompiler*` is ever looked
  up. Only XLA's own `ptxas` children run. `LD_LIBRARY_PATH`, `CUDA_PATH` and `CUDA_HOME`
  cannot reach this code path at all.
- cuDNN 9.21's own error log (`CUDNN_LOGLEVEL_DBG=2`) puts the failure inside
  `rtk(kernelNumRunning)->compile(compilerFlags, this->useNvrtcSassPath, true)` -- the SDPA
  engine compiles straight to SASS -- but does not surface the NVRTC diagnostic.
- With gdb attached: cuDNN's `nvrtcHelper::load()` (libcudnn_graph) fills a function-pointer
  table for the NVRTC API. Absent an override it does **not** point at the bundle's public
  `nvrtc*` symbols (libnvrtc_static, `nvrtcVersion` -> 13.1) but at an NVRTC embedded in
  `libcudnn_graph_static` (renamed `libcudnn_graph_static_v9_<sha>` symbols; its
  `nvrtcVersion` -> 13.2, with its own `nvrtcGetErrorString` table).
- Breaking on that embedded `nvrtcCompileProgram` and calling the embedded
  `nvrtcGetProgramLog`:

  ```
  NVRTC_ERROR_INVALID_OPTION
  nvrtc: error: invalid value for --gpu-architecture (-arch)
  ```

  The bundle carries only CUDA 13.1's builtins (`__nvrtc_builtins_static__13_1_*`), i.e. a
  cuDNN-9.21-for-CUDA-13.2 / CUDA-13.1-NVRTC mix in one static link, and the embedded
  compiler cannot resolve the H100 architecture it is handed.

## The hook

`nvrtcHelper::load()` starts with `getenv("CUDNN_NVRTC_OVERRIDE_PATH")`. If it is set, cuDNN
dlopens that library and dlsyms `nvrtcCreateProgram`, `nvrtcCompileProgram`, ... out of it
(failing with `"' not found in NVRTC override library: '"`), checking only that the NVRTC
major version matches CUDART's. Pointing it at a self-consistent shared libnvrtc makes the
SDPA engine compile. CUDA.jl's `libnvrtc.so.13.0.88` and `.13.3.33` both work: the plan
builds, and the output matches a plain-attention reference (max abs diff 7.8e-3, bf16).

Worth having beyond correctness -- at B=2, H=6, T=S=512, D=128, bf16, causal, the cuDNN
fused forward runs in **0.43-0.49 ms** against **4.2-4.4 ms** for XLA's fused plain
attention on the same H100.

## Proposed fix

PR to follow: `Reactant.XLA.set_cudnn_nvrtc_override!(libnvrtc; libnvrtc_builtins, force)`,
which validates a candidate (exists, loads, `nvrtcVersion` major == `Reactant_jll`'s CUDA
major), dlopens the builtins so the soname lookup needs no `LD_LIBRARY_PATH` entry, and
never overwrites a value the user set. `ReactantCUDAExt.__init__` calls it with
`CUDA_Runtime_jll`'s `libnvrtc_path` / `libnvrtc_builtins_path`, so `using CUDA` alongside
Reactant is enough.

The real fix is on the bundle side: link cuDNN's runtime-compiled engines against a matching
NVRTC, or ship `libnvrtc.so` + builtins under `lib/cuda` in `Reactant_jll` and set the
override from there. Happy to redo it that way if you'd rather -- the Julia-side helper is
what I could verify without rebuilding the jll.

Note that this also explains why the `--xla_gpu_enable_cudnn_fmha` rewriter path could never
have produced a working kernel in this bundle. (In the XLA that ships here the auto rewriter
itself appears to be gone, only the flag string remains, so the `__cudnn$fmhaSoftmax` /
`...Backward` custom calls are the way to reach it.) See also EnzymeAD/Enzyme-JAX#994.

## Environment

Reactant.jl main (v0.2.283 source), `Reactant_jll` v0.0.405, CUDA.jl v6.2.2, Julia 1.12.7,
Linux x86_64. H100 80GB, driver 580.105.8 / CUDA 13.0, XLA runtime 13.1, cuDNN 9.21.0.

Verified on that machine after the change: `test/core/cudnn_nvrtc.jl` 12/12,
`test/integration/cudnn.jl` 4/4 (red before it, with "No valid execution plans built").

AsafManela:
Fixes #3226. Requires EnzymeAD/ReactantBuilder#221.

**Rescoped** after @wsmoses's review: this no longer borrows CUDA.jl's libnvrtc. The JLL
now ships its own (ReactantBuilder#221), and this is just the Julia half that finds it,
so no CUDA toolkit outside the JLL is involved and the CUDA extension is untouched.

cuDNN's runtime-compiled engines JIT their kernels with NVRTC when they build an execution
plan. They never call the NVRTC linked statically into libReactantExtra.so -- cuDNN's
loader dlopens a *shared* libnvrtc named by `CUDNN_NVRTC_OVERRIDE_PATH`. With nothing to
point it at, every plan build fails and fused attention is unreachable. The gdb trace
showing cuDNN calling its own embedded NVRTC rather than the bundle's is in #3226.

`Reactant.XLA.__init__` now calls `configure_cudnn_nvrtc_override!`, which installs
`lib/cuda/lib/libnvrtc.so.<major>` when the JLL carries one, dlopens its builtins so the
soname lookup needs no `LD_LIBRARY_PATH` entry, and leaves an existing
`CUDNN_NVRTC_OVERRIDE_PATH` alone. On a JLL without the bundled NVRTC it does nothing.

That is the whole implementation -- 59 lines. An earlier revision also exported a public
`set_cudnn_nvrtc_override!(path; force)` with version validation, which I have dropped:
it existed to point cuDNN at CUDA.jl's libnvrtc, the validation cannot fail for a library
shipped in the same bundle, and users who want a different NVRTC can simply export
`CUDNN_NVRTC_OVERRIDE_PATH`, which this respects.

The builtins `dlopen` is deliberate: the shipped libnvrtc has no RUNPATH and resolves
`libnvrtc-builtins` by soname. I could not prove it necessary on my machine -- both nodes
I tested have a system CUDA in `ldconfig` that satisfies the lookup anyway -- but without
it the bundle would depend on a CUDA install outside the JLL, which is the thing this is
meant to remove.

### Testing

Built a JLL from ReactantBuilder#221 and installed it as the artifact on an H100 (driver
580.105.8, cuDNN 9.21.0). With **no environment variables set**:

```
jll libnvrtc    = .../lib/cuda/lib/libnvrtc.so.13
override in env = .../lib/cuda/lib/libnvrtc.so.13     <- set by Reactant itself
cuDNN NVRTC override helper            | 17 17
cuDNN runtime-compiled fused attention |  4  4
```

`test/integration/cudnn.jl` is a hand-emitted `__cudnn$fmhaSoftmax` custom call checked
against a reference attention; it fails to build a plan without a usable NVRTC. At
B=2, H=6, T=S=512, D=128, bf16, causal the cuDNN forward runs in 0.43-0.49 ms against
4.2-4.4 ms for XLA's fused plain attention.

I also confirmed that the NVRTC from the same redistributable as the bundled `ptxas`
(13.1.115) is sufficient, so no CUDA version bump is needed -- cuDNN needs a real shared
libnvrtc of matching major version, nothing more.

https://claude.ai/code/session_012k6KZhhCNSCvZFKzErHaHH

### DIFF
--- a/src/xla/NVRTC.jl
+++ b/src/xla/NVRTC.jl
@@ -1,0 +1,59 @@
+# cuDNN's runtime-compiled engines (fused attention / SDPA, runtime fusion) JIT their
+# kernels with NVRTC when an execution plan is built. They do not call the NVRTC linked
+# statically into libReactantExtra.so: cuDNN's loader dlopens a *shared* libnvrtc named by
+# `CUDNN_NVRTC_OVERRIDE_PATH`, and with nothing to point it at every plan build fails with
+# `CUDNN_STATUS_INTERNAL_ERROR_COMPILATION_FAILED ... compilationResult != NVRTC_SUCCESS`,
+# which XLA surfaces as "[cudnn_frontend] Error: No valid execution plans built.".
+# The CUDA bundles ship one in `lib/cuda/lib` for exactly this purpose.
+
+const CUDNN_NVRTC_OVERRIDE_ENV = "CUDNN_NVRTC_OVERRIDE_PATH"
+
+# Shared CUDA libraries `Reactant_jll` ships next to its `ptxas` and `libdevice`. Returns
+# `nothing` off-CUDA and on bundles predating the shipped NVRTC.
+function _reactant_jll_cuda_lib(prefix::AbstractString)
+    Reactant_jll.is_available() || return nothing
+    dir = joinpath(dirname(Reactant_jll.libReactantExtra_path), "cuda", "lib")
+    isdir(dir) || return nothing
+    names = filter(n -> startswith(n, prefix), readdir(dir))
+    isempty(names) && return nothing
+    # Prefer the shortest soname (`libnvrtc.so.13` over `libnvrtc.so.13.1.115`).
+    return joinpath(dir, first(sort(names; by=length)))
+end
+
+"""
+    reactant_jll_libnvrtc()
+
+Path to the shared NVRTC `Reactant_jll` ships in `lib/cuda/lib`, or `nothing`.
+"""
+reactant_jll_libnvrtc() = _reactant_jll_cuda_lib("libnvrtc.so.")
+
+"""
+    reactant_jll_libnvrtc_builtins()
+
+Path to the `libnvrtc-builtins` matching [`reactant_jll_libnvrtc`](@ref), or `nothing`.
+"""
+reactant_jll_libnvrtc_builtins() = _reactant_jll_cuda_lib("libnvrtc-builtins.so.")
+
+"""
+    configure_cudnn_nvrtc_override!()
+
+Point cuDNN's runtime-compiled engines at the shared NVRTC bundled in `Reactant_jll`.
+Returns the override path in effect, or `nothing` when the bundle carries no NVRTC.
+
+A `CUDNN_NVRTC_OVERRIDE_PATH` already set in the environment is left alone, so pointing
+cuDNN somewhere else is just a matter of exporting that variable. Must run before cuDNN
+builds its first runtime-compiled plan; cuDNN reads the variable once.
+"""
+function configure_cudnn_nvrtc_override!()
+    haskey(ENV, CUDNN_NVRTC_OVERRIDE_ENV) && return ENV[CUDNN_NVRTC_OVERRIDE_ENV]
+    libnvrtc = reactant_jll_libnvrtc()
+    libnvrtc === nothing && return nothing
+    builtins = reactant_jll_libnvrtc_builtins()
+    if builtins !== nothing
+        # libnvrtc has no RUNPATH and dlopens its builtins by soname, which resolves once
+        # the library is loaded in the process -- no LD_LIBRARY_PATH entry needed.
+        Libdl.dlopen(builtins, Libdl.RTLD_LAZY | Libdl.RTLD_GLOBAL)
+    end
+    ENV[CUDNN_NVRTC_OVERRIDE_ENV] = libnvrtc
+    return libnvrtc
+end
--- a/src/xla/XLA.jl
+++ b/src/xla/XLA.jl
@@ -27,6 +27,7 @@ function LLVMclopts(opts...)
     return MLIR.API.ReactantLLVMParseCommandLineOptions(length(args), args, C_NULL)
 end
 
+include("NVRTC.jl")
 include("Distributed.jl")
 include("Client.jl")
 include("Device.jl")
@@ -205,6 +206,10 @@ function __init__()
 
         initLogs = Libdl.dlsym(Reactant_jll.libReactantExtra_handle, "InitializeLogs")
         ccall(initLogs, Cvoid, ())
+
+        # Let cuDNN's runtime-compiled engines (fused attention) find a shared NVRTC.
+        # Has to happen before cuDNN builds its first such plan; see `NVRTC.jl`.
+        configure_cudnn_nvrtc_override!()
         # Add most log level
         # SetLogLevel(0)
 
--- a/test/core/cudnn_nvrtc.jl
+++ b/test/core/cudnn_nvrtc.jl
@@ -1,0 +1,37 @@
+using Reactant, Test
+
+# cuDNN's runtime-compiled engines (fused attention / SDPA, runtime fusion) JIT their
+# kernels with NVRTC when an execution plan is built, through a shared libnvrtc named by
+# `CUDNN_NVRTC_OVERRIDE_PATH`. Reactant points that at the one the JLL bundles. This covers
+# the lookup; the GPU end-to-end check lives in `integration/cudnn.jl`.
+
+@testset "cuDNN NVRTC override" begin
+    ENVKEY = Reactant.XLA.CUDNN_NVRTC_OVERRIDE_ENV
+    @test ENVKEY == "CUDNN_NVRTC_OVERRIDE_PATH"
+
+    # An override the user already set wins, whatever the bundle carries.
+    withenv(ENVKEY => "/user/choice/libnvrtc.so") do
+        @test Reactant.XLA.configure_cudnn_nvrtc_override!() == "/user/choice/libnvrtc.so"
+        @test ENV[ENVKEY] == "/user/choice/libnvrtc.so"
+    end
+
+    # The bundled NVRTC: present on CUDA bundles that ship it, absent otherwise. Whatever
+    # the answer, the library and its builtins must agree with each other.
+    libnvrtc = Reactant.XLA.reactant_jll_libnvrtc()
+    builtins = Reactant.XLA.reactant_jll_libnvrtc_builtins()
+    if libnvrtc === nothing
+        @test builtins === nothing
+        withenv(ENVKEY => nothing) do
+            @test Reactant.XLA.configure_cudnn_nvrtc_override!() === nothing
+            @test !haskey(ENV, ENVKEY)
+        end
+    else
+        @test isfile(libnvrtc)
+        @test occursin("libnvrtc.so.", basename(libnvrtc))
+        @test builtins !== nothing && isfile(builtins)
+        withenv(ENVKEY => nothing) do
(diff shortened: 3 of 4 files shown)


wsmoses:
I'm not sure this is the proper fix and I would be more inclined to figure out why the nvrtc in the libReactant jll doesn't work

AsafManela:
That's a fair push, and I did chase it — the short version is that the NVRTC in the jll works fine and is never called.

`nvrtcHelper::load()` (libcudnn_graph, inside libReactantExtra.so) fills a function-pointer table for the NVRTC API. Under gdb, every entry resolves to cuDNN's own statically linked symbols, not the bundle's public `nvrtc*`:

```
=== at nvrtcHelper::load()+1673: this=0x5585110
    this+0x10 = 0x7fff8cc64e20  libcudnn_graph_static_v9_6491af22...   <- nvrtcVersion
    this+0x18 = 0x7fff8cc78a40  libcudnn_graph_static_v9_de4d45ef...   <- nvrtcCreateProgram
    this+0x28 = 0x7fff8cc76090  libcudnn_graph_static_v9_de34ec09...   <- nvrtcCompileProgram
    this+0x30 = 0x7fff8cc65bf0  libcudnn_graph_static_v9_c2de0b4f...   <- nvrtcGetProgramLogSize
    this+0x38 = 0x7fff8cc65ce0  libcudnn_graph_static_v9_f43058e2...   <- nvrtcGetProgramLog
    ...
    public nvrtcVersion at 0x7fff9104cbe0 (DIFFERENT from the table entry)
```

That version entry is a two-instruction stub hardcoding 13.2:

```
0x7fff8cc64e2a <libcudnn_graph_static_v9_6491af22...+10>:  movl  $0xd,(%rdi)   # major = 13
0x7fff8cc64e30 <libcudnn_graph_static_v9_6491af22...+16>:  xor   %eax,%eax     # NVRTC_SUCCESS
0x7fff8cc64e32 <libcudnn_graph_static_v9_6491af22...+18>:  movl  $0x2,(%rsi)   # minor = 2
0x7fff8cc64e38 <libcudnn_graph_static_v9_6491af22...+24>:  ret
```

Calling it through the table gives `rc=0 major=13 minor=2`, while the bundle's public `nvrtcVersion` returns **13.1**. And 13.1 is what the rest of the bundle is: `nm` over the .so finds exactly one builtins version, `__nvrtc_builtins_static__13_1_*`, and the artifact ships `lib/cuda/bin/ptxas` = `release 13.1, V13.1.115` with a 13.1 libdevice.

Breaking on the table's compile entry:

```
=== embedded nvrtcCompileProgram #1 prog=0x834e840 numOptions=27
    -> rc=5
    embedded nvrtcGetErrorString(rc) = 'NVRTC_ERROR_INVALID_OPTION'
    === NVRTC LOG ===
    nvrtc: error: invalid value for --gpu-architecture (-arch)
```

`strace -f -e openat` confirms no `libnvrtc*` is ever dlopened, which is why `LD_LIBRARY_PATH` / `CUDA_PATH` can't reach any of this.

One data point against "cuDNN emits a 13.2-only arch spelling": overriding with CUDA.jl's **13.0** libnvrtc works just as well as 13.3. The option string is ordinary — it's specifically the in-bundle compiler that rejects it.

So I agree the proper fix is on the bundle side. As far as I can tell it's one of:

**(a) Align the versions** — move the hermetic CUDA/cuDNN pins (`cuda_redist_init_repositories` / `cudnn_redist_init_repository`) so the statically linked NVRTC and builtins match what cuDNN 9.21's runtime-compiled engines expect, instead of the current 13.1-everywhere / cuDNN-says-13.2 split.

**(b) Ship a shared libnvrtc in the artifact** — put `libnvrtc.so` + `libnvrtc-builtins.so` under `lib/cuda/` (where `ptxas` and `libdevice.10.bc` already live) and point `CUDNN_NVRTC_OVERRIDE_PATH` at the jll's own copy.

I can't do either here: no bazel, the output is a 3 GB artifact, and the jll is built by a separate pipeline rather than by a Reactant.jl PR. For (a) I also can't tell you *which* piece has to move — whether cuDNN embeds a complete NVRTC 13.2 that's being cross-linked against 13.1's builtins, or whether the static link merged the two — and cuDNN 9.x redists are per-CUDA-major, so "a cuDNN for 13.1" may not be a knob that exists. I'd rather not have you rebuild a 3 GB artifact on my hypothesis.

Note that (b) needs exactly the hook in this PR — the jll would call `set_cudnn_nvrtc_override!` with its own libnvrtc instead of CUDA.jl's — so this isn't really competing with the proper fix. Happy to go whichever way you prefer:

1. keep it as-is (auto-wires CUDA_Runtime_jll's libnvrtc at load time),
2. narrow it to the helper + tests with no automatic `__init__` wiring, so it does nothing by default and is purely a documented escape hatch, or
3. close it and leave #3226 standing as the diagnosis for whoever does the bundle fix.

3 is genuinely fine by me — the diagnosis is the useful half and it's already filed.

wsmoses:
@AsafManela since this seems llm generated, can you task your llm with actually fixing it jll side?

AsafManela:
It's llm generated with guidance ;) 
But I couldn't figure out where or how to fix the jll side. If you can provide some guidance I'm happy to guide claude that way.

AsafManela:
Pushed a fix for CI: the integration test was failing the `integration-ifrt`/`integration-pjrt`
jobs, and that was my fault.

`test/integration/cudnn.jl` only skipped when the backend was not CUDA. On the CUDA
runners it therefore ran its body and asserted that the override is installed -- which
cannot hold until the bundles actually ship a libnvrtc (EnzymeAD/ReactantBuilder#221). It
now also skips when `reactant_jll_libnvrtc()` returns `nothing`, i.e. when there is
nothing for cuDNN to dlopen and so nothing to exercise. It goes live by itself once the
JLL carries one.

For reference, `main` at the tip of this PR's base fails `core-ifrt` and `core-pjrt`
already, independently of this branch; those two are not from this change.

Verified both directions: on a stock JLL the testset skips cleanly, and on a locally built
JLL that does carry the bundled libnvrtc it still runs for real -- 8/8 and 4/4 on an H100.

PR #3235, opened 2026-08-30
Reserve an address range per XLA allocation

wsmoses:
The handle a reactant allocation returned was an 8-byte heap slot, so pointer arithmetic past it walked into neighboring slots, exact-match lookup was the only sound resolution, and a stray host write through a device pointer silently corrupted the slot (observed in MFEM as a `PjRtBuffer*` holding the bit pattern of `-0.5`).

Reserve an inaccessible `mmap` range as large as the buffer instead:

- interior pointers resolve to `(buffer, offset)` with a bounds check,
- foreign pointers report precisely instead of resolving to garbage,
- any host dereference of a device handle faults at the offender.

First of a series splitting #3223 into per-feature PRs.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -1,4 +1,10 @@
+#include <cstddef>
 #include <iostream>
+
+// AddressReservation.cpp; reserves inaccessible address ranges used as opaque
+// allocation handles.
+void *reactantReserveAddressRange(size_t nbytes);
+void reactantReleaseAddressRange(void *base, size_t nbytes);
 
 #include "mlir-c/IR.h"
 #include "mlir-c/Support.h"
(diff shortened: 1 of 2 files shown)


wsmoses:
Windows jll fixed: the address-range reservation now uses VirtualAlloc(MEM_RESERVE, PAGE_NOACCESS)/VirtualFree on _WIN32 and keeps mmap/PROT_NONE elsewhere (mingw has no sys/mman.h). Whole stack rebased on the fix; Linux cuda build re-verified.

wsmoses:
The earlier _WIN32 fix compiled, but `<windows.h>` at the top of API.cpp leaked its function-style macros (`ZeroMemory`, `IGNORE`, ...) into the Enzyme and MLIR headers included after it. Since API.cpp also has includes deep in the file body, there is no safe position for `windows.h` there — the reservation calls now live in an isolated `AddressReservation.cpp` (picked up by the BUILD srcs glob), and API.cpp just declares the two helpers.

wsmoses:
Note: the Windows-fix force-push earlier today unintentionally dropped the #3236 merge from this branch (the rebase happened concurrently). Restored — this PR now carries both the address-range reservation and the DtoD host staging, matching the #3236 merge. Also: #3239 (exec-pipeline retry fallbacks) has been dropped from the stack — with EnzymeAD/Enzyme-JAX#3024 merged the suite needs zero fallbacks, and the remaining EnzymeAD/Enzyme-JAX#3000 trigger now aborts in a way no in-process retry can catch; #3240 stacks directly on #3238.

wsmoses:
Un-stacked the series per request: the DtoD commit moved out of this PR into #3243 (targeting main), so this PR is back to the address-range reservation alone. #3237 and #3238 now target main independently; #3240 stacks only on this PR (it needs the size-tracked ranges); #3241 merges last. Every head compiles standalone (Linux CUDA ReactantExtraLib).

PR #3239, opened 2026-08-30
reactantXLAExec: retry the exec pipeline on invalid IR, with and without slice motion

wsmoses:
Two-stage insurance for the exec pipeline, plus a debugging lever:

1. A hlo-opt pattern interplay (slice-of-broadcast motion) can leave type-inconsistent IR; verify the module after the passes and, on failure, rebuild from source and retry with the slice-motion pattern group disabled. (The interplay itself is EnzymeAD/Enzyme-JAX#3000 / fixed cases in EnzymeAD/Enzyme-JAX#3024 — this keeps the runtime alive for the next one.)
2. If both optimizer attempts leave inconsistent IR, run once more without the hlo optimizer entirely — slower kernels beat dying.
3. `REACTANT_XLA_NO_HLO_OPT` skips the optimizer up front, to split optimizer miscompiles from raising ones without a rebuild.

Part of the #3223 split; stacked on #3238.

### DIFF
--- a/deps/ReactantExtra/API.cpp
+++ b/deps/ReactantExtra/API.cpp
@@ -3786,7 +3786,8 @@ REACTANT_ABI void reactantXLAExec(LinkableRuntime **__restrict__ lrtP,
     pm.addPass(mlir::stablehlo::createStablehloRefineShapesPass());
     pm.addNestedPass<mlir::func::FuncOp>(
         stablehlo::createStablehloCanonicalizeDynamismPass());
-    pm.addPass(mlir::enzyme::createEnzymeHLOOptPass());
+    if (!getenv("REACTANT_XLA_NO_HLO_OPT"))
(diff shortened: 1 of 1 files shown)


wsmoses:
Dropped from the stack rather than merged: with EnzymeAD/Enzyme-JAX#3024 in, the full MFEM suite runs with zero fallback firings, and the one remaining known trigger (EnzymeAD/Enzyme-JAX#3000) now manifests at Enzyme-JAX main as a hard abort inside enzyme-hlo-opt (failed slice type inference → report_fatal_error) that no in-process retry can catch — even the without-slice-motion retry mask no longer avoids it (details on that issue). Invalid IR still fails loudly at the post-pipeline verify, and the `REACTANT_XLA_NO_HLO_OPT` debugging switch moved into #3241. #3240 now stacks directly on #3238.

PR #3245, opened 2026-08-31
Test target GPU instead of CUDA

luraess:
This PR fixes tests running on any-GPU while it currently targets CUDA-only. The file two directories over already uses the suggested API.

### DIFF
--- a/test/core/sharding/optimize_comm.jl
+++ b/test/core/sharding/optimize_comm.jl
@@ -3,7 +3,9 @@ using Reactant, Test, FileCheck
 const addressable_devices = Reactant.addressable_devices()
 
 const RunningOnTPU = contains(string(Reactant.devices()[1]), "TPU")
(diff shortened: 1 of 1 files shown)


test/core/sharding/optimize_comm.jl
```diff
@@ -3,7 +3,7 @@ using Reactant, Test, FileCheck
 const addressable_devices = Reactant.addressable_devices()
 
 const RunningOnTPU = contains(string(Reactant.devices()[1]), "TPU")
-const RunningOnGPU = contains(string(Reactant.devices()[1]), "CUDA")
+const RunningOnGPU = Reactant.XLA.platform_name(Reactant.XLA.default_backend()) != "cpu"
```
wsmoses: change to or rocm, because this otherwise is true on tpu too?

PR #3248, opened 2026-09-02
Replace `mlir_c` with `Reactant_jll.libReactantExtra`

### LINKED ISSUE #3190: Reactant is not relocatable due to `MLIR.API.mlir_c`

Pangoraw:
Alternative fix for #3190 we discussed in the previous meeting.

### DIFF
--- a/src/compiler/OptimizationPasses.jl
+++ b/src/compiler/OptimizationPasses.jl
@@ -2,6 +2,7 @@
 
 using ..Reactant:
     Reactant, MLIR, OptimizeCommunicationOptions, ShardyPropagationOptions, CompileOptions
+using Reactant_jll: Reactant_jll
 
 const BFLOAT16_COMPILE_TYPE = Ref{DataType}(Float32)
 const DEBUG_KERNEL = Ref{Bool}(false)
@@ -235,7 +236,7 @@ function run_pass_pipeline!(
     MLIR.IR.enable_verifier!(pm, enable_verifier)
     opm = MLIR.IR.OpPassManager(pm)
     # TODO: why isn't this being auto-generated?
-    @ccall MLIR.API.mlir_c.addSdyPropagationPipeline(
+    @ccall Reactant_jll.libReactantExtra.addSdyPropagationPipeline(
         opm::MLIR.API.MlirOpPassManager,
         propagation_options.keep_sharding_rules::UInt8,
         propagation_options.conservative_propagation::UInt8,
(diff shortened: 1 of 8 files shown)


src/mlir/MLIR.jl
```diff
@@ -62,7 +60,7 @@ module API
     function ClientCompileWithProto(
         client, cmod, compile_options_proto::Vector{UInt8}, compile_options_proto_size
     )
-        @ccall mlir_c.ClientCompileWithProto(
+        @ccall Reactant_jll.libReactantExtra.ClientCompileWithProto(
```
wsmoses: minor, but why the ccall instead of doing MLIR.API.ClientCompileWithProto or w/e

src/compiler/OptimizationPasses.jl
```diff
@@ -235,7 +236,7 @@ function run_pass_pipeline!(
     MLIR.IR.enable_verifier!(pm, enable_verifier)
     opm = MLIR.IR.OpPassManager(pm)
     # TODO: why isn't this being auto-generated?
-    @ccall MLIR.API.mlir_c.addSdyPropagationPipeline(
+    @ccall Reactant_jll.libReactantExtra.addSdyPropagationPipeline(
```
wsmoses: same comment for here and others

Pangoraw: looks like C ABI was wrong for this one function, fixed now.

PR #3266, opened 2026-09-10
fix: improve type inference on `mapreduce` for `TracedRArray`

### LINKED ISSUE #3261: `return_type` of a reduction over a `TracedRArray` widens to `Union{TracedRArray, TracedRNumber}` inside a trace
Inside a `@compile` trace, `Core.Compiler.return_type` (and therefore `Base.promote_op`) reports a non-concrete `Union{TracedRArray, TracedRNumber}` for reductions over a `TracedRArray`. Outside a trace, inference of the very same call is exact.

## MWE

Only dependency is Reactant:

```julia
using Reactant
using Reactant: TracedRArray

function f(v)
    println("inside  a trace: ", Core.Compiler.return_type(sum, Tuple{typeof(v)}))
    return sum(v)
end

println("outside a trace: ", Base.infer_return_type(sum, (TracedRArray{Float32, 1},)))
x = Reactant.to_rarray(rand(Float32, 4))
@compile f(x)
```

```
outside a trace: Reactant.TracedRNumber{Float32}
inside  a trace: Union{TracedRArray, Reactant.TracedRNumber}
```

## Scope

Not every call is affected — it is specifically the reductions that go from an array to a scalar. Querying `return_type` from inside a trace:

| call | inferred inside a trace | inferred outside |
| --- | --- | --- |
| `identity(::TracedRArray{Float32,1})` | `TracedRArray{Float32, 1}` | same |
| `length(::TracedRArray{Float32,1})` | `Int64` | same |
| `(+)(::TracedRNumber{Float32}, ::TracedRNumber{Float32})` | `TracedRNumber{Float32}` | same |
| `abs(::TracedRNumber{Float32})` | `TracedRNumber{Float32}` | same |
| `sum(::Vector{Float32})` | `Float32` | same |
| `sum(::TracedRArray{Float32,1})` | **`Union{TracedRArray, TracedRNumber}`** | `TracedRNumber{Float32}` |
| `maximum(::TracedRArray{Float32,1})` | **`Union{TracedRArray, TracedRNumber}`** | `TracedRNumber{Float32}` |

Same result for `Float64`, and for `prod`/`minimum`.

## Why it matters

[DispatchDoctor.jl](https://github.com/MilesCranmer/DispatchDoctor.jl)'s `@stable` inserts a `Base.promote_op` call into every function it wraps and throws a `TypeInstabilityError` when the result is not concrete. Any `@stable` function that reduces a `TracedRArray` therefore throws as soon as it is traced, even though it is perfectly inferrable outside the trace:

```
DispatchDoctor.TypeInstabilityError: Instability detected in `colsum!!` with arguments
`(Reactant.TracedRNumber{Float32}, Reactant.TracedRArray{Float32, 1})`.
Inferred to be `Union{Reactant.TracedRArray, Reactant.TracedRNumber}`, which is not a concrete type.
```

We hit this in [CoolPDLP.jl](https://github.com/JuliaDecisionFocusedLearning/CoolPDLP.jl) and had to opt that one method out of the type-stability checks. Anything else that inspects inferred types while tracing (`promote_op`-based `similar`/output-type computations, for instance) would see the same widening.

## Versions

- Reactant v0.2.285
- Julia 1.12.6, x86_64-linux-gnu, CPU backend

cc @gdalle

gdalle:
Fixes #3261:

- Some of the internals of `overloaded_mapreduce` still resist inference
- The key fix is to annotate the return with the proper element type and shape
- I also fixed a variable captured in closure bug while I was at it

### DIFF
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -673,9 +673,9 @@ end
 
 @noinline function reshape(
     x::TracedRArray{T,N},
-    dims::Union{Vector{Int},NTuple{M,Int}};
+    dims::Vector{Int};
     location=mlir_stacktrace("reshape", @__FILE__, @__LINE__),
-) where {T,N,M}
+) where {T,N}
     @assert length(x) == prod(dims)
 
     # HLO reshape semantics collapse the opposite way
--- a/src/Ops.jl
+++ b/src/Ops.jl
@@ -673,9 +673,9 @@ end
 
 @noinline function reshape(
     x::TracedRArray{T,N},
-    dims::Vector{Int};
+    dims::Union{Vector{Int},NTuple{M,Int}};
     location=mlir_stacktrace("reshape", @__FILE__, @__LINE__),
-) where {T,N}
+) where {T,N,M}
     @assert length(x) == prod(dims)
 
     # HLO reshape semantics collapse the opposite way
--- a/src/TracedRArray.jl
+++ b/src/TracedRArray.jl
@@ -192,41 +192,60 @@ function overloaded_mapreduce(
     init=Base._InitialValue(),
 ) where {T,N}
     original_dims = dims
-    dims isa Int && (dims = Int64[dims])
-    dims isa Colon && (dims = collect(Int64, 1:N))
-    dims isa Vector{Int64} || (dims = collect(Int64, dims))
-    dims = sort(dims)
+    # don't reassign dims to avoid the "captured variable in closure" type instability 
+    normalized_dims = if dims isa Int
+        Int64[dims]
+    elseif dims isa Colon
+        collect(Int64, 1:N)
+    elseif dims isa Vector{Int64}
+        dims
+    else
+        collect(Int64, dims)
+    end |> sort
 
     op_in_T = unwrapped_eltype(Core.Compiler.return_type(f, Tuple{T}))
     # `op` may accumulate in a wider type than it consumes: `+(::Bool, ::Bool)::Int64`, so
     # `sum` over booleans is an `Int64` in Base. `__default_init` reports that wider type
     # (it is the type of `op`'s identity element), so reduce in it.
     init_val = __default_init(op_in_T, op)
-    op_in_T = unwrapped_eltype(typeof(init_val))
-    reduce_init = Reactant.promote_to(TracedRNumber{op_in_T}, init_val)
+    # TODO: double-check that this is the eltype of the output
+    op_out_T = unwrapped_eltype(typeof(init_val))
+    reduce_init = Reactant.promote_to(TracedRNumber{op_out_T}, init_val)
 
     # Widen *after* applying `f`, not before. `f` decides the element type actually being
     # reduced, so converting `A` up front is undone by anything narrowing -- e.g. the
     # predicate in `sum(x -> x == 1, a)` returns `Bool` whatever `A` was converted to.
     # Reducing in the narrow type is silently wrong: `stablehlo.add` over `i1` is a logical
     # `or`, and small integers wrap.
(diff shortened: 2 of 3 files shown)


src/Ops.jl
```diff
@@ -673,9 +673,9 @@ end
 
 @noinline function reshape(
     x::TracedRArray{T,N},
-    dims::Vector{Int};
+    dims::Union{Vector{Int},NTuple{M,Int}};
```
wsmoses: can we do without this potentially? I worry a bit about the blowup in fns generated by all the reshapes

src/TracedRArray.jl
```diff
@@ -192,41 +192,62 @@ function overloaded_mapreduce(
...
     if res isa TracedRNumber
-        res = TracedRArray{unwrapped_eltype(res),0}((), res.mlir_data, ())
+        res = TracedRArray{op_out_T,0}((), res.mlir_data, ())
     end
-    return @opcall reshape(res, [ifelse(i in dims, 1, size(A, i)) for i in 1:N])
+    # use ntuple so that the array rank is inferrable, the corresponding method for reshape was added
+    shape = ntuple(i -> ifelse(i in normalized_dims, 1, size(A, i)), Val(N))
+    res_reshaped = @opcall reshape(res, shape)
+    # force return type inference
+    return TracedRArray{op_out_T,N}((), res_reshaped.mlir_data, shape)
```
wsmoses: don't make a new traced r array, just type assert it as ::TracedRArray{...}

src/TracedRArray.jl
```diff
@@ -192,25 +192,35 @@ function overloaded_mapreduce(
...
+    )
 
     op_in_T = unwrapped_eltype(Core.Compiler.return_type(f, Tuple{T}))
     # `op` may accumulate in a wider type than it consumes: `+(::Bool, ::Bool)::Int64`, so
     # `sum` over booleans is an `Int64` in Base. `__default_init` reports that wider type
     # (it is the type of `op`'s identity element), so reduce in it.
     init_val = __default_init(op_in_T, op)
-    op_in_T = unwrapped_eltype(typeof(init_val))
-    reduce_init = Reactant.promote_to(TracedRNumber{op_in_T}, init_val)
+    # TODO: double-check that this is the eltype of the output
```
wsmoses: remove these todos, etc. I think your change is simply adding `res_reshaped::TracedRArray{op_out_T,N}`

gdalle:
I think the remaining failures are not related to this change?

src/TracedRArray.jl
```diff
@@ -191,42 +191,59 @@ function overloaded_mapreduce(
...
+    )
 
     op_in_T = unwrapped_eltype(Core.Compiler.return_type(f, Tuple{T}))
     # `op` may accumulate in a wider type than it consumes: `+(::Bool, ::Bool)::Int64`, so
     # `sum` over booleans is an `Int64` in Base. `__default_init` reports that wider type
     # (it is the type of `op`'s identity element), so reduce in it.
     init_val = __default_init(op_in_T, op)
-    op_in_T = unwrapped_eltype(typeof(init_val))
-    reduce_init = Reactant.promote_to(TracedRNumber{op_in_T}, init_val)
+    op_out_T = unwrapped_eltype(typeof(init_val))
```
wsmoses, line 212: can you avoid rename churn and try to keep this as slim as possible?

gdalle: The rename was voluntary to highlight that this is a different type than the input type but sure

wsmoses: lets try to keep things to one change at a time

PR #3270, opened 2026-09-11
move KernelInterface integration code to src

### LINKED ISSUE #3263: Why doesn't `adapt(::ReactantBackend, a)` create a `ConcreteRSomething`?
I'm a bit surprised by this behavior
```julia
julia> using Reactant, KernelAbstractions, Adapt

julia> backend = Base.get_extension(Reactant, :ReactantKernelAbstractionsExt).ReactantBackend()
ReactantKernelAbstractionsExt.ReactantBackend()

julia> adapt(backend, [1.0, 2.0])
2-element Vector{Float64}:
 1.0
 2.0
```

mofeing:
fixes #3263

the inferfaces defined in KernelAbstractions has been moved to a new package named KernelInterface without dependencies. this PR moves the code that implements such interfaces to the main Reactant library so that users like [Oceananigans](https://github.com/CliMA/Oceananigans.jl/blob/8abb332a4cb73c45f67f38b57f22b02970b0895f/ext/OceananigansReactantExt/Architectures.jl#L17), [SpeedyWeather](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/60aff02917609d3f8e878aa0a7c435b1f6a9afb3/SpeedyWeatherInternals/ext/SpeedyWeatherInternalsReactantExt.jl#L14) or #3263 don't need to access internals with `Base.get_extension(Reactant, :ReactantKernelAbstractionsExt)`.

This requires KernelAbstractions v0.10, which hasn't been released yet, so it will stay as a draft PR until the release.


wsmoses:
I'm going to temporarily block this. Like mentioned on the issue this is still an implementation detail and doesn't yet implement everything expected in a KA backend.

Can discuss design in a later meeting

mofeing:
No problem from my side

PR #3271, opened 2026-09-11
Error on @trace if assignments to untraced mutable struct fields

ChrisRackauckas-Claude:
## Summary

Split out of https://github.com/EnzymeAD/Reactant.jl/pull/3232#issuecomment-5623499905, where this change was flagged as independent of the enum tracing work.

When a `@trace if` branch assigns a new value to a field of a captured mutable struct whose slot is untraced (e.g. a plain `Int` or `Float64` field), the `if` result has no location to be written back into. Previously the assignment was silently discarded; `if_condition` now checks such `:resarg` paths and raises an error explaining that the field must be traced before the `if`, e.g. via `ReactantCore.promote_to_traced`.

Caveat: the check only inspects collected result paths, so it is not a general detector of unsupported mutations — a leaf the tracing machinery never tracks produces no path and remains undetectable (this is noted in a comment on `check_untraced_branch_state`).

The tutorial's conditional-control-flow section now documents the requirement on mutable struct fields.

## Test plan

- [x] `core/control_flow` passes locally (Linux CPU, PJRT): 150/150, including the new "if: assignment to an untraced struct field" testset which asserts the error fires for plain `Int` and `Float64` fields and that an untouched untraced field still works.
- [x] Verified locally that the numeric case throws `"untraced location"` with this change and silently dropped the assignment without it.

🤖 Generated with [Devin](https://devin.ai) (harness: Devin CLI, model: SWE-2 High) — local CLI session, no shareable URL.

### DIFF
--- a/docs/src/tutorials/control-flow.md
+++ b/docs/src/tutorials/control-flow.md
@@ -120,6 +120,13 @@ In our simple example, the condition is passed directly as an argument but
 the same mechanism is applied to conditions which are computed from within
 a function from traced arguments, leading to a traced condition.
 
+Branches can also mutate fields of a captured mutable struct, but only when
+the field already holds a traced value — the `if` result is written back into
+that location. A field holding a plain number or other untraced value has no
+location to write into, so assigning it inside `@trace if` raises an error.
+Parameterize the field type and initialize it with a traced value, e.g. via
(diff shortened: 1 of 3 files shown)


jumerckx (approved):
This looks correct to me, thanks!

Issue #253, opened 2024-11-09
32-Bit CI


avik-pal:
we need this to confirm that passing an `Int` accidentally doesn't cause breakages (see https://github.com/EnzymeAD/Reactant.jl/pull/252#discussion_r1835486794)

PR #3272, opened 2026-09-11
Accept Integer for register_backend priority kwarg

ChrisRackauckas-Claude:
## Summary

`using Reactant` cannot load at all on 32-bit Julia. Two `Int64`s in the backend
registration path reject the `Int32` values that integer literals produce on i686:

1. `register_backend(platform_name; priority::Int64, ...)` — `register_backend("cpu"; priority=100)`
   in `Accelerators.CPU.__init__` throws
   `TypeError: in keyword argument priority, expected Int64, got a value of type Int32`.
2. `Ref(0)` for the `pjrt_client_count` / `ifrt_client_count` fields, which are typed
   `Base.RefValue{Int64}` — `Ref(0)` is a `RefValue{Int32}` on i686 and there is no
   `convert(::Type{RefValue{Int64}}, ::RefValue{Int32})`, so the same call throws
   `MethodError` one argument further along.

Fixing only (1) turns the `TypeError` into a `MethodError` in the same `__init__`, so both
are needed. `priority` becomes `::Int` (machine `Int`, per review — every in-repo call site
passes a bare literal); the `RegisteredBackend.priority::Int64` field converts it, and the
counts get explicit `Int64` literals to match their field type.

Found while investigating [SciMLSensitivity.jl's `Core1` 32-bit CI lane](https://github.com/SciML/SciMLSensitivity.jl/actions/runs/34650148455/job/103431340580),
where `using Reactant` fails with exactly (1):

```
ERROR: LoadError: InitError: TypeError: in keyword argument priority, expected Int64, got a value of type Int32
Stacktrace:
  [1] __init__()
    @ Reactant.Accelerators.CPU ~/.julia/packages/Reactant/ct5sa/src/accelerators/CPU.jl:50
```

## Verification

Reactant has no 32-bit CI lane (`.github/workflows/CI.yml` covers ubuntu-24.04,
ubuntu-22.04-arm, macOS, windows, TPU), and this bug is invisible on 64-bit — `Ref(0)` *is*
`RefValue{Int64}` there — so no test added here could fail before and pass after on any
job this repo runs. See #253, which asked for exactly this lane and was closed without one.

Verified instead on real i686 Julia by extracting `RegisteredBackend` / `register_backend`
verbatim from `src/accelerators/Registration.jl` at each commit and calling
`register_backend("cpu"; priority=100, pjrt_initialize_function=()->nothing)`:

| `Registration.jl` at | i686 Julia 1.12.7 |
|---|---|
| `main` (`priority::Int64`, `Ref(0)`) | `TypeError: in keyword argument priority, expected Int64, got a value of type Int32` |
| `20810d590` (`priority::Int`, `Ref(0)`) | `MethodError: Cannot convert an object of type Base.RefValue{Int32} to an object of type Base.RefValue{Int64}` |
| `8b7c28136` (this PR head) | `register_backend OK` |

This PR head also OK on i686 Julia 1.11.9, and on x86_64 Julia 1.10.11 and 1.12.7:

```
===== AFTER FIX, i686 Julia 1.12.7 =====      Julia Int = Int32   register_backend OK
===== AFTER FIX, i686 Julia 1.11.9 =====      Julia Int = Int32   register_backend OK
===== AFTER FIX, x86_64 Julia 1.12.7 =====    Julia Int = Int64   register_backend OK
===== AFTER FIX, x86_64 Julia 1.10.11 =====   Julia Int = Int64   register_backend OK
```

`JuliaFormatter` v1 (`style = "blue"`) reports the changed file already formatted.

## What was NOT verified

- **`using Reactant` end-to-end on i686.** Blocked before Reactant's `__init__` by a
  separate upstream bug: `BFloat16s` segfaults during precompilation on i686
  (`ProcessSignaled(11)`, LLVM `DAGTypeLegalizer::SoftPromoteHalfResult` during AOT
  output). Reproduced locally on i686 Julia 1.11.9 and 1.12.7, and visible on GitHub's
  i686 1.13 runner in the [most recent SciMLSensitivity x86 run](https://github.com/SciML/SciMLSensitivity.jl/actions/runs/34656934999/job/103461409522).
  Whether that run reaches Reactant's `__init__` varies run to run; the linked job above
  is one that did. That bug is not in this repo and is not addressed here.
- **Reactant's own test suite** was not run for this change — it needs XLA and hours, and
  none of it exercises a 32-bit path.
- The `downgrade` jobs and the TPU jobs failing on this PR also fail on `main` (`9eef079`);
  they are pre-existing and unrelated.

## For the reviewer

- The `Ref(0)` → `Ref(Int64(0))` hunk was reverted by a review suggestion in `26fe25a14`
  and is restored in `8b7c28136`; the table above is the evidence that it is load-bearing.
  The alternative is to type the struct fields `Base.RefValue{Int}` and keep `Ref(0)` —
  also correct, and more consistent with `priority::Int`. Say the word if you prefer that.
- Related but deliberately out of scope: `::Int64` *argument* annotations elsewhere
  (`src/Ops.jl`, `src/xla/CompileOptions.jl:68-70`, `src/xla/PJRT/LoadedExecutable.jl:66-70`)
  have the same 32-bit hazard, but they are on tracing/execution paths, not `__init__`, so
  they do not block loading. The `::Int64`s in `src/mlir/libMLIR_h.jl` are `@ccall`
  signatures and are correct as-is.

Please ignore until reviewed by @ChrisRackauckas.

---

Original fix (`a8508b11c`, `89a50694b`) authored with Devin CLI 3000.10.21 (models SWE-2
High / SWE-2 Max); local session transcript, no shareable URL:
`/home/crackauc/.local/share/devin/cli/summaries/history_56dcef04f3304e25.md`

Investigation and `8b7c28136` 🤖 Generated with [Claude Code](https://claude.com/claude-code) 2.1.269 (model: claude-opus-5[1m])

https://claude.ai/code/session_01RvyK9WmNSp261e2Ffq2SSr

### DIFF
--- a/src/accelerators/Registration.jl
+++ b/src/accelerators/Registration.jl
@@ -29,7 +29,6 @@ function register_backend(
     preinitialize_setup_function=Returns(nothing),
 )
     @assert pjrt_initialize_function !== nothing || ifrt_initialize_function !== nothing "atleast one of pjrt_initialize_function or ifrt_initialize_function must be provided."
-    priority = Int64(priority)
 
     for backend in RegisteredBackends
         @assert backend.platform_name != platform_name "Backend with platform_name: \
--- a/src/accelerators/Registration.jl
+++ b/src/accelerators/Registration.jl
@@ -23,7 +23,7 @@ end
 
 function register_backend(
     platform_name::String;
-    priority::Integer,
+    priority::Int,
     pjrt_initialize_function=nothing,
     ifrt_initialize_function=nothing,
     preinitialize_setup_function=Returns(nothing),
@@ -44,8 +44,8 @@ function register_backend(
             priority,
             pjrt_initialize_function,
             ifrt_initialize_function,
-            Ref(Int64(0)),
-            Ref(Int64(0)),
+            Ref(0),
+            Ref(0),
             preinitialize_setup_function,
         )
         push!(RegisteredBackends, backend)
--- a/src/accelerators/Registration.jl
+++ b/src/accelerators/Registration.jl
@@ -44,8 +44,8 @@ function register_backend(
             priority,
             pjrt_initialize_function,
             ifrt_initialize_function,
-            Ref(0),
-            Ref(0),
+            Ref(Int64(0)),
+            Ref(Int64(0)),
             preinitialize_setup_function,
         )
         push!(RegisteredBackends, backend)
--- a/src/accelerators/Registration.jl
+++ b/src/accelerators/Registration.jl
@@ -43,8 +43,8 @@ function register_backend(
             priority,
             pjrt_initialize_function,
             ifrt_initialize_function,
-            Ref(0),
-            Ref(0),
+            Ref(Int64(0)),
+            Ref(Int64(0)),
             preinitialize_setup_function,
         )
         push!(RegisteredBackends, backend)
--- a/src/accelerators/Registration.jl
+++ b/src/accelerators/Registration.jl
@@ -23,12 +23,13 @@ end
 
 function register_backend(
     platform_name::String;
-    priority::Int64,
+    priority::Integer,
     pjrt_initialize_function=nothing,
     ifrt_initialize_function=nothing,
     preinitialize_setup_function=Returns(nothing),
 )
     @assert pjrt_initialize_function !== nothing || ifrt_initialize_function !== nothing "atleast one of pjrt_initialize_function or ifrt_initialize_function must be provided."
+    priority = Int64(priority)
 
     for backend in RegisteredBackends
         @assert backend.platform_name != platform_name "Backend with platform_name: \


wsmoses:
@ChrisRackauckas-Claude why not fix the callsite?

ChrisRackauckas:
The call site isn't incorrect here? Why is this not matching what should be traced from a standard `Int`? It should either be `Int` or `Integer`, but `Int64` is not right.

wsmoses:
This is an internal function for initializing the back ends, we should control both sides (and user code isn't involved here, right?). This eventually calls some external api calls that presumably take an i64, so making it take multiple input types would just lead to unnecessary compilation?

ChrisRackauckas:
No it would only compile for the function subset of type calls that actually occur.

wsmoses:
But are we sure that all call sites are Int and not Int64 in some cases as expected?

ChrisRackauckas:
Let me do a scan, if that's the case I'l just have it changed to Int instead of Integer.

ChrisRackauckas-Claude:
Scanned all `register_backend` call sites. Short answer: yes for the kwarg — but the scan found the same 32-bit bug one line deeper, which I've now fixed.

**Call sites** — all pass plain `Int` literals, none pass `Int64` explicitly:
- `src/accelerators/CPU.jl` — `priority=100`
- `src/accelerators/GPU.jl` — `priority=500` (cuda, rocm)
- `src/accelerators/TPU.jl` — `priority=1000`
- `src/accelerators/Trainium.jl` — `priority=1000`
- `src/accelerators/TT.jl` — `priority=1000`

Literals are `Int` (`Int32` on 32-bit), so `priority::Integer` covers every call site. `register_backend` isn't exported, so these are the only call sites in-repo; any external caller via `Reactant.Registration` is strictly better off with `Integer`. `Int64` is still expected in exactly one place — the `RegisteredBackend.priority::Int64` field — which the `Int64(priority)` conversion satisfies. The `best_priority = -1` comparison in `initialize_backends` promotes fine against `Int64`.

**Bug found by the scan**: `Ref(0)` for `pjrt_client_count`/`ifrt_client_count` (fields typed `Base.RefValue{Int64}`). On 32-bit `Ref(0)` produces `RefValue{Int32}`, and `convert(::Type{RefValue{Int64}}, ::RefValue{Int32})` doesn't exist — so `register_backend` would still throw a `MethodError` during `__init__` on i686 even with the relaxed kwarg. Verified locally:

```
MethodError(convert, (Base.RefValue{Int64}, Base.RefValue{Int32}(0)))
```

Fixed in 89a5069 by using `Ref(Int64(0))` for both counts.

**Related, not fixed here**: the same `::Int64` annotation pattern exists in `src/Ops.jl` (`index_vector_dim::Int64`, `feature_index::Int64`, `Vector{Int64}` params, `reduction_input_size_override::Int64=-1`) with callers passing `Int` values (e.g. `index_vector_dim=ndims(start_indices)` in `src/TracedUtils.jl`, `feature_index - 1` in `Ops.jl`). Those are tracing paths, not `__init__`, so they don't block `using Reactant` on 32-bit — but they'd throw the same `TypeError` if ever exercised there. The `::Int64`s in `src/mlir/libMLIR_h.jl` are `@ccall` signatures and are correct as-is — fixed-width C ABI, and Julia auto-converts ccall arguments.

ChrisRackauckas:
Yeah @wsmoses lots of places do this in `Int`. What's your call here? Either:

1. Everywhere else needs to promote it away from the machine Int. This is very brittle... I wouldn't do that.
2. Only accept `Int` here.
3. Just make it `Integer.

Your pick.

wsmoses:
I'm okay with 2

src/accelerators/Registration.jl
```diff
@@ -23,12 +23,13 @@ end
 
 function register_backend(
     platform_name::String;
-    priority::Int64,
+    priority::Integer,
```
ChrisRackauckas:
```suggestion
    priority::Int,
```

src/accelerators/Registration.jl
```diff
@@ -43,8 +44,8 @@ function register_backend(
             priority,
             pjrt_initialize_function,
             ifrt_initialize_function,
-            Ref(0),
-            Ref(0),
+            Ref(Int64(0)),
+            Ref(Int64(0)),
```
ChrisRackauckas, line 47:
```suggestion
            Ref(0),
            Ref(0),
```

src/accelerators/Registration.jl
```diff
@@ -23,12 +23,13 @@ end
...
 function register_backend(
     platform_name::String;
-    priority::Int64,
+    priority::Int,
     pjrt_initialize_function=nothing,
     ifrt_initialize_function=nothing,
     preinitialize_setup_function=Returns(nothing),
 )
     @assert pjrt_initialize_function !== nothing || ifrt_initialize_function !== nothing "atleast one of pjrt_initialize_function or ifrt_initialize_function must be provided."
+    priority = Int64(priority)
```
ChrisRackauckas-Claude:
```suggestion
```

ChrisRackauckas:
Okay the Ref changes are actually needed.

PR #3274, opened 2026-09-12
Accept Integer for accelerator backend priority (32-bit)

ChrisRackauckas-Claude:
## Summary
- On i686, bare integer literals are `Int32`, but `register_backend(; priority::Int64, …)` rejected them and broke `Reactant.Accelerators.CPU.__init__`.
- Widen the kwarg to `Integer` and store `Int64(priority)`; also write `priority=Int64(…)` at call sites.

## Test plan
- [ ] `using Reactant` on 32-bit Julia no longer throws `TypeError: priority expected Int64, got Int32`
- [ ] Existing 64-bit CI still green

Unblocks SciML/NeuralOperators.jl x86 CI.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/accelerators/CPU.jl
+++ b/src/accelerators/CPU.jl
@@ -49,7 +49,7 @@ end
 function __init__()
     register_backend(
         "cpu";
-        priority=100,
+        priority=Int64(100),
         pjrt_initialize_function=make_pjrt_client,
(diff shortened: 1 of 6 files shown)


ChrisRackauckas-Claude:
Duplicate of #3272 (same Integer priority fix).

PR #3311, opened 2026-09-23
Run batch-to-stablehlo within the enzyme pass

### LINKED ISSUE #3217: Batched forward mode fails to lower: enzyme.extract unsupported for export, Ops.batch emits a rank-mismatched transpose
### Environment

**Environment** — Julia 1.12.6 x86_64-linux-gnu · Reactant 0.2.280 · Reactant_jll 0.0.405+0 · Enzyme 0.13.199 · EnzymeCore 0.8.21 · Enzyme_jll 0.0.290+0 · Backend CPU (PJRT) · measured 2026-08-24

### Setup

Three ways to get the same NS Jacobian columns of the same small RHS (NS = 4, N = 12), on CPU.

### (a) Serial — works

```julia
ntuple(Val(NS)) do s
    Enzyme.autodiff(Forward, Const(rhs), Duplicated, Duplicated(u, seed_s))[1]
end
```

Compiles in ~29 s and emits NS separate `enzyme.fwddiff` ops (measured: 4 for NS = 4), correct values. That op count is exactly what an outer derivative then has to cross, which is the motivation for wanting a batched form.

### (b) BatchDuplicated — fails at XLA export

```julia
Enzyme.autodiff(Forward, Const(rhs), BatchDuplicated,
                BatchDuplicated(u, (v1, v2, v3, v4)))
```

```
error: 'enzyme.extract' op unsupported op for export to XLA
note: see current operation:
  %2 = "enzyme.extract"(%0) <{index = 0 : i64}> : (tensor<4x12xf64>) -> tensor<12xf64>
```

`enzyme.extract` survives the pipeline with no lowering.

### (c) Ops.batch around a width-1 fwddiff — fails in the batch lowering

```julia
Ops.batch([umat, seedmat], [NS]) do uu, vv
    Enzyme.autodiff(Forward, Const(rhs), Duplicated, Duplicated(uu, vv))[1]
end
```

```
error: TransposeOp operand rank 1 does not match permutation size 2
error: 'stablehlo.transpose' op failed to infer returned types
  note: %3 = "stablehlo.transpose"(%2) <{permutation = array<i64: 1, 0>}>
        : (tensor<12xf64>) -> tensor<12x4xf64>
```

### Expected

Either route should produce one `enzyme.fwddiff` of width NS.

### Why it matters

A coloured Jacobian built from N serial forward derivatives does not share one differentiated callee — each colour carries roughly its own copy of the RHS. On a real model this was measured at 120.1 MB against 34.8 MB for the same program with a finite-difference Jacobian, a 3.5x module and a 6.1x trace time. Batching is the standard remedy and is currently unavailable.

Pangoraw:
Closes #3217.


wsmoses:
this only resolves one of the cases from that issue right?

Pangoraw:
yeah the pipeline change is case b, c is done with https://github.com/EnzymeAD/Enzyme/pull/3235 and https://github.com/EnzymeAD/Enzyme/pull/3175 (and added a test for it here). a was already working.

wsmoses (approved):

